PR #2, opened 2015-09-04
Curvature correction

vepiteski:
Need to correct other functions as well (..._par  and cg)


dpo:
Going to close this, since we incorporated these changes over time.

PR #4, opened 2016-05-06
Matrix exponential

jagot:
I have added a routine to calculate the effect of exp(τA) on a vector v (useful for
time stepping of PDEs), where τ and v may be complex and A real symmetric
(possibly also valid for complex Hermitian A's). I have tried to implement it in
such a way that one does not need to (de)allocate a lot of working memory, if
it is used repeatedly, as it would be in time stepping. Still, there is some work to
be done, since for the test runs, ~600 k allocations totalling ~450 MB occur during
2000 time steps (10% garbage collection time, which is not good). I have not tried
to optimize it yet by calling BLAS directly, as is done in e.g. cg_lanczos. Using
Julia's profiling, it seems most of the time is spent in BLAS' dotc and Julia's
matmul, anyway.

When it comes to stopping criteria for the Krylov iteration, it is not my strong suit,
so I simply lifted it from cg_lanczos.jl. Could you have a look and see if it makes
sense?

Furthermore, I could not think of an especially good way to test the algorithm,
except for the case I would want to use it, namely propagation of the time-dependent
Schrödinger equation. I implemented time-stepping for a simple 1D particle-in-a-box,
and compare it with the solution found by Crank–Nicolson, which is very efficient
for this case (one order of magnitude faster for this case). If there is a better way of
testing, let me know.

Lastly, in the future, I would like to use this algorithm on a GPU, that is, perform
the Krylov iterations on the GPU, which is fast for matrix–vector products, and
calculate the subspace exponential on the CPU. Are there any plans in this
direction for Krylov.jl?

### DIFF
--- a/src/exp_lanczos.jl
+++ b/src/exp_lanczos.jl
@@ -49,7 +49,7 @@ function exp_lanczos!{T<:Number,R<:Real}(A :: LinearOperator,
     V[:,1] = v/β₀
     
     ε = atol + rtol * β₀
-    verbose && @printf("Initial norm: β₀ %g, stopping threshold: %g\n", β₀, ε)
+    verbose && @printf("Initial norm: β₀ %e, stopping threshold: %e\n", β₀, ε)
 
     fin = false
     j = 1
@@ -75,7 +75,7 @@ function exp_lanczos!{T<:Number,R<:Real}(A :: LinearOperator,
         σ = -ω * σ
         ω = ω * ω
         
-        verbose && @printf("iter %d, α[%d] %g, β[%d] %g, γ %g, ω %g, σ %g\n",
+        verbose && @printf("iter %d, α[%d] %e, β[%d] %e, γ %e, ω %e, σ %e\n",
                           j, j, α[j], j, β[j],
                           γ, ω, σ)
 
--- a/src/exp_lanczos.jl
+++ b/src/exp_lanczos.jl
@@ -30,9 +30,16 @@ function exp_lanczos_work{T<:Number}(A :: LinearOperator,
     V,α,β
 end
 
-function expT(α,β,v,τ,β₀)
-    ee = eigfact(SymTridiagonal(α,β))
-    β₀*v*ee[:vectors]*Diagonal(exp(τ*ee[:values]))*ee[:vectors][1,:]'
+if VERSION ≥ v"0.5.0-dev"
+    function expT(α,β,v,τ,β₀)
+        ee = eigfact(SymTridiagonal(α,β))
+        β₀*v*ee[:vectors]*Diagonal(exp(τ*ee[:values]))*ee[:vectors][1,:]
+    end
+else
+    function expT(α,β,v,τ,β₀)
+        ee = eigfact(SymTridiagonal(α,β))
+        β₀*v*ee[:vectors]*Diagonal(exp(τ*ee[:values]))*ee[:vectors][1,:]'
+    end
 end
 
 function exp_lanczos!{T<:Number,R<:Real}(A :: LinearOperator,
@@ -47,14 +54,14 @@ function exp_lanczos!{T<:Number,R<:Real}(A :: LinearOperator,
                                                verbose :: Bool=false)
     β₀ = norm(v)
     V[:,1] = v/β₀
-    
+
     ε = atol + rtol * β₀
     verbose && @printf("Initial norm: β₀ %e, stopping threshold: %e\n", β₀, ε)
 
     fin = false
     j = 1
     jj = 1 # Which Krylov subspace to use, in the end
-    
+
     σ = β₀
     ω = 0
     γ = 1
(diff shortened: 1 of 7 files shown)


dpo:
Many thanks for your contribution. This will probably take me a little while to review.

jagot:
I noticed that the Travis checks fail. They work on my computer, but I will fix them for Travis as well. Is it OK if I change .travis.yml to depend on the master branch of LinearOperators.jl? I need the commit I pushed a few days ago.

dpo:
I updated the `develop` branch of LinearOperators. Soon, we'll be able to rely on `master`.

jagot:
Failing tests are now due to #5, I have fixed the regressions I introduced with the commits up until 2cd8e19.

src/exp_lanczos.jl
```diff
@@ -0,0 +1,109 @@
...
+#
+# The implementation follows
+# Saad, Y. (1992). Analysis of some krylov subspace
+# approximations. SIAM Journal on Numerical Analysis.
+#
+# The error estimates/stopping criteria are borrowed from
+# cg_lanczos.jl
+#
+# Stefanos Carlström, <stefanos.carlstrom@gmail.com>
+# Lund, Sweden, May 2016.
```
dpo: Would you mind putting this into a docstring?

src/exp_lanczos.jl
```diff
@@ -0,0 +1,109 @@
...
+                                     verbose :: Bool=false)
+    n = size(v, 1);
+    (size(A, 1) == n & size(A, 2) == n) || error("Inconsistent problem size, $(size(A))×$(size(v))");
+    verbose && @printf("exp Lanczos: %d×%d system\n", n, n);
+
+    V = zeros(T, n, m+1)
+    α = zeros(real(T), m)
+    β = zeros(real(T), m)
+
+    V,α,β
```
dpo: I know there are no strong Julia conventions for style, but in the interest of consistency and readability, I'd prefer using `return (V, α, β)` with spacing. Same thing elsewhere about spacing around arithmetic operators (à la Python). I find that really improves readability.

src/exp_lanczos.jl
```diff
@@ -0,0 +1,109 @@
...
+function exp_lanczos!{T<:Number,R<:Real}(A :: LinearOperator,
+                                               v :: AbstractVector{T},
+                                               τ :: T, m :: Integer,
+                                               vp :: AbstractVector{T},
+                                               V :: Matrix{T},
+                                               α :: Vector{R},
+                                               β :: Vector{R};
+                                               atol :: Float64=1.0e-8,
+                                               rtol :: Float64=1.0e-4,
+                                               verbose :: Bool=false)
```
dpo: The indentation is a bit off here.

src/exp_lanczos.jl
```diff
@@ -0,0 +1,109 @@
...
+
+    V = zeros(T, n, m+1)
+    α = zeros(real(T), m)
+    β = zeros(real(T), m)
+
+    V,α,β
+end
+
+if VERSION ≥ v"0.5.0-dev"
+    function expT(α,β,v,τ,β₀)
```
dpo: These arguments should probably be typed?!

src/exp_lanczos.jl
```diff
@@ -0,0 +1,109 @@
...
+        β[j] = norm(V[:,j+1])
+        V[:,j+1] /= β[j]
+
+        ω = β[j] * γ
+        σ = -ω * σ
+        ω = ω * ω
+
+        verbose && @printf("iter %d, α[%d] %e, β[%d] %e, γ %e, ω %e, σ %e\n",
+                          j, j, α[j], j, β[j],
+                          γ, ω, σ)
```
dpo: Indentation is a bit off here too. I've no problem using long lines in Julia, so this could probably all fit on one line.

src/exp_lanczos.jl
dpo: Also I like to use two spaces for indentation.

src/exp_lanczos.jl
```diff
@@ -0,0 +1,109 @@
...
+
+# One-shot version
+function exp_lanczos{T<:Number}(A :: LinearOperator, v :: Array{T,1},
+                                τ :: T, m :: Integer;
+                                atol :: Float64=1.0e-8,
+                                rtol :: Float64=1.0e-6,
+                                verbose :: Bool=false)
+    l_work = exp_lanczos_work(A, v, m, verbose)
+    vp = similar(v)
+    exp_lanczos!(A, v, τ, m, vp, l_work...; βtol = βtol, verbose = verbose)
```
dpo: No space around `=` for keyword args.

test/test_exp_lanczos.jl
```diff
@@ -0,0 +1,188 @@
+plot_p = "TEST_KRYLOV_PLOT" in keys(ENV)
```
dpo: This test is a bit intense, with extra requirements. I think the test should be kept to a minimum, just validated computations. It's ok to move the rest to a demo in the `examples` folder.

dpo:
Thanks for your contribution!

jagot:
I hope you are happy with the changes.

I am still not sure about the stopping criteria, but lifting those found in cg_lanczos.jl seemed to work for the cases I tested.

When https://github.com/JuliaLang/julia/pull/16274 has been addressed properly, I will submit a new pull request here to make use of the possibility to not allocate LAPACK work arrays every time the Krylov exponentiation is performed.

abelsiqueira:
Hi all, this PR got lost into the backlog of PRs which I'm trying to clean-up. I'm guessing this probably doesn't work as expected anymore since it's from 3 years ago and is being pulled into `develop` which is also not being used anymore.

@jagot, sorry for the delay in dealing with this. Do you want to look into it in more details or have you moved on?
Looking at it, it looks like an application to solve DEs. Maybe this could be used in the docs?

jagot:
An improved version that also supports Arnoldi, based on the work in https://github.com/JuliaDiffEq/ExponentialUtilities.jl, was merged there some time ago. Maybe this PR is a bit out-of-scope for Krylov.jl, which seems mostly focused on solving equation systems, so I think we can close it.

dpo:
Ok, thank you.

PR #9, opened 2016-08-02
Lslq paper - 2-norm error estimate

restrin:
This pull-request includes the addition of Gauss-Radau quadrature based upper-bound estimate of the LSLQ (and consequently LSQR) 2-norm forward error ||x_k - x*||_2. The shifted linear system that needs to be solved as part of Gauss-Radau quadrature is done via Cholesky decomposition since T_k - (\lambda_est)*I is guaranteed to be SPD.

A new stopping criterion has been added, if ||x_k - x*||_2 <= err_ubnd <= etol*||x_k||.

LSLQ also now takes an extra argument, \lambda_est, which is an underestimate of the smallest eigenvalue of A'*A. \lambda_est becomes \lambda_est + \lambda^2 since eigenvalues get shifted right when regularization is applied.

This requires \lambda_est > 0, otherwise no new behaviour should be exhibited.

### DIFF
--- a/src/lslq.jl
+++ b/src/lslq.jl
@@ -63,7 +63,7 @@ function lslq(A :: AbstractLinearOperator, b :: Array{Float64,1}, x_exact :: Vec
               λ :: Float64=0.0, atol :: Float64=1.0e-8, btol :: Float64=1.0e-8,
               etol :: Float64=1.0e-8, window :: Int=5,
               itmax :: Int=0, conlim :: Float64=1.0e+8, 
-              a :: Float64=0.0, verbose :: Bool=false)
+              λest :: Float64=0.0, verbose :: Bool=false)
 
   m, n = size(A)
   size(b, 1) == m || error("Inconsistent problem size")
@@ -73,6 +73,10 @@ function lslq(A :: AbstractLinearOperator, b :: Array{Float64,1}, x_exact :: Vec
   sqd && (λ = 1.0)
   λ² = λ * λ
   ctol = conlim > 0.0 ? 1/conlim : 0.0
+
+  # Check if estimate to minimum eigenvalue of A'A provided
+  # If not but λ² > 0, use that instead
+  λest = max(λ², λest)
 
   # Initialize Golub-Kahan process.
   # β₁ M u₁ = b.
@@ -117,9 +121,9 @@ function lslq(A :: AbstractLinearOperator, b :: Array{Float64,1}, x_exact :: Vec
   s = 0.0
   c = -1.0
 
-  if (a > 0)
+  if (λest > 0)
     μ = sqrt(ᾱ )         # = μ₁
-    ρ = sqrt(μ * μ - a)  # = ρ₁
+    ρ = sqrt(μ * μ - λest)  # = ρ₁
   end
 
   w = zeros(n)       # = w₀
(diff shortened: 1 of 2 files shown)


src/lslq.jl
```diff
@@ -74,9 +74,9 @@ function lslq(A :: AbstractLinearOperator, b :: Array{Float64,1}, x_exact :: Vec
   λ² = λ * λ
   ctol = conlim > 0.0 ? 1/conlim : 0.0
 
-  # Check if estimate to minimum eigenvalue of A'A provided
-  # If not but λ² > 0, use that instead
-  λ_est = max(λ², λ_est)
+  # Eigenvalue estimate is sum of regularization and
+  # eigenvalue estimate of A'A
+  λ_est = λ_est + λ²
```
dpo: :+1:

dpo: Thinking more about this, since we're solving a least-squares problem and supplying A (and not A'A), shouldn't the user be asked to provide an estimate of the smallest singular value of A instead? Then we could set lambda_est = sigma_est^2 + lambda^2.

restrin: I agree, I've changed it to take a singular value estimate with lambda_est = sigma_est^2 + lambda^2.

There should already be if-guards around all behaviour involving the bound estimates checking if lambda_est > 0 (equivalent to at least one of sigma_est > 0 or lambda_est > 0).

src/lslq_methods.jl
```diff
@@ -1,15 +1,15 @@
...
-                                                etol :: Float64=1.0e-8, window :: Int=5,
-                                                itmax :: Int=0, conlim :: Float64=1.0e+8, verbose :: Bool=false) =
-  lslq(LinearOperator(A), b, x_exact, M=M, N=N, sqd=sqd, λ=λ, atol=atol, btol=btol, etol=etol, window=window, itmax=itmax, conlim=conlim, verbose=verbose);
+lslq{TA <: Number, Tb <: Number, Tc <: Number}(A :: Array{TA,2}, b :: Array{Tb,1}, x_exact :: Array{Tc,1};
+                                 M :: AbstractLinearOperator=opEye(size(A,1)), N :: AbstractLinearOperator=opEye(size(A,2)),
+                                 sqd :: Bool=false,
+                                 λ :: Float64=0.0, atol :: Float64=1.0e-8, btol :: Float64=1.0e-8,
+                                 etol :: Float64=1.0e-8, window :: Int=5,
+                                 itmax :: Int=0, conlim :: Float64=1.0e+8, σ_est :: Float64=0.0, verbose :: Bool=false) =
+  lslq(LinearOperator(A), b, x_exact, M=M, N=N, sqd=sqd, λ=λ, atol=atol, btol=btol, etol=etol, window=window, itmax=itmax, conlim=conlim, σ_est=0.0, verbose=verbose);
```
dpo: This just caused me a small headache. I couldn't understand why LSLQ always thought σ_est = 0... Line 7 should say `σ_est=σ_est` (the value that was passed should be passed along).

No worries though. I fixed it locally.

restrin: Oh I didn't even notice I did that! Sorry about that.

dpo: No problem. I'm now happy to report that the error estimate works also on rank-deficient problems!!! 🎉

dpo:
I'll close this now that the strategy changed.

PR #16, opened 2016-10-17
Return correctly on zero-residual solution

### LINKED ISSUE #15: zero rhs - return is inconsistent
At the end of the method, it returns x and stats, but in some places, it returns only x (e.g., `cg` when `b` is zero). Is this by design, or should I change it?

abelsiqueira:
Closes #15

### DIFF
--- a/test/test_cg.jl
+++ b/test/test_cg.jl
@@ -54,3 +54,8 @@ b = B * ones(n)
 (x, stats) = cg(B, b, itmax=2n)
 @test x ≈ ones(n)
 @test stats.solved
+
+# Test b == 0
+(x, stats) = cg(A, zeros(size(A,1)))
+@test x == zeros(size(A,1))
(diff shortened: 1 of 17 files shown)


test/test_cg.jl
```diff
@@ -54,3 +54,8 @@ b = B * ones(n)
 (x, stats) = cg(B, b, itmax=2n)
 @test x ≈ ones(n)
 @test stats.solved
+
+# Test b == 0
+(x, stats) = cg(A, zeros(size(A,1)))
+@test x == zeros(size(A,1))
+@test stats.status == "x = 0 is a zero-residual solution"
```
dpo: Great, thank you! If that's not abusing, do you think you could add a similar test for the other methods?

abelsiqueira:
Updated

dpo:
Many, many thanks!

PR #25, opened 2017-01-26
Draft of dispatch on types

dpo:
The main goals here are:
1. allow calls to iterative methods with different data types (e.g., Float16, BigFloat, etc.)
2. still permit efficient BLAS calls whenever possible (i.e., when the type is `<: BlasReal`)

For complex data, it doesn't seem that Julia calls `BLAS.cdot` or `BLAS.zdot` on those types. So it's probably ok to just call `dot`.

@abelsiqueira

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -15,19 +15,19 @@ export cg
 
 The method does _not_ abort if A is not definite.
 """
-function cg{T <: Real}(A :: AbstractLinearOperator, b :: Vector{T};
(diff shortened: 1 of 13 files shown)


abelsiqueira:
This is looking very good to me. Do you think this beats IterativeSolvers.jl?

dpo:
I haven't benchmarked them lately, but right now, it's mostly for convenience. This shouldn't improve the performance of Krylov.jl, it just simplifies the code and makes it more generic, while at the same time ensuring that BLAS calls are used when possible.

PR #27, opened 2017-06-09
Create minres.jl

DahitoMA:
A simple correction in minres code by moving the stopping condition relating to the machine accuracy.

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -134,11 +134,6 @@ function minres{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
     β = sqrt(β)
     ANorm² = ANorm² + α * α + oldβ * oldβ + β * β
 
-    if iter == 1
-      # A'b = 0 so x = 0 is a minimum least-squares solution
-      β / β₁ ≤ 10 * ϵM && return (x, SimpleStats(true, true, [β₁], [0.0], "x = 0 is a minimum least-squares solution"))
-    end
-
     # Apply rotation to obtain
     #  [ δₖ    ϵₖ₊₁    ] = [ cs  sn ] [ δbarₖ  0    ]
     #  [ γbar  δbarₖ₊₁ ]   [ sn -cs ] [ αₖ     βₖ₊₁ ]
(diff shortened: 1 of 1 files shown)


src/minres.jl
```diff
@@ -213,7 +213,7 @@ function minres{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
     ill_cond = ill_cond_mach | ill_cond_lim
     solved = solved_mach | solved_lim | zero_resid_mach | zero_resid_lim | fwd_err
   end
-
+
```
dpo: Thank you. Could you please just remove this extra space? You can simply add a new commit and push to your branch. The updates will appear here automatically.

src/minres.jl
```diff
@@ -195,6 +190,11 @@ function minres{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
 
     verbose && @printf("%5d  %7.1e  %7.1e  %7.1e  %8.1e  %8.1e  %7.1e  %7.1e\n",
                        iter, test1, test2, β, cs, sn, ANorm, Acond)
+    
+    if iter == 1
+      # A'b = 0 so x = 0 is a minimum least-squares solution
+      β / β₁ ≤ 10 * ϵM && return (x, SimpleStats(true, true, [β₁], [0.0], "x = 0 is a minimum least-squares solution"))
+    end
```
dpo: The message should be changed a bit here. Just "x is a minimum least-squares solution" should do.

src/minres.jl
```diff
@@ -213,7 +213,6 @@ function minres{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
     ill_cond = ill_cond_mach | ill_cond_lim
     solved = solved_mach | solved_lim | zero_resid_mach | zero_resid_lim | fwd_err
   end
-
```
dpo: No need to delete the line. There should be a blank line but it should be empty (the previous commit added whitespace on that line). Thank you.

dpo:
Great! Thank you!

PR #29, opened 2017-07-10
An implementation of cr

DahitoMA:
add cr method
add cr tests


src/cr.jl
```diff
@@ -0,0 +1,96 @@
...
+  verbose && @printf("%5s %6s %10s %10s %10s %10s\n", "Iter", "‖x‖", "‖r‖", "q", "α", "σ")
+  verbose && @printf("    %d  %8.1e    %8.1e    %8.1e", iter, xNorm, rNorm, m)
+
+  solved = rNorm <= ε
+  tired = iter >= itmax
+  on_boundary = false
+  status = "unknown"
+
+  while ! (solved || tired)
+    α = ρ / @knrm2(n, q)^2 # step
```
dpo: This should probably be a call to `@kdot` to avoid computing a square root and then squaring the result.

src/cr.jl
```diff
@@ -0,0 +1,96 @@
...
+    if (radius > 0.0) & (α > σ)
+      α = σ
+      on_boundary = true
+    end
+
+    @kaxpy!(n,  α,  p, x)
+    xNorm = @knrm2(n, x)
+    push!(xNorms, xNorm)
+    Ax = A * x
+    m = @kdot(n, -b, x) + 1/2 * @kdot(n, x, Ax)
```
dpo: `-@kdot(n, b, x)` would avoid multiplying an entire vector by -1 and creating a temporary vector for `-b`. The `½` could also be just `0.5`.

src/cr.jl
```diff
@@ -0,0 +1,96 @@
...
+
+    solved = (rNorm <= ε) | on_boundary
+    tired = iter >= itmax
+
+    if !solved
+      ρbar = ρ
+      ρ = @kdot(n, r, Ar)
+      β = ρ / ρbar # step for the direction calculus
+      p = r + β * p # descent direction
+      q = Ar + β * q
```
dpo: These two could be calls to `@kaxpy`.

src/cr.jl
```diff
@@ -0,0 +1,96 @@
...
+    push!(mvalues, m)
+    @kaxpy!(n, -α, q, r) # residual
+    rNorm = @knrm2(n, r)
+    push!(rNorms, rNorm)
+    Ar = A * r
+    ArNorm = @knrm2(n, Ar)
+    push!(ArNorms, ArNorm)
+
+    solved = (rNorm <= ε) | on_boundary
+    tired = iter >= itmax
```
dpo: Here you could add a line `(solved | tired) && continu` to avoid the following `if ... end`.

src/cr.jl
DahitoMA: It seems that @kaxpy modifies the last argument, so it would also modify r and Ar, not only p and q.

dpo: Yes, what I meant is a `@kscal` followed by a `@kaxpy` with factor 1. But it's not clear that it would be faster.

dpo:
The test `xnorm == radius` fails on some platforms:
https://travis-ci.org/JuliaSmoothOptimizers/Krylov.jl/jobs/252165315#L351.

It would be safer to use [`@test_approx_eq`](https://docs.julialang.org/en/release-0.5/stdlib/test/#Base.Test.@test_approx_eq) or [`@test_approx_eq_eps`](https://docs.julialang.org/en/release-0.5/stdlib/test/#Base.Test.@test_approx_eq_eps).

test/test_cr.jl
```diff
@@ -0,0 +1,85 @@
...
+
+# Code coverage
+(x, stats) = cr(A, b)
+show(stats)
+
+radius = 0.75 * norm(x)
+(x, stats) = cr(A, b, atol, rtol, itmax, radius)
+show(stats)
+@test(stats.solved)
+@test(abs(radius - norm(x)) <= cr_tol * radius)
```
dpo: Should this be `test_approx_eq` like below? There's also `@test_approx_eq_eps` if you want to specify the tolerance.

src/cr.jl
```diff
@@ -0,0 +1,96 @@
...
+    if (radius > 0.0) & (α > σ)
+      α = σ
+      on_boundary = true
+    end
+
+    @kaxpy!(n,  α,  p, x)
+    xNorm = @knrm2(n, x)
+    push!(xNorms, xNorm)
+    Ax = A * x
+    m = - @kdot(n, b, x) + 0.5 * @kdot(n, x, Ax)
```
dpo: This is useful while developing your trust-region method but it's so expensive that it probably shouldn't be in the default version of `cr`. Perhaps it could be updated recursively as in `cg`: initially, `Ax = 0` and recursively, we have `Ax = Ax + α A * p = Ax + α * q`. But still, the update will probably suffer substantial numerical errors.

dpo: Same comment about `xNorm`.

src/cr.jl
```diff
@@ -0,0 +1,96 @@
...
+    iter = iter + 1
+    verbose && @printf("    %d  %8.1e    %8.1e    %8.1e", iter, xNorm, rNorm, m)
+
+    solved = (rNorm <= ε) | on_boundary
+    tired = iter >= itmax
+    
+    (solved || tired) && continue
+    ρbar = ρ
+    ρ = @kdot(n, r, Ar)
+    β = ρ / ρbar # step for the direction calculus
```
dpo: "calculus" should be "computation" I think?!

src/variants.jl
```diff
@@ -1,6 +1,6 @@
 # Variants
 
-for fn in (:cg_lanczos, :cg_lanczos_shift_seq, :cg, :cgls, :cgne,
+for fn in (:cr, :cg_lanczos, :cg_lanczos_shift_seq, :cg, :cgls, :cgne,
```
dpo: Let's keep them in alphabetical order :-).

dpo:
Almost there! Thank you!

dpo:
I think the failure with Julia 0.4 is because it did not allow macros to have multiple definitions. I suppose we could phase out version 0.4.

dpo:
See #30.

PR #35, opened 2017-09-22
Return the two roots in 'to_boundary'

DahitoMA:
Return the two roots for the function to_boundary
update cg.jl and cr.jl


dpo:
`test_aux.jl` should also be updated.

dpo:
Thank you!

PR #38, opened 2017-11-25
0.7 compatibility

dpo:


dpo:
Appveyor download error appears to be transient.

PR #41, opened 2017-12-03
tolerances for optimization

dpo:
The tolerance named `atol` is now called `axtol`. New tolerances `atol` and `rtol` play the role of absolute and relative tolerances on the optimality conditions.


dpo:
@DahitoMA

PR #44, opened 2018-05-17
Symmlq

restrin:
Implementation of SYMMLQ with its error estimate, CG error estimate and sliding window improvement. Mildly tested, although further testing for corner cases would be good.


dpo:
Many thanks, Ron! This looks excellent. But would you mind rebasing your branch to resolve the conflicts?

src/symmlq.jl
```diff
@@ -0,0 +1,275 @@
...
+  ϵold = 0.0
+  cold = 1.0
+
+  wbar = vold
+  ζold = 0.0
+  ζbar = β₁/γbar
+
+  ANorm² = α * α + β * β
+
+  x = zeros(T, n)
```
dpo: `x` is defined twice (lines 41 and 77).

src/symmlq.jl
```diff
@@ -0,0 +1,275 @@
...
+
+  γmax = -Inf
+  γmin = Inf
+  ANorm = 0.0
+  Acond = 0.0
+  rNorm = β₁
+  rcgNorm = β₁
+  xNorm = 0.0
+  xNormcg = 0.0
+  xcgNorm = 0.0
```
dpo: What's the difference between these two? Also xcgNorm should be abs(zetabar) I think?!

dpo:
There are conflicts that prevent me from merging. Rebasing also gives lots of conflicts. Do you think you could start a fresh PR that's in sync with `master` so we keep your name on the commits? Thanks!

dpo:
Closing in favor of #48

PR #46, opened 2018-06-01
Add case of zero curvature in crls

DahitoMA:
- Add a boolean 'psd' to detect if A is positive semi-definite
- Check curvature in the direction p
- If zero curvature is encountered, choose the direction of best decrease between p and A'r
- Add test for A positive semi-definite

### DIFF
--- a/src/crls.jl
+++ b/src/crls.jl
@@ -94,6 +94,7 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
       if ApNorm² ≤ ε * norm(q) * pNorm # q is linear in the direction p
         psd = true # det(AᵀA) = 0
         pAr = @kdot(n, p, Ar) # pᵀAᵀr
+        ArNorm² = ArNorm^2
         if abs(pAr) ≤ ε * pNorm * ArNorm # q is constant in the direction p
           p = Ar # p = Aᵀr
           q = A' * s
@@ -108,7 +109,6 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
           if !descent
             σ = minimum(to_boundary(x, p, radius)) # < 0
           end
-          ArNorm² = ArNorm^2
           ν = min(ArNorm² / (2 * γ), maximum(to_boundary(x, Ar, radius)))
           δ = -σ * pAr + ν * ArNorm² + σ^2 * ApNorm² - ν^2 * γ
           if δ > 0.0
--- a/src/crls.jl
+++ b/src/crls.jl
@@ -83,37 +83,35 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
   psd = false
 
   while ! (solved || tired)
-    α = γ / @kdot(n, q, q)     # dot(q, q);
+    qNorm² = @kdot(n, q, q) # dot(q, q)
+    α = γ / qNorm²
 
     # if a trust-region constraint is give, compute step to the boundary
     # (note that α > 0 in CRLS)
     if radius > 0.0
-      σ = maximum(to_boundary(x, p, radius))
       pNorm = @knrm2(n, p)
+      σ = maximum(to_boundary(x, p, radius, flip = false, dNorm = pNorm))
       ApNorm² = @knrm2(m, Ap)^2
-      if ApNorm² ≤ ε * norm(q) * pNorm # q is linear in the direction p
+      if ApNorm² ≤ ε * sqrt(qNorm²) * pNorm # the quadratic is linear in the direction p
         psd = true # det(AᵀA) = 0
         pAr = @kdot(n, p, Ar) # pᵀAᵀr
-        ArNorm² = ArNorm^2
-        if abs(pAr) ≤ ε * pNorm * ArNorm # q is constant in the direction p
+        if abs(pAr) ≤ ε * pNorm * ArNorm # the quadratic is constant in the direction p
           p = Ar # p = Aᵀr
+          pNorm = ArNorm
           q = A' * s
-          if γ > 0.0
-            α = min(ArNorm² / γ, maximum(to_boundary(x, p, radius))) # q is minimal in the direction A'r for α = ‖Ar‖²/γ
-          else
-            # q is linear in the direction Aᵀr
-            α = maximum(to_boundary(x, p, radius))
-          end
+          α = maximum(to_boundary(x, p, radius, flip = false, dNorm = pNorm))
+          γ > 0.0 && (α = min(ArNorm^2 / γ, α)) # the quadratic is minimal in the direction Aᵀr for α = ‖Ar‖²/γ
         else
           descent = pAr > 0.0
           if !descent
-            σ = minimum(to_boundary(x, p, radius)) # < 0
+            σ = minimum(to_boundary(x, p, radius, flip = false, dNorm = pNorm)) # < 0
           end
-          ν = min(ArNorm² / γ, maximum(to_boundary(x, Ar, radius)))
+          ArNorm² = ArNorm * ArNorm
+          ν = min(ArNorm² / γ, maximum(to_boundary(x, Ar, radius, flip = false, dNorm = ArNorm)))
           δ = -σ * pAr + ν * ArNorm² + σ^2 * ApNorm² - ν^2 * γ
           if δ > 0.0
-            # direction A'r engenders a bigger decrease
-            p = Ar # p = A'r
+            # direction Aᵀr engenders a bigger decrease
+            p = Ar # p = Aᵀr
             q = A' * s
             α = ν
           else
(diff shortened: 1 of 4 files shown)


test/test_crls.jl
```diff
@@ -59,3 +59,19 @@ A = [eye(Int, 3); rand(1:10, 2, 3)]
...
 @test stats.solved
+
+# Test A positive semi-definite
+radius = 10.
+m,n = 10,7
+U=qr(rand(m,m))[1]
+V=qr(rand(n,n))[1]
+V = V'
+S = zeros(m,n)
+S[2,2]=1e-6;S[3,3]=1;S[4,4]=4;S[5,5]=20;S[6,6]=15;S[7,7]=1e5;
```
dpo: Could you just say `S = diagm(1.0e-6, 1, 4, 20, 15, 1.0e5)`?

test/test_crls.jl
```diff
@@ -59,3 +59,19 @@ A = [eye(Int, 3); rand(1:10, 2, 3)]
 b = A * ones(Int, 3)
 (x, stats) = crls(A, b)
 @test stats.solved
+
+# Test A positive semi-definite
+radius = 10.
+m,n = 10,7
+U=qr(rand(m,m))[1]
+V=qr(rand(n,n))[1]
+V = V'
```
dpo: This is not necessary. Both V and V' are orthogonal.

test/test_crls.jl
DahitoMA: It seems that diagm gives a square matrix but S needs to be m x n.

dpo: Ok. This should still be a bit more readable:
```
S = [diagm([0, 1.0e-6, 1, 4, 20, 15, 1.0e5]) ; zeros(3,7)]
```

test/test_crls.jl
```diff
@@ -59,3 +59,19 @@ A = [eye(Int, 3); rand(1:10, 2, 3)]
...
+S = zeros(m,n)
+S[2,2]=1e-6;S[3,3]=1;S[4,4]=4;S[5,5]=20;S[6,6]=15;S[7,7]=1e5;
+A = U * S * V'
+p = V[:,1]; b = A'\p;
+Aop = LinearOperator(A);
+(x, stats) = crls(Aop, b, radius=radius)
```
dpo: No need to create `Aop`.

test/test_crls.jl
```diff
@@ -59,3 +59,19 @@ A = [eye(Int, 3); rand(1:10, 2, 3)]
...
+@test stats.solved
+@test stats.status == "zero-curvature encountered"
+@test norm(x) ≤ radius
```
dpo: Should they be equal in this case?

DahitoMA: The step length α is optimal so theoretically the norm of x could be lower than the radius.

src/crls.jl
```diff
@@ -80,25 +80,63 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
...
     # (note that α > 0 in CRLS)
-    σ = radius > 0.0 ? maximum(to_boundary(x, p, radius)) : α
-    if (radius > 0.0) & (α > σ)
-      α = σ
-      on_boundary = true
+    if radius > 0.0
+      σ = maximum(to_boundary(x, p, radius))
+      pNorm = @knrm2(n, p)
+      ApNorm² = @knrm2(m, Ap)^2
+      if ApNorm² ≤ ε * norm(q) * pNorm # q is linear in the direction p
```
dpo: `norm(q)` (squared) was already computed on line 86.

src/crls.jl
```diff
@@ -80,25 +80,63 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
...
 
     # if a trust-region constraint is give, compute step to the boundary
     # (note that α > 0 in CRLS)
-    σ = radius > 0.0 ? maximum(to_boundary(x, p, radius)) : α
-    if (radius > 0.0) & (α > σ)
-      α = σ
-      on_boundary = true
+    if radius > 0.0
+      σ = maximum(to_boundary(x, p, radius))
+      pNorm = @knrm2(n, p)
```
dpo: `to_boundary()` will also compute `pNorm`. Could we modify `to_boundary()` so that we can pass in `pNorm` if it is already available?

src/crls.jl
```diff
@@ -80,25 +80,63 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
...
+      ApNorm² = @knrm2(m, Ap)^2
+      if ApNorm² ≤ ε * norm(q) * pNorm # q is linear in the direction p
+        psd = true # det(AᵀA) = 0
+        pAr = @kdot(n, p, Ar) # pᵀAᵀr
+        ArNorm² = ArNorm^2
```
dpo: We should only compute this if/when it is needed.

src/crls.jl
```diff
@@ -80,25 +80,63 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
...
+        if abs(pAr) ≤ ε * pNorm * ArNorm # q is constant in the direction p
```
dpo: There's a confusion with the vector `q` in this comment.

src/crls.jl
```diff
@@ -80,25 +80,63 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
...
+          p = Ar # p = Aᵀr
+          q = A' * s
+          if γ > 0.0
+            α = min(ArNorm² / γ, maximum(to_boundary(x, p, radius))) # q is minimal in the direction A'r for α = ‖Ar‖²/γ
+          else
+            # q is linear in the direction Aᵀr
+            α = maximum(to_boundary(x, p, radius))
```
dpo: This alpha could be computed before the if.

dpo:
Tests are failing because of
```
INFO: Testing Krylov
Test Failed
  Expression: maximum(Krylov.to_boundary(x, d, 5.0, flip=true)) == 1.8099751242241782
   Evaluated: 1.809975124224178 == 1.8099751242241782
ERROR: LoadError: LoadError: There was an error during testing
while loading C:\Users\appveyor\.julia\v0.6\Krylov\test\test_aux.jl, in expression starting on line 77
while loading C:\Users\appveyor\.julia\v0.6\Krylov\test\runtests.jl, in expression starting on line 7
===============================[ ERROR: Krylov ]================================
```
and
```
Test Failed
  Expression: norm(x) ≤ radius
   Evaluated: 10.000000000000002 ≤ 10.0
ERROR: LoadError: LoadError: There was an error during testing
while loading /Users/travis/.julia/v0.6/Krylov/test/test_crls.jl, in expression starting on line 74
while loading /Users/travis/.julia/v0.6/Krylov/test/runtests.jl, in expression starting on line 12
```

src/crls.jl
```diff
@@ -80,25 +80,61 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
...
+        if abs(pAr) ≤ ε * pNorm * ArNorm # the quadratic is constant in the direction p
+          p = Ar # p = Aᵀr
+          pNorm = ArNorm
+          q = A' * s
+          α = maximum(to_boundary(x, p, radius, flip = false, dNorm = pNorm))
+          γ > 0.0 && (α = min(ArNorm^2 / γ, α)) # the quadratic is minimal in the direction Aᵀr for α = ‖Ar‖²/γ
+        else
+          descent = pAr > 0.0
+          if !descent
+            σ = minimum(to_boundary(x, p, radius, flip = false, dNorm = pNorm)) # < 0
```
dpo: Aren't we calling `to_boundary()` twice (here and on line 93)?

dpo:
Thank you! Do you think you could try to cook up examples where the regions of `crls.jl` that are [not currently tested](https://codecov.io/gh/JuliaSmoothOptimizers/Krylov.jl/pull/46/diff?src=pr&el=tree#diff-c3JjL2NybHMuamw=) would be activated?

src/crls.jl
DahitoMA: That is true, I will change it.

src/crls.jl
```diff
@@ -80,25 +80,43 @@ function crls{T <: Number}(A :: AbstractLinearOperator, b :: Vector{T};
...
+      ζ = to_boundary(x, p, radius, flip = false, dNorm = pNorm)
+      σ = maximum(ζ)
+      ApNorm² = @knrm2(m, Ap)^2
+      if ApNorm² ≤ ε * sqrt(qNorm²) * pNorm # the quadratic is constant in the direction p
+        psd = true # det(AᵀA) = 0
+        p = Ar # p = Aᵀr
+        pNorm = ArNorm
+        q = A' * s
+        α = maximum(to_boundary(x, p, radius, flip = false, dNorm = pNorm))
+        γ > 0.0 && (α = min(ArNorm^2 / γ, α)) # the quadratic is minimal in the direction Aᵀr for α = ‖Ar‖²/γ
```
dpo: Could γ be zero? What would happen then? If s = AA'r = 0, then r'AA'r = 0, i.e., A'r = 0, and we are optimal?!

DahitoMA: That is true, γ cannot be equal to zero if we are in the While loop. I will change this.

dpo:
Great, thank you!

PR #51, opened 2018-06-18
delete warning and SparseVector for high-dimensional problems

Xx-fractale-xX:


dpo:
There seems to be a small issue [here](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/src/lslq.jl#L216). `βL` should probably be initialized just before the `if` on line 200.

test/test_variants.jl
```diff
@@ -0,0 +1,36 @@
...
+
+b_dense1 = ones(Float32,5)
+b_dense2 = ones(Float64,5)
+
+b_sparse1 = convert(SparseVector{Float32,Int32},b_dense1)
+b_sparse2 = convert(SparseVector{Float64,Int64},b_dense2)
+
+for fn in (:cg_lanczos, :cg, :cgls, :cgne, :cr, :craig, :craigmr, :crls, :crmr, :lslq, :lsmr, :lsqr, :minres, :symmlq)
+	@eval begin
+		$fn(A_sparse1, b_sparse1)
```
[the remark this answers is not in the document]
Xx-fractale-xX: I forgot to test Int32 and Int64 types for elements of A and b. I will modify indentation in the same time.

Xx-fractale-xX:
I will initialize ```αL``` and ```βL``` in the same time of other constants to avoid this issue.

src/variants.jl
```diff
@@ -7,13 +7,21 @@ for fn in (:cg_lanczos, :cg_lanczos_shift_seq, :cg, :cgls, :cgne,
...
       $fn(LinearOperator(A), convert(Vector{Float64}, b), args...; kwargs...)
 
-    # Variant for A given as a sparse matrix
+    # Variant for A given as a sparse matrix and b given as a dense vector
     $fn{TA <: Number, Tb <: Number, IA <: Integer}(A :: SparseMatrixCSC{TA,IA}, b :: Array{Tb,1}, args...; kwargs...) =
       $fn(LinearOperator(A), convert(Vector{Float64}, b), args...; kwargs...)
 
+    # Variant for A given as a dense array and b given as a sparse vector
+    $fn{TA <: Number, Tb <: Number, Ib <: Integer}(A :: Array{TA,2}, b :: SparseVector{Tb,Ib}, args...; kwargs...) =
+      $fn(LinearOperator(A), convert(Vector{Float64}, b), args...; kwargs...)
```
dpo: Is it necessary to convert the sparse vector to dense?

dpo: I think it is necessary as the vectors generated during the iterations will likely not be sparse.

dpo:
Thank you!

PR #53, opened 2018-07-23
preallocation for LinearOperator

### LINKED ISSUE #45: Use operators with preallocation when possible
[`test_cg.jl`](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/examples/test_cg.jl#L15) gives an example of operator with preallocation. All methods should use one whenever possible. This requires updating [`variants.jl`](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/src/variants.jl) and checking that there are no side effects due to reusing storage.

Xx-fractale-xX:
fixes #45

### DIFF
--- a/src/variants.jl
+++ b/src/variants.jl
@@ -1,27 +1,56 @@
-# Variants
+# Define a generic linear operator with preallocation
+function preallocated_LinearOperator(A)
+  (n, m) = size(A)
+  Ap = zeros(n)
+  Atq = zeros(m)
+  return LinearOperator(n, m, false, false, p -> A_mul_B!(Ap, A, p), q -> At_mul_B!(Atq, A, q), q -> At_mul_B!(Atq, A, q))
+end
 
-for fn in (:cg_lanczos, :cg_lanczos_shift_seq, :cg, :cgls, :cgne,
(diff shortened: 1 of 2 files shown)


src/variants.jl
```diff
@@ -1,27 +1,56 @@
...
 
     # Variant for A given as a sparse matrix and b given as a sparse vector
     $fn{TA <: Number, Tb <: Number, IA <: Integer, Ib <: Integer}(A :: SparseMatrixCSC{TA,IA}, b :: SparseVector{Tb,Ib}, args...; kwargs...) =
-      $fn(LinearOperator(A), convert(Vector{Float64}, b), args...; kwargs...)
-
+      $fn(preallocated_LinearOperator(A), convert(Vector{Float64}, b), args...; kwargs...)
   end
 end
+
+# Define a symmetric and hermitian linear operator with preallocation
```
dpo: Let's just not mention "hermitian" at the moment, since complex operators aren't supported yet.

dpo:
Looks great, thank you!

src/variants.jl
Xx-fractale-xX: Ok, I modify it immediately.

PR #57, opened 2018-08-10
DQGMRES algorithm

Xx-fractale-xX:
We can optimize the way the mem+2 diagonals are stored in H. It depends if we want to keep H dense or sparse.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,12 @@ os:
 julia:
   - 0.6
   - 0.7
+  - 1.0
   - nightly
 
 matrix:
   allow_failures:
-    - julia: 0.7
+    - julia: 1.0
     - julia: nightly
 
 branches:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,13 +4,15 @@ environment:
     - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
     - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7-latest-win32.exe"
     - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7-latest-win64.exe"
+    - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/1.0/julia-1.0-latest-win32.exe"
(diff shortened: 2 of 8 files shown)


dpo:
Tremendous that you got around to implementing my favorite variant of GMRES! Many thanks! I'll look closely soon. Were you able to cross-check the results with the Matlab implementation?

Xx-fractale-xX:
No, I didn't cross-check the results with the Matlab implementation. But I thought that there is some little mistakes in Matlab code.

For example the current Givens rotation is compute by 
```matlab
[c(kpos), s(kpos), H(k,2)] = SymGivens(H(k,2), H(k+1,1))
```
in Matlab whereas I used
```julia
(c[rot_pos], s[rot_pos], H[iter,2]) = sym_givens(H[iter,2], H[iter,1])
```
in Julia. The first two diagonals have the same number of coefficient.

I saw another problem with positions of Givens rotations.
```matlab
jpos = mod(j-1, mem+1) + 1
Hjk = c(jpos) * H(j,kk1) + s(jpos) * H(j+1,kk)
H(j+1,kk) = s(jpos) * H(j,kk1) - c(jpos) * H(j+1,kk)
H(j,kk1) = Hjk
```
So ```jpos``` can be equal to ```mem + 1``` although the dimension of ```c``` and ```s``` is ```mem```.

In both case, Matlab needs to increase implicitly ```H```,  ```c``` and ```s```. 
And It's possible that there is other effects on the algorithm.
Therefore I don't know if it's a good idea to take Matlab code as reference.

Xx-fractale-xX:
Yesterday, I remarked that it's not necessary to store all the first diagonal of H.
At each iteration H[iter,1] is filled during the incomplete Arnoldi factorization but is replace by a zero right after with the QR factorization. 
So at each iteration the subdiagonal of our Hessenberg matrix contains at most one non-zero coefficient.
We can store it only as a float variable.

dpo:
This branch also needs rebasing. Thanks!

Xx-fractale-xX:
I rebased ```dqgmres``` and ```cgne_preconditioned``` branches.

dpo:
I wasn't able to rebase this against `master`. But I was able to cross-check DQGMRES locally against full GMRES (from IterativeSolvers). So it all looks good. I simply changed the format to display the residual. I'll open a new PR right away.

PR #60, opened 2018-09-14
An implementation of DQGMRES

dpo:


dpo:
Nevermind the appveyor/branch failure. That's not even supposed to be running.

tbonniot:
Thank you for cross-checking DQGMRES againt full GMRES! I forgot to do it.
You take the first version without the new storage for the first lower diagonal. Did you do that on purpose?

dpo:
I haven't yet fully understood the second version, so I used the version I understand. Perhaps you can open a new pull request with the improvement?

src/dqgmres.jl
```diff
@@ -0,0 +1,137 @@
...
+    z = M * V[:,pos] # P[:,pos]
+    w = A * z # V[:,next_pos]
+    for i = max(1, iter-mem+1) : iter
+      ipos = mod(i-1, mem+1) + 1
+      jpos = iter-i+2 # Indice of the diagonal
+      H[i,jpos] = @kdot(n, w, V[:,ipos]) # H[i.jpos] = < w , V[:,ipos] >
+      @kaxpy!(n, -H[i,jpos], V[:,ipos], w) # w = w - H[i,jpos] * V[:,ipos]
+    end
+    # jpos = iter-(iter+1)+2 = 1
+    H[iter,1] = @knrm2(n, w)
```
dpo: Actually I'm still confused. Why is this in `H[iter, 1]`? In the Arnoldi method, this value appears in `H[iter, iter+1]`.

dpo: Nevermind. I think it's correct.

Xx-fractale-xX: Sorry to answer only now, I wasn't notified when you commented.
In the code, `H` stored the diagonals of the original `H_arnoldi` matrix.

 An original Hessenberg matrix : 
[1 2 3 x x]                            
 [4 5 6 7 x]                            
[x 8 9 1 2]  
  [x x 3 4 5]
[x x x 6 7] 
 [x x x x 8]

will be stored as 

[4 1 2 3]
 [8 5 6 7]
 [3 9 1 2]
 [6 4 5 x]
 [8 7 x x]
                                                  
Furthermore, `H[:, i]` is the i° diagonal of `H_arnoldi`. So, `H[iter, iter+1]` is the iter° coefficient of the first diagonal => `H[iter, 1]`.

With the previous example, `H_arnoldi[6, 5] =  H[5, 1] = 8`.

And because of the fact that we did a QR factorization at each iteration and update the same matrix `H`. Only one coefficient of `H[:, 1]` isn't equal to zero => `H[iter, 1]`.
It's why it could be interesting to limit the fill-in of zero coefficients in the sparse matrix `H` by only storing `H[:,1]` (H[iter,1] ) in a simple float variable `h`.

PR #61, opened 2018-09-14
Preconditioned CGNE

Xx-fractale-xX:


dpo:
Thank you. Does preconditioning work in the rectangular case?

Xx-fractale-xX:
Yes, if dimensions of A are `m x n`, a square preconditioner `m x m` works. I will add a test with rectangular case.

Xx-fractale-xX:
cgne
I found data from an old rectangular problems. I verified the solution, it works.

test/test_cgne.jl
```diff
@@ -73,3 +73,23 @@ A = [eye(Int, 3); rand(1:10, 2, 3)]
...
+
+# Test preconditioner in rectangular case
+# Compute k and l constants such that y = k * exp(l * x) from observations
+# Minimise sum(eᵢ²) with eᵢ = l * xᵢ + ln(k) - ln(yᵢ)
+xᵢ = [0.97167; 1.92917; 3.07500; 4.03583; 5.02500; 5.98000; 7.03417; 8.06833; 9.02417; 9.92667]
+yᵢ = [0.721012; 0.729326; 0.795348; 0.852749; 0.886014; 0.791215; 0.920449; 0.912621; 0.810742; 0.979552]
+A = hcat(xᵢ, ones(10))
+b = log.(yᵢ)
+M = LinearOperator(Diagonal(1 ./ (A * A')))
+(x, stats) = cgne(A, b, M=M)
```
dpo: There needs to be a newline at the end of the file. Also we should check the solution somehow.

dpo:
That's an over-determined problem. It's almost surely inconsistent, so that's a test case for CGLS, LSQR and LSMR. CGNE is normally used on under-determined problems, where we seek the minimum norm solution. The same goes for CRAIG. Such examples would be great for the documentation though.

Xx-fractale-xX:
Do you want to keep this example or I find another one which is under-determined?

dpo:
CGNE should diverge on an example like this. This example is good for CGLS, LSQR and LSMR.

Xx-fractale-xX:
I found an interesting under-determined problem:
under-determined problem
I didn't detailed how I computed A but I think it could be a good example for the documentation too.

dpo:
It's a great example! Many thanks!

PR #62, opened 2018-09-14
Improve storage of Hessenberg matrix in DQGMRES

Xx-fractale-xX:


src/dqgmres.jl
```diff
@@ -105,19 +104,19 @@ function dqgmres{T <: Number}(A :: AbstractLinearOperator, b :: AbstractVector{T
     # Compute and apply current (symmetric) Givens rotation
     # [ck  sk] [ H[iter,iter]   ] = [ρ]
     # [sk -ck] [ H[iter+1,iter] ]   [0].
-    (c[rot_pos], s[rot_pos], H[iter,2]) = sym_givens(H[iter,2], H[iter,1])
-    H[iter,1]   = 0
+    (c[rot_pos], s[rot_pos], H[iter,1]) = sym_givens(H[iter,1], h)
+    # H[iter+1,iter] = 0, we can rewrite in h at the next iteration.
```
dpo: overwrite h

Xx-fractale-xX:
I completely optimized `DQGMRES`. The biggest improvement is that we don't need to store the  sparse Hessenberg matrix anymore. We can use a dense vector instead.

To summarize, we only need ```2n + 4``` vectors of dimension ```mem``` with `n` the dimension of the problem.

I also add new comments.

dpo:
Ok, I understand what you are doing, now. Thank you!
Before we merge, would you mind posting the results of a `@benchmark` of the old and new `dqgmres` to see if the memory savings are apparent?

Xx-fractale-xX:
I did some benchmarks, but I don't understand why there is still so many allocations...

However, solving the problem requires less time. I think that it's a consequence of working with a dense vector instead of a sparse matrix for the Hessenberg matrix H. The Read/Write operations are faster. 

```julia
julia> AA = rand(1000,1000)

julia> A = preallocated_LinearOperator(AA)

julia> b = rand(1000)

julia> mem = 50

julia> @benchmark dqgmres_old($A,$b,memory=$mem)
BenchmarkTools.Trial: 
  memory estimate:  2.34 GiB
  allocs estimate:  1220656
  --------------
  minimum time:     2.599 s (4.78% GC)
  median time:      2.612 s (4.88% GC)
  mean time:        2.612 s (4.88% GC)
  maximum time:     2.626 s (4.97% GC)
  --------------
  samples:          2
  evals/sample:     1

julia> @benchmark dqgmres($A,$b,memory=$mem)
BenchmarkTools.Trial: 
  memory estimate:  2.33 GiB
  allocs estimate:  1213131
  --------------
  minimum time:     1.531 s (8.04% GC)
  median time:      1.552 s (8.07% GC)
  mean time:        1.562 s (8.01% GC)
  maximum time:     1.612 s (7.85% GC)
  --------------
  samples:          4
  evals/sample:     1

**Comment by Xx-fractale-xX** [comment]
I investigated the problem and it's seem that the problem come from the lines:
```julia
H[diag] = @kdot(n, w, V[:,ipos]) # hᵢ.ₘ = < A * vₘ , vᵢ >
@kaxpy!(n, -H[diag], V[:,ipos], w) # w ← w - hᵢ.ₘ * vᵢ
@kaxpy!(n, -H[diag], P[:,ipos], z) # z ← z - hᵢ.ₘ * p
```
It's when a column of a matrix (V in this case) is involved, I suppose that julia do a copy for the operation?
It's a little bit problematic that allocations increase over iterations. The goal of `DIOM` and `DQGMRES` is to work with a fix "amount" of memory.

dpo:
I wonder if it works to replace `V[:,ipos]` by `@view V[:,ipos]`, and similarly for the others.

src/dqgmres.jl
```diff
@@ -45,88 +45,97 @@ function dqgmres{T <: Number}(A :: AbstractLinearOperator, b :: AbstractVector{T
...
+      @kaxpy!(n, -H[diag], V[:,ipos], w) # w ← w - hᵢ.ₘ * vᵢ
     end
-    # jpos = iter-(iter+1)+2 = 1
-    H[iter,1] = @knrm2(n, w)
-    if H[iter,1] ≉ 0 # if H[iter,1] ≈ 0 => "lucky breakdown"
-      V[:,next_pos] = w / H[iter,1]
+    # Compute hₘ₊₁.ₘ and vₘ₊₁.
+    H[1] = @knrm2(n, w) # hₘ₊₁.ₘ = ‖vₘ₊₁‖
+    if H[1] ≉ 0 # hₘ₊₁.ₘ ≈ 0 ⇒ "lucky breakdown"
+      V[:,next_pos] = w / H[1] # vₘ₊₁ = w / hₘ₊₁.ₘ
```
dpo: Try using `@.` in front of vector updates so they are performed in place (the vector on the left-hand side must have been preallocated). For example, `@. V[:, next_pos] .= w / H[1]`. Otherwise, I believe Julia make copies and temporary vectors.

src/dqgmres.jl
```diff
@@ -45,88 +45,97 @@ function dqgmres{T <: Number}(A :: AbstractLinearOperator, b :: AbstractVector{T
...
-      @kaxpy!(n, -H[i,jpos], P[:,ipos], z) # z = z - H[i,jpos] * P[:,ipos]
+      ipos = mod(i-1, mem) + 1 # Position corresponding to pᵢ in the circular stack P.
+      diag = iter - i + 2
+      # z ← z - hᵢ.ₘ * pᵢ
+      @kaxpy!(n, -H[diag], P[:,ipos], z)
     end
-    P[:,pos] = z / H[iter,2]
-    @kaxpy!(n, g[pos], P[:,pos], x) # x = x + g[pos] * P[:,pos]
+    # pₘ = z / hₘ.ₘ
+    P[:,pos] = z / H[2]
```
dpo: `@.`

src/dqgmres.jl
Xx-fractale-xX: I tried it and I didn't see any difference in benchmarks.
I also tried `@view` but `BLAS.dot` or `BLAS.kaxpy!` don't accept `SubArray` as parameters....

dpo: > BLAS.dot or BLAS.kaxpy! don't accept SubArray as parameters

Right, the BLAS functions will only accept vectors, but we could modify `kaxpy!` so it takes `AbstractVector`s instead of `Vector`s.

Xx-fractale-xX:
I did the modifications and ran the same benchmark:
```julia
julia> @benchmark dqgmres($A,$b,memory=$mem)
BenchmarkTools.Trial: 
  memory estimate:  41.59 MiB
  allocs estimate:  1035631
  --------------
  minimum time:     1.048 s (0.41% GC)
  median time:      1.066 s (0.40% GC)
  mean time:        1.090 s (0.36% GC)
  maximum time:     1.154 s (0.19% GC)
  --------------
  samples:          5
  evals/sample:     1
```
We still have a lot of allocations but it's only take 41.59Mb, it's maybe what need Julia for running the algorithm?

Xx-fractale-xX:
I did the same optimization for `DIOM`.
```julia
julia> @benchmark diom($A,$b,memory=$50)

BenchmarkTools.Trial: 
  memory estimate:  2.36 GiB
  allocs estimate:  1323923
  --------------
  minimum time:     1.573 s (6.58% GC)
  median time:      1.604 s (6.33% GC)
  mean time:        1.612 s (6.52% GC)
  maximum time:     1.669 s (7.13% GC)
  --------------
  samples:          4
  evals/sample:     1

julia> @benchmark diom($A,$b,memory=$50)

BenchmarkTools.Trial: 
  memory estimate:  41.61 MiB
  allocs estimate:  1035650
  --------------
  minimum time:     1.168 s (0.32% GC)
  median time:      1.175 s (0.25% GC)
  mean time:        1.209 s (0.23% GC)
  maximum time:     1.285 s (0.11% GC)
  --------------
  samples:          5
  evals/sample:     1

```

Xx-fractale-xX:
I used ProfileView to verify that allocations are not a consequence of type inference. And it's not seem to be the case.
profileview

dpo:
> it's maybe what need Julia for running the algorithm?

Based on the number of nonzeros in A and the number of vectors needed for each method, we should be able to figure out more or less what the memory requirement should be.

Xx-fractale-xX:
The allocations come from the `view` function. With `julia 0.6`, a `view` needs to do  26 allocations. With the different loops at each iteration we have to do a lot of small allocations.
They improved it with `julia 1.0`, only  2 allocations are needed for a `view`.
I found an explanation here : https://stackoverflow.com/questions/47590839/unexpected-memory-allocation-when-using-array-views-julia/47607539#47607539

To summarize :
- When we want to use a slice of A matrix => `view(A,:,j)` or `view(A,i,:)`
- Update a vector by a combination of other vectors => `@.`

`@views` need 2 more allocations than `@view`, it's why I changed it too .

dpo:
Great, thank you! Would you mind running the benchmark again? What's the size of the matrix you're using, and how many nonzeros does it have?

Xx-fractale-xX:
I create a `tests.jl` file where I run `DQGMRES` and `DIOM` for different size of problem `n` and different memory value `mem`. A is a full matrix with `n²` nonzeros.

```julia
using BenchmarkTools
include("src/Krylov.jl")
using Krylov

# Compilation
diom(rand(10,10),rand(10))
dqgmres(rand(10,10),rand(10))

# Tests
for n in (500, 1000,2000)
	A = rand(n,n)
	b = rand(n)
	for mem in (2, 10, 20, 50)
		println("Test with n=",n," and mem=",mem)
		println("DIOM:")
		@btime diom($A,$b,memory=$mem,itmax=$n)
		println("DQGMRES:")
		@btime dqgmres($A,$b,memory=$mem,itmax=$n)
		println()
	end
end
```
The results are:
```julia
julia> include("tests.jl")
Test with n=500 and mem=2
DIOM:
  21.642 ms (18534 allocations: 2.51 MiB)
DQGMRES:
  23.946 ms (18529 allocations: 2.50 MiB)

Test with n=500 and mem=10
DIOM:
  27.918 ms (46212 allocations: 3.35 MiB)
DQGMRES:
  27.725 ms (46207 allocations: 3.35 MiB)

Test with n=500 and mem=20
DIOM:
  38.693 ms (80177 allocations: 4.39 MiB)
DQGMRES:
  35.614 ms (80172 allocations: 4.39 MiB)

Test with n=500 and mem=50
DIOM:
  60.482 ms (177872 allocations: 7.39 MiB)
DQGMRES:
  65.220 ms (177867 allocations: 7.39 MiB)

Test with n=1000 and mem=2
DIOM:
  407.434 ms (44038 allocations: 8.91 MiB)
DQGMRES:
  432.256 ms (44033 allocations: 8.90 MiB)

Test with n=1000 and mem=10
DIOM:
  418.585 ms (123576 allocations: 10.97 MiB)
DQGMRES:
  432.456 ms (123571 allocations: 10.96 MiB)

Test with n=1000 and mem=20
DIOM:
  470.501 ms (222096 allocations: 13.53 MiB)
DQGMRES:
  517.749 ms (222091 allocations: 13.52 MiB)

Test with n=1000 and mem=50
DIOM:
  531.387 ms (511656 allocations: 21.06 MiB)
DQGMRES:
  535.668 ms (511651 allocations: 21.05 MiB)
```

dpo:
In DQGMRES, we need
* mem n-vectors for V
* 4 n-vectors for x, r, Av and Mv
* mem n-vectors for P
* 2 mem-vectors for c and s
* 1 (mem+2)-vector for H.
So for your values of mem, we technically require:
```julia
julia> storage(mem, n) = mem*n + 4*n + mem*n + 2*mem + mem + 2

julia> for n in (500, 1000, 2000)
    for mem in (2, 10, 20, 50)
        @printf("%4d  %2d  %.2f MiB\n", n, mem, storage(mem, n) * 8 / 1024 / 1024)
    end
 end

 500   2  0.03 MiB
 500  10  0.09 MiB
 500  20  0.17 MiB
 500  50  0.40 MiB
1000   2  0.06 MiB
1000  10  0.18 MiB
1000  20  0.34 MiB
1000  50  0.79 MiB
2000   2  0.12 MiB
2000  10  0.37 MiB
2000  20  0.67 MiB
2000  50  1.59 MiB
```
I just realized that the default preconditioner, `opEye`, is allocating. We should define a non-allocating identity somewhere.
If we take the n allocations due to `opeEye` into account, we get
```julia
julia> storage(mem, n) = mem*n + 3*n + mem*n + 2*mem + mem + 2 + n * n

 500   2  1.93 MiB
 500  10  2.00 MiB
 500  20  2.07 MiB
 500  50  2.30 MiB
1000   2  7.68 MiB
1000  10  7.81 MiB
1000  20  7.96 MiB
1000  50  8.42 MiB
2000   2  30.62 MiB
2000  10  30.87 MiB
2000  20  31.17 MiB
2000  50  32.09 MiB
```
which is already closer to what you report. Still, something else is going on.

Xx-fractale-xX:
I finally found the problem... The problem comes from :
```julia
z = M * view(V,:,pos)
w = A * z
```
The proper way for updating `w` and `z` is
```julia
z[:] = M * view(V,:,pos)
w[:] = A * z
```
Otherwise Julia do a lot of allocations here.

This time I ran tests with preallocated preconditioners:
```julia
using BenchmarkTools
using LinearOperators
include("src/Krylov.jl")
using Krylov

function preallocated_LinearOperator(A)
  (n, m) = size(A)
  Ap = zeros(n)
  Atq = zeros(m)
  return LinearOperator(n, m, false, false, p -> A_mul_B!(Ap, A, p), q -> At_mul_B!(Atq, A, q), q -> At_mul_B!(Atq, A, q))
end

# Compilation
diom(rand(10,10),rand(10))
dqgmres(rand(10,10),rand(10))

# Tests
for n in (500, 1000,2000)
	AA = rand(n,n)
	A = preallocated_LinearOperator(AA)
	MM = eye(n)
	M = preallocated_LinearOperator(MM)
	b = rand(n)
	for mem in (2, 10, 20, 50)
		println("Test with n=",n," and mem=",mem)
		println("DIOM:")
		@btime diom($A,$b,memory=$mem,itmax=$n,M=$M)
		println("DQGMRES:")
		@btime dqgmres($A,$b,memory=$mem,itmax=$n,M=$M)
	end
end
```

The results are : 
```julia
Test with n=500 and mem=2
DIOM:
  75.538 ms (4024 allocations: 232.56 KiB)
DQGMRES:
  76.555 ms (4021 allocations: 228.45 KiB)
Test with n=500 and mem=10
DIOM:
  80.054 ms (15886 allocations: 851.03 KiB)
DQGMRES:
  79.430 ms (15883 allocations: 846.98 KiB)
Test with n=500 and mem=20
DIOM:
  83.818 ms (30441 allocations: 1.57 MiB)
DQGMRES:
  86.960 ms (30438 allocations: 1.57 MiB)
Test with n=500 and mem=50
DIOM:
  98.033 ms (72306 allocations: 3.72 MiB)
DQGMRES:
  99.902 ms (72303 allocations: 3.72 MiB)
Test with n=1000 and mem=2
DIOM:
  873.862 ms (8027 allocations: 463.16 KiB)
DQGMRES:
  799.853 ms (8024 allocations: 455.17 KiB)
Test with n=1000 and mem=10
DIOM:
  870.960 ms (31889 allocations: 1.67 MiB)
DQGMRES:
  1.393 s (31886 allocations: 1.66 MiB)
Test with n=1000 and mem=20
DIOM:
  1.029 s (61444 allocations: 3.17 MiB)
DQGMRES:
  955.647 ms (61441 allocations: 3.16 MiB)
Test with n=1000 and mem=50
DIOM:
  887.362 ms (148309 allocations: 7.61 MiB)
DQGMRES:
  911.540 ms (148306 allocations: 7.60 MiB)
Test with n=2000 and mem=2
DIOM:
  7.846 s (16030 allocations: 924.44 KiB)
DQGMRES:
  7.389 s (16027 allocations: 908.64 KiB)
Test with n=2000 and mem=10
DIOM:
  8.046 s (63890 allocations: 3.34 MiB)
DQGMRES:
  7.126 s (63887 allocations: 3.32 MiB)
Test with n=2000 and mem=20
DIOM:
  8.243 s (123445 allocations: 6.37 MiB)
DQGMRES:
  6.642 s (123442 allocations: 6.35 MiB)
Test with n=2000 and mem=50
DIOM:
  6.862 s (300310 allocations: 15.38 MiB)
DQGMRES:
  6.975 s (300307 allocations: 15.37 MiB)
```
The other element to take into account is the `48 bytes` that need Julia to allocated for a `view`.
In `DQGMRES` we need to do `3*mem + 2` views per iteration (when iter ≥ mem).
```julia
# Each number is a Float64 and need 8 bytes.
storage_data(mem, n) = mem*n + 4*n + mem*n + 2*mem + mem + 2
# Each view need 48 bytes.
storage_view(mem, n) = (3*mem + 2) * n
# Cumulative storage in MB.
total_storage(mem, n) = (48 * storage_view(mem, n) + 8 * storage_data(mem, n)) / (1024 * 1024) 
for n in (500, 1000, 2000)
    for mem in (2, 10, 20, 50)
        @printf("%4d  %2d  %.2f MiB\n", n, mem, total_storage(mem, n))
    end
end

 500   2  0.21 MiB
 500  10  0.82 MiB
 500  20  1.59 MiB
 500  50  3.88 MiB
1000   2  0.43 MiB
1000  10  1.65 MiB
1000  20  3.17 MiB
1000  50  7.75 MiB
2000   2  0.85 MiB
2000  10  3.30 MiB
2000  20  6.35 MiB
2000  50  15.50 MiB
```
The number of allocation is also right, we have `(3*mem + 2)n` allocations for each view and less than `30` allocations for the initialization of the others variables.

Xx-fractale-xX:
The `unsafe_view` function from `ArrayViews` doesn't allocated memory. But It's only compatible with julia 0.7 and 1.0 for the moment.

dpo:
That is a huge improvement in terms of allocations, but the run times have increased. Do you have any idea why?

In fact, I'm a bit surprised because
> The proper way for updating w and z is
```
z[:] = M * view(V,:,pos)
w[:] = A * z
```
is going to copy the result of, say, `A * z` (i.e., the contents of the preallocated vector `Ap` into `w`. I thought `w = A * z` would simply be a pointer operation. Are you saying it's not the case?

Here's the right way to do it:
```julia
julia> using FastClosures
julia> function preallocated_LinearOperator(A)
         (n, m) = size(A)
         Ap = zeros(n); prod = @closure p -> mul!(Ap, A, p);
         Atq = zeros(m); tprod = @closure q -> mul!(Atq, transpose(A), q);
         return LinearOperator{eltype(A),typeof(prod),typeof(tprod),typeof(tprod)}(n, m, false, false, prod, tprod, tprod)
       end

julia> function testprod(A)
       y = rand(size(A,1))
       z = A * y
       return z
       end

julia> n = 500;
julia> AA = rand(n,n);
julia> A = preallocated_LinearOperator(AA);

julia> @btime testprod(A);
  23.092 μs (1 allocation: 4.06 KiB)
```
Now it's really non-allocating because 500 * 8 / 1024 = 3.90625 KiB.

Xx-fractale-xX:
```I thought w = A * z would simply be a pointer operation. Are you saying it's not the case?```

With Julia 0.6 it's not the case but I don't understand why. I tested  `preallocated_LinearOperator` with `@closure` and had the same problem.

But I also tested it with `julia-0.7` and we don't have this problem (even without `@closure` in `preallocated_LinearOperator`. So I suppose that it's a bug with `julia 0.6`....

I think it's the good time for moving to julia 0.7 and 1.0.

dpo:
I agree, thank you. After the transition to 0.7 and 1.0, we'll have to update the definition of preallocated operators to use `@closure`.

Xx-fractale-xX:
I did a last commit to keep the classic syntax `w = A * z`. 
Maybe you want to merge it before @abelsiqueira rebase #65?

PR #63, opened 2018-09-27
An implementation of DIOM

Xx-fractale-xX:
Little brother of `DQGMRES`.

Currently, the factorization is `H` is done without pivoting.
Like `DQGMRES`, we can avoid the storage of H and I found a way to preconditioned it too. 
This algorithm needs `2n + 1` vectors of `mem` coefficients.

### DIFF
--- a/src/diom.jl
+++ b/src/diom.jl
@@ -156,7 +156,7 @@ function diom{T <: Number}(A :: AbstractLinearOperator, b :: AbstractVector{T};
 
     # Update x_old and residual norm.
     if !p[next_pos]
-      x_old = copy(x)
+      copy!(x_old, x)
       # ‖ b - Axₘ ‖ = hₘ₊₁.ₘ * |ξₘ / uₘ.ₘ| without pivoting
       rNorm = H[1] * abs(ξ / H[2])
     else
--- a/src/diom.jl
+++ b/src/diom.jl
@@ -94,24 +94,20 @@ function diom{T <: Number}(A :: AbstractLinearOperator, b :: AbstractVector{T};
       V[:,next_pos] = w / H[1] # vₘ₊₁ = w / hₘ₊₁.ₘ
     end
 
-    if iter == 1
-      # l₂.₁ ← h₂.₁ / u₁.₁
-      L[next_pos] = H[1] / H[2]
-    else
-      # Update LU factorization of Hₘ by computing the last row of Lₘ and the next pivot lₘ₊₁.ₘ of Lₘ.
+    # Update LU factorization of Hₘ by computing the last column of Uₘ and the next pivot lₘ₊₁.ₘ of Lₘ.
+    if iter ≥ 2
       # uᵢ.ₘ ← hᵢ.ₘ - lᵢ.ᵢ₋₁ * uᵢ₋₁.ₘ
       for i = max(2,iter-mem+2) : iter
         lpos = mod(i-1, mem) + 1 # Position of lᵢ.ᵢ₋₁ in the circular stack
         jpos = iter - i + 2
         H[jpos] = H[jpos] - L[lpos] * H[jpos+1]
       end
-      # lₘ₊₁.ₘ ← hₘ₊₁.ₘ / uₘ.ₘ
-      L[next_pos] = H[1] / H[2]
-
       # Compute ξₘ the last composant of zₘ = β(Lₘ)⁻¹e₁.
       # ξₘ = -lₘ.ₘ₋₁ * ξₘ₋₁
       ξ = - L[pos] * ξ
     end
+    # lₘ₊₁.ₘ ← hₘ₊₁.ₘ / uₘ.ₘ
+    L[next_pos] = H[1] / H[2]
 
     # Compute the last column of Pₘ = Vₘ(Uₘ)⁻¹.
     for i = max(1,iter-mem+1) : iter - 1
--- a/src/diom.jl
+++ b/src/diom.jl
@@ -6,8 +6,7 @@
 # PWS Publishing Company, Boston, USA, 1996.
 #
 # Y. Saad, Practical use of some krylov subspace methods for solving indefinite and nonsymmetric linear systems.
-# SIAM journal on scientific and statistical computing.
-# Vol. 5, No. 1, March 1984.
+# SIAM journal on scientific and statistical computing, 5(1), pp. 203--228, 1984.
 # 
 # Alexis Montoison, <alexis.montoison@polymtl.ca>
 # Montreal, September 2018.
--- a/src/diom.jl
+++ b/src/diom.jl
@@ -54,16 +54,16 @@ function diom{T <: Number}(A :: AbstractLinearOperator, b :: AbstractVector{T};
 
   # Set up workspace.
   mem = min(memory, itmax) # Memory.
-  P = zeros(n, mem) # Directions for x.
-  V = zeros(n, mem) # Preconditioned Krylov vectors.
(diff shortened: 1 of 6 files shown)


Xx-fractale-xX:
I didn't passed tests on julia 0.7:
`ERROR: LoadError: LoadError: UndefVarError: A_mul_B! not defined`

I don't understand why, It works on my laptop...
I tested it with `include(Krylov.jl")` and using `Main.Krylov`.

dpo:
Possibly because this package hasn't yet been upgraded to 0.7 and so doesn't pull the correct version of LinearOperators?! You're probably using a recent enough version on your laptop. First we need to upgrade to 0.7 (in another PR).

Xx-fractale-xX:
I found the problem, @abel renamed the function `A_mul_B!` by `mul!` with the version `5.0` of `LinearOperators`. 
And we need `A_mul_B!` to preallocated LinearOperators in `variants.jl`.
I think it's preferable to phase out julia 0.6 and modify `variants.jl`. But we can also add `LinearOperators 4.1` in the `REQUIRE`.

abelsiqueira:
`A_mul_B` and similar were removed from 1.0 [link](https://github.com/JuliaLang/julia/issues/5332) and `mul!` with dispatch is the new way to handle it. `LinearOperators` v0.4.1 still uses `A_mul_B` and will only work on 0.6/0.7.
There should be a last release limiting LinearOperators on REQUIRE, then dropping 0.6.

Xx-fractale-xX:
LU factorization with partial pivoting don't break if H is singular. Furthermore the process is more stable, the biggest pivot is chosen at each step.

I'm quite sure that this version was never implemented. In the article ```Practical use of some Krylov subspace methods for solving indefinite and nonsymmetric linear systems```, Youcef Saad talked about this variant but he made a mistake (from my point of view) in the update of x. I finally found an other way to do it which seems correct for me.

test/test_diom.jl
```diff
@@ -0,0 +1,71 @@
+include("get_div_grad.jl")
```
dpo: DIOM and DQGMRES should really be tested on unsymmetric systems as well.

Xx-fractale-xX:
I also updated the reference of the article ```Practical use of some krylov subspace methods for solving indefinite and nonsymmetric linear systems```.

src/diom.jl
```diff
@@ -0,0 +1,177 @@
...
+# SIAM journal on scientific and statistical computing, 5(1), pp. 203--228, 1984.
+# 
+# Alexis Montoison, <alexis.montoison@polymtl.ca>
+# Montreal, September 2018.
+
+export diom
+
+"""Solve the consistent linear system Ax = b using direct incomplete orthogonalization method.
+
+DIOM is similar to SYMMLQ for nonsymmetric problems.
```
dpo: I don't think this is accurate. FOM is to the Arnoldi process as CG is to the Lanczos process. I think it would be more accurate to say that DIOM is similar to CG with partial reorthogonalization.

src/diom.jl
```diff
@@ -0,0 +1,177 @@
...
+It's a more economical algorithm based upon the LU factorization with partial pivoting.
+
+In the particular case where A is symmetric indefinite, DIOM is theorecally equivalent 
+to SYMMLQ but slightly more economical.
```
dpo: I don't think this is true either; DIOM and SYMMLQ do not minimize the same quantities.

src/diom.jl
```diff
@@ -0,0 +1,177 @@
...
+
+    # Update iteration index.
+    iter = iter + 1
+
+    # Set position in circulars stacks.
+    pos = mod(iter-1, mem) + 1 # Position corresponding to pₘ and vₘ in circular stacks P and V.
+    next_pos = mod(iter, mem) + 1 # Position corresponding to vₘ₊₁ in the circular stack V.
+
+    # Incomplete Arnoldi procedure.
+    z = M * V[:,pos] # Forms pₘ
```
dpo: I wonder why Github isn't rendering UTF-8 characters correctly. Is your editor set to use UTF-8 encoding?

dpo:
Great, thank you!

src/diom.jl
Xx-fractale-xX: Yes, my editor is set to use UTF-8 encoding. Github is rendering UTF-8 correctly for me, maybe it's your web browser?
encodage

src/diom.jl
Xx-fractale-xX: Yes, it's not norm(x) that is minimized. I think we could delete this sentence?

src/diom.jl
Xx-fractale-xX: Ok, I will rewrite the docString.

src/diom.jl
dpo: Both Safari and Chrome seem to be missing characters...

src/diom.jl
```diff
@@ -0,0 +1,173 @@
...
+  # Set up workspace.
+  mem = min(memory, itmax) # Memory.
+  V = zeros(n, mem) # Preconditioned Krylov vectors, orthogonal basis for {r₀, AM⁻¹r₀, (AM⁻¹)²r₀, ..., (AM⁻¹)ᵐ⁻¹r₀}.
+  P = zeros(n, mem) # Directions for x : Pₘ = Vₘ(Uₘ)⁻¹.
+  H = zeros(mem+2)  # Last column of the band hessenberg matrix Hₘ = LₘUₘ.
+  # Each column has at most mem + 1 nonzero elements. hᵢ.ₘ is stored as H[m-i+2].
+  # m-i+2 represents the indice of the diagonal where hᵢ.ₘ is located.
+  # In addition of that, the last column of Uₘ is stored in H.
+  L = zeros(mem) # Last mem Pivots of Lₘ.
+  p = zeros(Bool, mem) # Last mem permutations.
```
dpo: Maybe `p` could be a `BitArray`?

src/diom.jl
```diff
@@ -0,0 +1,173 @@
...
+    w = A * z # Forms vₘ₊₁
+    for i = max(1, iter-mem+1) : iter
+      ipos = mod(i-1, mem) + 1 # Position corresponding to vᵢ in the circular stack V.
+      diag = iter - i + 2
+      H[diag] = @kdot(n, w, V[:,ipos]) # hᵢ.ₘ = < A * vₘ , vᵢ >
+      @kaxpy!(n, -H[diag], V[:,ipos], w) # w ← w - hᵢ.ₘ * vᵢ
+    end
+    # Compute hₘ₊₁.ₘ and vₘ₊₁.
+    H[1] = @knrm2(n, w) # hₘ₊₁.ₘ = ‖vₘ₊₁‖
+    if H[1] ≉ 0 # hₘ₊₁.ₘ ≈ 0 ⇒ "lucky breakdown"
```
dpo: Isn't there a tolerance involved in `≉`? In my experience, we want to test for `!= 0`.

src/diom.jl
Xx-fractale-xX: Yes, I didn't know that a `Bool` was an `Uint8` in Julia and not a single `bit`.

src/diom.jl
Xx-fractale-xX: Yes, there is a tolerance involved in `≈`, `x ≈ y` return true if `norm(x-y) ≤ rtol * |max(x,y)|`. With rtol that depend of the type of x and y. It's `sqrt(eps)` with eps the precision machine of Float16, Float32, Float64....

But when we do `x ≈ 0`, It's mean that `|x| ≤ rtol * |x| ↔ x = 0`. I used this notation because I find it nicer than `==`. Also I like this notation because it means that x is a zero for the computer in the working precision but it's possible that it's not a zero in another precision or in exact arithmetic.

dpo: OK.

dpo:
Thank you!

PR #64, opened 2018-09-28
Update REQUIRE

abelsiqueira:
This could be the last 0.6 release, and then we drop 0.6, and then the PRs are rebased, or; PRs are rebased after this, then we make the last release and drop 0.6.


abelsiqueira:
This just limits LinearOperators and drops compat, so I'm merging to test the other PRs.

PR #69, opened 2018-10-18
An implementation of CGS

Xx-fractale-xX:
Classic version of CGS with right preconditioning.
cgs

### DIFF
--- a/src/cgs.jl
+++ b/src/cgs.jl
@@ -1,0 +1,109 @@
+# An implementation of CGS for the solution of the square linear system Ax = b.
+#
+# This method is described in
+#
+# Y. Saad, Iterative methods for sparse linear systems.
+# PWS Publishing Company, Boston, USA, 1996.
+#
+# P. Sonneveld, CGS, A Fast Lanczos-Type Solver for Nonsymmetric Linear systems.
+# SIAM Journal on Scientific and Statistical Computing, 10(1), pp. 36--52, 1989.
+# 
+# Alexis Montoison, <alexis.montoison@polymtl.ca>
+# Montreal, October 2018.
+
+export cgs
+
+"""Solve the consistent linear system Ax = b using conjugate gradient squared algorithm.
+
+From "Iterative Methods for Sparse Linear Systems (Y. Saad)" :
+
+«The method is based on a polynomial variant of the conjugate gradients algorithm.
+Although related to the so-called bi-conjugate gradients (BCG) algorithm,
+it does not involve adjoint matrix-vector multiplications, and the expected convergence
+rate is about twice that of the BCG algorithm.
+
+The Conjugate Gradient Squared algorithm works quite well in many cases.
+However, one difficulty is that, since the polynomials are squared, rounding errors
+tend to be more damaging than in the standard BCG algorithm. In particular, very
+high variations of the residual vectors often cause the residual norms computed 
+to become inaccurate.
+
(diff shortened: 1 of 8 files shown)


src/cgs.jl
```diff
@@ -0,0 +1,120 @@
...
+it does not involve adjoint matrix-vector multiplications, and the expected convergence
+rate is about twice that of the BCG algorithm.
+
+The Conjugate Gradient Squared algorithm works quite well in many cases.
+However, one difficulty is that, since the polynomials are squared, rounding errors
+tend to be more damaging than in the standard BCG algorithm. In particular, very
+high variations of the residual vectors often cause the residual norms computed 
+to become inaccurate. 
+
+TFQMR and BICGSTAB were developed to remedy this difficulty.
```
dpo: Is this description copy-pasted from a paper or a book?

src/cgs.jl
```diff
@@ -0,0 +1,120 @@
...
+  p = copy(r₀) # p₀
+
+  # Stopping criterion.
+  solved = rNorm ≤ ε
+  tired = iter ≥ itmax
+  status = "unknown"
+
+  while !(solved || tired)
+
+    # yₘ = M⁻¹pₘ
```
dpo: Would you mind placing the comments to the right of the statements to improve readability?

test/test_cgs.jl
```diff
@@ -0,0 +1,61 @@
+include("get_div_grad.jl")
```
dpo: I think all these examples are symmetric, right? We should test CGS on unsymmetric systems.

src/cgs.jl
Xx-fractale-xX: Yes, I took it in ```Iterative methods for sparse linear systems```.

dpo: Then there should be a sentence to cite the source. For example, you could start the docstring with `From "Iterative Methods for Sparse Linear Systems (C. T. Kelley):" ...`

src/cgs.jl
Xx-fractale-xX: Ok, I do the modifications.

test/test_cgs.jl
Xx-fractale-xX: Yes, I took the same tests that I used for ```DQGMRES``` and ```DIOM```. I will find two new tests, a not symmetric definite positive matrix and an other one not symmetric and indefinite.

src/cgs.jl
Xx-fractale-xX: Ok, I modify that too. Thanks!

dpo:
This branch needs to be rebased now.

src/cgs.jl
```diff
@@ -0,0 +1,109 @@
...
+
+The Conjugate Gradient Squared algorithm works quite well in many cases.
+However, one difficulty is that, since the polynomials are squared, rounding errors
+tend to be more damaging than in the standard BCG algorithm. In particular, very
+high variations of the residual vectors often cause the residual norms computed 
+to become inaccurate.
+
+TFQMR and BICGSTAB were developed to remedy this difficulty.»
+
+This implementation allows a right preconditioning with M.
```
dpo: ...allows a right preconditioner M.

dpo:
Looks great, thank you!

src/cgs.jl
Xx-fractale-xX: Thanks! I also updated `DIOM` and `DQGMRES`, I made the same mistake.

PR #65, opened 2018-10-01
Drop 0.6

abelsiqueira:
Don't merge yet, waiting for #64.

PR #72, opened 2018-10-26
updates for julia 0.7

dpo:
Supersedes #65.

@Xx-fractale-xX The CRMR test causes NaNs. I commented it out.

### DIFF
--- a/test/test_cgs.jl
+++ b/test/test_cgs.jl
@@ -4,7 +4,7 @@ cgs_tol = 1.0e-6
 
 # Symmetric and positive definite systems (cubic spline matrix).
 n = 10
-A = spdiagm((ones(n-1), 4*ones(n), ones(n-1)), (-1, 0, 1))
+A = spdiagm(-1 => ones(n-1), 0 => 4*ones(n), 1 => ones(n-1))
 b = A * [1:n;]
 (x, stats) = cgs(A, b)
 r = b - A * x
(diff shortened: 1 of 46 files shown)


test/test_cgs.jl
```diff
@@ -44,7 +44,7 @@ resid = norm(r) / norm(b)
 @test(stats.solved)
 
 # Code coverage.
-(x, stats) = cgs(full(A), b)
+(x, stats) = cgs(Array(A), b)
```
abelsiqueira: A is already an array here, so I think this is not required.

test/test_diom.jl
```diff
@@ -44,7 +44,7 @@ resid = norm(r) / norm(b)
 @test(stats.solved)
 
 # Code coverage.
-(x, stats) = diom(full(A), b)
+(x, stats) = diom(Array(A), b)
```
abelsiqueira: Here as well

abelsiqueira:
Thanks for updating #65. I found these two `Array(A)`, which may be from a previous version, because I think it's not needed for code coverage anymore.

abelsiqueira:
Should we release a 0.6 version, before merging this?

dpo:
> Should we release a 0.6 version, before merging this?

We could, but I'm not sure we have to. There's never been a release of Krylov.jl and I think we should go over each solver and check memory allocations (as in #62). I'll have a PR after this one to illustrate how for a couple of solvers. I'm also in the process of adding benchmarks (with `PkgBenchmark`).

dpo:
Fixed.

abelsiqueira:
Thanks

PR #74, opened 2018-10-27
Benchmarks

dpo:
Here are initial benchmarks using `PkgBenchmark`. You can run the benchmarks from the REPL using
```julia
julia> using PkgBenchmark
julia> bmark = PkgBenchmark.benchmarkpkg("Krylov")
julia> export_markdown(stdout, bmark)
```
The benchmarks can also be exported to a gist, etc. The important point is that we can compare the impact of each pull request. It would be great if the process could be automated somehow.

I also updated the simple benchmark in `cg_bmark.jl`. It seems that KrylovMethods is currently broken, so here are the current results against IterativeSolvers:
```
N = 32
Krylov
  18.856 ms (16 allocations: 770.67 KiB)
IterativeSolvers
  19.477 ms (23 allocations: 1.25 MiB)
N = 64
Krylov
  380.921 ms (17 allocations: 6.00 MiB)
IterativeSolvers
  393.178 ms (23 allocations: 10.00 MiB)
N = 128
Krylov
  7.791 s (18 allocations: 48.01 MiB)
IterativeSolvers
  7.961 s (23 allocations: 80.00 MiB)
```
They are quite favorable, especially in terms of allocations.


dpo:
Some updates based on scripts by @Xx-fractale-xX. Thanks!

Xx-fractale-xX:
Looks good! Just a minor comment, you could add a `REQUIRE` in the `benchmark` folder that contains all the dependencies. (`MatrixMarket`,`IterativeSolvers`,`MatrixDepot`,`FastClosures`,...).

If this packages aren't installed when we do a  `PkgBenchmark.benchmarkpkg("Krylov")`, Julia will install them for the bencharks and delete them at the end.

PR #75, opened 2018-10-28
Allocations tests for CR

Xx-fractale-xX:


dpo:
There's a lot more going on here than just the allocation tests (changes in comments, semicolons, etc.). Please keep each pull request focused on a single topic. Please update this one so it contains only the allocation tests. It would be great to submit another one to remove the (old) semicolons. Thanks for all the pull requests!

test/test_alloc.jl
```diff
@@ -21,3 +30,12 @@ expected_cg_lanczos_bytes = storage_cg_lanczos_bytes(n)
 cg_lanczos(A, b)  # warmup
 actual_cg_lanczos_bytes = @allocated cg_lanczos(A, b)
 @test actual_cg_lanczos_bytes ≤ 1.1 * expected_cg_lanczos_bytes
+
+# without preconditioner and with (Ap, Aᵗp) preallocated, CGNE needs 3 n-vectors: x, r, p
```
dpo: This is a bit confusing because CGNE is for under-determined systems. The requirements should be two m-vectors and one n-vector if A is m-by-n, no?

Xx-fractale-xX: Yes, I forgot that CGNE is for under-determined systems. I will correct that. Thank you.

Xx-fractale-xX: I will add CGNE allocations in an other pull request and keep this one focus with CR allocations.

dpo:
Great, thank you!

Issue #52, opened 2018-06-20
Preconditioners


dpo:
We should add preconditioners in the obvious places (e.g., `cg()`).

# Preconditioners roadmap

- [x] CG
- [x] CG_LANCZOS
- [ ] CG_LANCZOS_SHIFT_SEQ
- [x] CR
- [x] MINRES
- [x] SYMMLQ
- [x] CGLS
- [x] CGNE
- [x] CRAIG
- [x] CRAIGMR
- [x] CRMR
- [x] LSLQ
- [x] LSMR
- [x] LSQR
- [x] CRLS

Xx-fractale-xX:
# Preconditioners roadmap

- [ ] CG
- [ ] CG_LANCZOS
- [ ] CG_LANCZOS_SHIFT_SEQ
- [ ] CR
- [x] MINRES
- [x] SYMMLQ
- [x] CGLS
- [ ] CGNE
- [ ] CRAIG
- [ ] CRAIGMR
- [ ] CRMR
- [x] LSLQ
- [x] LSMR
- [x] LSQR
- [x] CRLS

Xx-fractale-xX:
As issue #45, we should use preallocation for preconditioners. I don't know if we can update `variants.jl` this time, `M` is a keyword argument.
Maybe we could move `preallocated_LinearOperator` and `preallocated_symmetric_LinearOperator` functions in `LinearOperators`?

dpo:
Yes, we should certainly have preallocated operators in LinearOperators.

Xx-fractale-xX:
Almost finished! `CRAIG` and `CRAIGMR` are the last ones.

PR #76, opened 2018-10-29
Preconditioned CG_LANCZOS

amontoison:
#52 
I preconditioned those methods the same way that I did in `DQGMRES` in `DIOM`.  Lanczos orthogonalization is just a special case of Arnoldi orthogonalization, so we can add a preconditioner the same way.

However I don't really understood how we have the relations for updating `p` and `x`. I only know how to derived relations with a LU factorization of `Tₘ` and  it's quite different.  I suppose that we modified them with the help of the relation `σⱼ² (δⱼ - ωⱼ₋₁ / γⱼ₋₁) = pⱼᵀ A pⱼ`.

### DIFF
--- a/src/cg_lanczos.jl
+++ b/src/cg_lanczos.jl
@@ -35,17 +35,17 @@ function cg_lanczos(A :: AbstractLinearOperator, b :: AbstractVector{T};
   MisI = isa(M, opEye)
 
   # Initial state.
-  x = zeros(T, n);
-  Mv = copy(b)
-  v = M * Mv
-  β = sqrt(@kdot(n, v, Mv))
+  x = zeros(T, n)           # x₀
+  Mv = copy(b)              # Mv₁ ← b
+  v = M * Mv                # v₁ = M⁻¹ * Mv₁
+  β = sqrt(@kdot(n, v, Mv)) # β₁ = v₁ᵀ M v₁
   β == 0 && return x, LanczosStats(true, [0.0], false, 0.0, 0.0, "x = 0 is a zero-residual solution")
   p = copy(v)
 
   # Initialize Lanczos process.
   # β₁Mv₁ = b
-  @kscal!(n, 1.0/β, v)
-  MisI || @kscal!(n, 1.0/β, Mv)
+  @kscal!(n, 1.0/β, v)          # v₁  ←  v₁ / β₁
+  MisI || @kscal!(n, 1.0/β, Mv) # Mv₁ ← Mv₁ / β₁
   Mv_prev = copy(Mv)
 
   iter = 0;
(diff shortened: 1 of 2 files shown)


src/cg_lanczos.jl
```diff
@@ -51,29 +54,30 @@ function cg_lanczos(A :: AbstractLinearOperator, b :: AbstractVector{T};
   verbose && @printf("%5d  %8.1e\n", iter, rNorm);
 
   indefinite = false;
-  solved = rNorm <= ε;
-  tired = iter >= itmax;
+  solved = rNorm ≤ ε;
+  tired = iter ≥ itmax;
```
dpo: Same comment here as in #75. Please keep a pull request to a single topic. It's great to submit another pull requests with syntactic updates. Otherwise, it takes me much longer to review a pull request like this one. Thank you.

Xx-fractale-xX: I cleaned each pull request and created new pull requests when it was about another topic. It will be easier to review now.

src/cg_lanczos.jl
```diff
@@ -18,10 +18,13 @@ symmetric linear system
   Ax = b
 
 The method does _not_ abort if A is not definite.
+
+This implementation allows a right preconditioner M.
```
dpo: A left preconditioner would be more interesting here. If *M* is symmetric and positive definite, we really want to do centered preconditioning, i.e., *M<sup>1/2</sup> A M<sup>1/2</sup> x = M<sup>1/2</sup> b*. When you write out CG_Lanczos for this system, you can transform it so only *M* is needed (not *M<sup>1/2</sup>*) and everything happens as if you were solving *MAx = Mb*.

If you look at standard CG with preconditioner, you'll see that the first thing we do is precondition the initial residual (i.e., *b*): https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/src/cg.jl#L33. CG_Lanczos and CG should be perfectly equivalent (on paper).

Xx-fractale-xX: I updated it, it's a right preconditioner now for both methods.

dpo: Right or left?

Xx-fractale-xX: Left, I made a mistake in my last message.

src/cg_lanczos.jl
```diff
@@ -18,22 +18,27 @@ symmetric linear system
...
 
   n = size(b, 1);
   (size(A, 1) == n & size(A, 2) == n) || error("Inconsistent problem size");
   verbose && @printf("CG Lanczos: system of %d equations in %d variables\n", n, n);
 
   # Initial state.
   x = zeros(T, n);
-  β = @knrm2(n, b)
+  r₀ = M * b
+  β = @knrm2(n, r₀)
```
dpo: I think this should be `@kdot(n, r₀, b)`.

src/cg_lanczos.jl
```diff
@@ -119,13 +128,14 @@ function cg_lanczos_shift_seq(A :: AbstractLinearOperator, b :: AbstractVector{T
   # Initial state.
   ## Distribute x similarly to shifts.
   x = zeros(Tb, n, nshifts);
-  β = @knrm2(n, b)
+  r₀ = M * b
+  β = @knrm2(n, r₀)
```
dpo: Same here.

src/cg_lanczos.jl
```diff
@@ -161,21 +171,29 @@ function cg_lanczos_shift_seq(A :: AbstractLinearOperator, b :: AbstractVector{T
   # Main loop.
   while ! (solved || tired)
     # Form next Lanczos vector.
-    v_next = A * v;
+    Mv = M * v
+    if Mv === v
+      ρ = 1
+    else
+      ρ = @kdot(n, v, Mv)
```
dpo: Won't `@kdot` return 1 when `Mv === v`?

src/cg_lanczos.jl
```diff
@@ -161,21 +171,29 @@ function cg_lanczos_shift_seq(A :: AbstractLinearOperator, b :: AbstractVector{T
...
+      @. v_prev = v
     end
     β = @knrm2(n, v_next)
-    v = v_next / β;
+    @. v = v_next / β;
 
-    # Check curvature: v'(A + sᵢI)v = v'Av + sᵢ ‖v‖² = δ + sᵢ because ‖v‖ = 1.
-    # It is possible to show that σⱼ² (δⱼ + sᵢ - ωⱼ₋₁ / γⱼ₋₁) = pⱼᵀ (A + sᵢ I) pⱼ.
+    # Check curvature: vᵗ M(A + sᵢI) v = vᵗMAv + sᵢ ‖M½v‖² = δ + ρ * sᵢ.
+    # With M = I, ρ = ‖v‖² = 1.
```
dpo: More generally, *v'Mv = ‖v‖<sub>M</sub>² = ‖M<sup>½</sup>v‖² = 1*.

src/cg_lanczos.jl
Xx-fractale-xX: We want that `βMv₁ = b with ‖v₁‖_M = 1`
`βv₁ = M⁻¹b` and by product `β²<v₁,Mv₁> = bᵗM⁻¹b`.
So `β = sqrt(bᵗM⁻¹b)`.

Ok, I see the problem, `β = sqrt(@kdot(n, r₀, b))`

dpo: You're right.

dpo:
This needs to be rebased.

amontoison:
With the new `opEye`, I will update this pull request next week.

amontoison:
It's seem that the expression of `‖rₖ‖₂` is different in `CG_LANCZOS_SHIFT_SEQ` with a preconditioner. I will add it later in an other pull request.

amontoison:
I was reading the original article `Fast CG-Based Methods for Tikhonov-Phillips Regularization` and I found some mistakes in my indices.
it should be correctly preconditioned and documented now.

dpo:
Thank you!

PR #77, opened 2018-10-30
Allocations tests for CGS, DIOM and DQGMRES

Xx-fractale-xX:


src/cgs.jl
```diff
@@ -69,22 +70,21 @@ function cgs(A :: AbstractLinearOperator, b :: AbstractVector{T};
...
-    @kscal!(n, β, p)         # pₘ₊₁ = uₘ₊₁ + βₘ * (qₘ + βₘ * pₘ)
-    @kaxpy!(n, 1.0, q, p)
-    @kscal!(n, β, p)
-    @kaxpy!(n, 1.0, u, p)
+    y = M * p                 # yₘ = M⁻¹pₘ
+    v = A * y                 # vₘ = Ayₘ
+    σ = @kdot(n, v, r₀)       # σₘ = < AM⁻¹pₘ,r₀ >
+    α = ρ / σ                 # αₘ = ρₘ / σₘ
+    @. q = u - α * v          # qₘ = uₘ - αₘ * AM⁻¹pₘ
+    @. u = u + q              # uₘ₊½ = uₘ + qₘ
```
dpo: This is a `@kaxpy` but I'm not sure if it's still faster than the native Julia operation.

Xx-fractale-xX: Oh, I didn't see it, thanks.
```julia
using LinearAlgebra
using BenchmarkTools

u = rand(1000)
q = rand(1000)
julia> @btime @. $u = $u + $q
729.458 ns (0 allocations: 0 bytes)
julia> @btime BLAS.axpy!(1000, 1.0, $q, 1, $u, 1)
146.617 ns (0 allocations: 0 bytes)

u = rand(5000)
q = rand(5000)
julia> @btime @. $u = $u + $q
2.755 μs (0 allocations: 0 bytes)
julia> @btime BLAS.axpy!(5000, 1.0, $q, 1, $u, 1)
1.024 μs (0 allocations: 0 bytes)
```
It's seem that `@axpy` is still faster.

dpo: Great!

test/test_alloc.jl
```diff
@@ -21,3 +21,38 @@ expected_cg_lanczos_bytes = storage_cg_lanczos_bytes(n)
 cg_lanczos(A, b)  # warmup
 actual_cg_lanczos_bytes = @allocated cg_lanczos(A, b)
 @test actual_cg_lanczos_bytes ≤ 1.1 * expected_cg_lanczos_bytes
+
+# without preconditioner and with Ap preallocated, CGS needs 6 n-vectors: x, r₀, r, u, p, q
```
dpo: If r₀ is always b, do we need to store it. Couldn't we simply point to b?

test/test_alloc.jl
```diff
@@ -21,3 +21,38 @@ expected_cg_lanczos_bytes = storage_cg_lanczos_bytes(n)
...
+# - 1 mem-vector: L
+# - 1 (mem+2)-vector: H
+# - 1 mem-bitArray: p
+storage_diom(mem, n) = (3 * n) + (2 * n * mem) + (mem) + (mem + 2) + (mem / 64)
+storage_diom_bytes(mem, n) = 8 * storage_diom(mem, n)
+
+# expected_diom_bytes = storage_diom_bytes(mem, n)
+# diom(A, b, M=M, memory=mem)  # warmup
+# actual_diom_bytes = @allocated diom(A, b, M=M, memory=mem)
+# @test actual_diom_bytes ≤ 1.1 * expected_diom_bytes
```
dpo: Why is this commented out?

dpo: Do we exceed 10% more allocated space with the views?

dpo: Does it work if `opEye` returns `convert(Vector{T}, v)` instead of just `v`?

Xx-fractale-xX: Yes It works, but `convert(Vector{Float64}, view[V,:,i])` allocate a new vector everytime. It's worst than preallocated an `opEye`.
```julia
Expression: actual_dqgmres_bytes ≤ 1.05expected_dqgmres_bytes
Evaluated: 22201216 ≤ 5.780544e6
```

dpo: Right, ok. What I don't understand is where the problem comes from. Our only operations with operators are:
```julia
z = M * view(V,:,pos)
w = A * z
```
So if `M` is a non-allocating opEye, then `z` is a view. But as soon as we do `A * z` on the next line, the result is stored in an actual vector (preallocated), and `w` points to that vector. Later, `z` is not used, only `w` is used. So where does the problem come from?

Xx-fractale-xX: I found the problem and it was not easy...

The problem was not from those operations (directly).
``` julia
z = M * view(V,:,pos)
w = A * z
```
All value in `V[:,pos]` were equals to `NaN` at the beginning of step. The problem come from the operations at the end of the last step.

I found it with the help of some `println` :
```julia
z = M * view(V,:,pos) # Forms pₘ
w = A * z # Forms vₘ₊₁
for i = max(1, iter-mem+1) : iter
      ipos = mod(i-1, mem) + 1 # Position corresponding to vᵢ in the circular stack V.
      diag = iter - i + 2
      println("iter:", iter, " i:",i," w[1] ",w[1])
      H[diag] = @kdot(n, w, view(V,:,ipos)) # hᵢ.ₘ = < A * vₘ , vᵢ >
      println("iter:", iter, " i:",i," H[diag] ",H[diag])
      @kaxpy!(n, -H[diag], view(V,:,ipos), w) # w ← w - hᵢ.ₘ * vᵢ
      println("iter:", iter, " i:",i," w[1] ",w[1])
end
```
With a `nonallocating_opEye`:
```julia
iter:1 i:1 w[1] 0.016572815184059706
iter:1 i:1 H[diag] 0.18749999999999997
iter:1 i:1 w[1] 0.015537014235055975
iter:2 i:1 w[1] 0.15070229474239932
iter:2 i:1 H[diag] 0.41926274578121114
iter:2 i:1 w[1] 0.14838617340926818
iter:2 i:2 w[1] 0.14838617340926818
iter:2 i:2 H[diag] 2.1291666666666647
iter:2 i:2 w[1] 0.06948363999393418
iter:3 i:2 w[1] 0.35488547857475394
iter:3 i:2 H[diag] 1.1090223422256065
iter:3 i:2 w[1] 0.34199793096665637 XXX
...
...
...
iter:13 i:13 w[1] -1.7201081594181266e184
iter:13 i:13 H[diag] 3.847236420724795e186
iter:13 i:13 w[1] -2.6103244066888948e185
iter:14 i:13 w[1] -0.1678050399580333
iter:14 i:13 H[diag] 1.1747083383409987e186
iter:14 i:13 w[1] Inf
iter:14 i:14 w[1] Inf
iter:14 i:14 H[diag] NaN
iter:14 i:14 w[1] NaN
iter:15 i:14 w[1] NaN
```

With a `preallocated_opEye` or a classic `opEye`:
``` julia
iter:1 i:1 w[1] 0.016572815184059706
iter:1 i:1 H[diag] 0.18749999999999997
iter:1 i:1 w[1] 0.015537014235055975
iter:2 i:1 w[1] 0.15070229474239932
iter:2 i:1 H[diag] 0.41926274578121114
iter:2 i:1 w[1] 0.14838617340926818
iter:2 i:2 w[1] 0.14838617340926818
iter:2 i:2 H[diag] 2.1291666666666647
iter:2 i:2 w[1] 0.06948363999393418
iter:3 i:2 w[1] 0.35488547857475394
iter:3 i:2 H[diag] 1.1090223422256007
iter:3 i:2 w[1] 0.3137873936827896 XXX
...
...
...
iter:13 i:13 w[1] -0.028634427439454206
iter:13 i:13 H[diag] 5.951330043236927
iter:13 i:13 w[1] -0.03978498986384402
iter:14 i:13 w[1] -0.10053197656509674
iter:14 i:13 H[diag] 2.9061910495701317
iter:14 i:13 w[1] -0.10597708958689267
iter:14 i:14 w[1] -0.10597708958689267
iter:14 i:14 H[diag] 5.812556558561877
iter:14 i:14 w[1] -0.02640472155568238
iter:15 i:14 w[1] -0.09744783677986579
```

I tested it with `mem=2` and because of the `nonallocating_opEye` we have `w = view(V,:,pos)`. But during the Arnoldi process we can have `ipos = pos`.
The operation 
```julia
@kaxpy!(n, -H[diag], view(V,:,ipos), w) = @kaxpy!(n, -H[diag], view(V,:,pos), w) 
                                        = @kaxpy!(n, -H[diag], view(V,:,pos), view(V,:,pos))
```

We have a side effect.
```julia
V[:,pos] = -H[diag] *  V[:,pos] - H[diag] *  V[:,pos]
         = - 2 H[diag] * V[:,pos]
```
instead of 
```julia
V[:,pos] = -H[diag] * V[:,pos]  + V[:,pos]
         = (1-H[diag]) * V[:,pos]
```
The solution is to isolate the last step of the `for loop` and do a `@kscal!` instead of `@kaxpy!` which is even faster!

dpo: > because of the nonallocating_opEye we have w = view(V,:,pos)

That's what I don't understand. I would understand that `z = view(V,:,pos)`, but `w = A * z` must be a standard vector.

Xx-fractale-xX: Oh you're right I reverse `w` and `z`.  I have no idea where the problem comes from now...

dpo: The problem must come from [here](https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/77/files#diff-d9dc38e111bc1f829fd0d34ba3adb6f3R119). We're updating `z`, which is a view, so we're changing `V`. I guess `opEye` should come with preallocation by default, and we could supply an `unsafe_opEye` that's not allocating. But it's a bit annoying that some methods require one and some require the other.

Xx-fractale-xX: We still have the possibility to use the preallocated vector of Ap (w). We don't need it anymore until the next step.
``` julia
    @. w = z # or copyto!(w, z)
    for i = max(1,iter-mem) : iter-1
      ipos = mod(i-1, mem) + 1 # Position corresponding to pᵢ in the circular stack P.
      diag = iter - i + 2
      # z ← z - hᵢ.ₘ * pᵢ
      @kaxpy!(n, -H[diag], view(P,:,ipos), w)
    end
    # pₘ = z / hₘ.ₘ
    @. P[:,pos] = w / H[2]
```
With that it works for an `unsafe_opEye`.

dpo: That's good when there's no preconditioner. Perhaps you can skip it if the preconditioner is not an `unsafe_opEye`? If we use a preconditioner with preallocation, `z` will be a vector.

Xx-fractale-xX: I found a way to detect is we have an `unsafe_opEye`, we can use `===` to see if two objects are the same thing in memory.

test/test_alloc.jl
```diff
@@ -21,3 +23,38 @@ expected_cg_lanczos_bytes = storage_cg_lanczos_bytes(n)
...
+storage_cgs(n) = 5 * n
+storage_cgs_bytes(n) = 8 * storage_cgs(n)
+expected_cgs_bytes = storage_cgs_bytes(n)
+cgs(A, b, M=M)  # warmup
+actual_cgs_bytes = @allocated cgs(A, b, M=M)
+@test actual_cgs_bytes ≤ 1.1 * expected_cgs_bytes
+
+# without preconditioner and with Ap preallocated, DQGMRES needs:
+# - 1 n-vectors: x
+# - 2 (n*mem)-matrix: P, V
```
dpo: matrices

test/test_alloc.jl
```diff
@@ -21,3 +23,38 @@ expected_cg_lanczos_bytes = storage_cg_lanczos_bytes(n)
...
+storage_dqgmres_bytes(mem, n) = 8 * storage_dqgmres(mem, n)
+
+expected_dqgmres_bytes = storage_dqgmres_bytes(mem, n)
+dqgmres(A, b, M=M, memory=mem)  # warmup
+actual_dqgmres_bytes = @allocated dqgmres(A, b, M=M, memory=mem)
+@test actual_dqgmres_bytes ≤ 1.05 * expected_dqgmres_bytes
+
+# without preconditioner and with Ap preallocated, DIOM needs:
+# - 2 n-vectors: x, x_old
+# - 2 (n*mem)-matrix: P, V
```
dpo: matrices

Xx-fractale-xX:
I rebased this branch too.

Xx-fractale-xX:
I updated the pull request, It only contains CGS allocations now.

PR #83, opened 2018-11-07
An implementaton of USYMQR

amontoison:
I will add `USYMLQ` in an other pull request.
I didn't preconditioned the trigonalization for the moment.
I will do it for both `USYMQR` and `USYMLQ` methods later.


amontoison:
I updated `usymqr`, the tridiagonalization in precondtitioned in elliptic norms like you did in `usymlqr`.
The methods is generalized for nonsymmetric matrix and it should also work in multiple precision.

amontoison:
It seems that `usymlqr` has difficulties to solve the linear systems in the allocation tests on 32 bits computers.
But I don't understand why...
Because of that the memory used for ||r|| and ||A' r|| become preponderate and the test failed.

PR #85, opened 2018-11-10
Logging

dpo:
I'm trying to use logging instead of `@printf` and the `verbose` flag in CG. At the moment, there's a substantial performance loss (even when NOT logging!) and also many extra allocations. Here are the benchmarks: https://gist.github.com/a63ab424dd9a848d9c0d267102d282f6.

On examples like `nasa1824`, there's a factor of 16 in terms of allocations!!! On `ex5`, there's a factor of almost 4 in terms of run time.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,10 +1,13 @@
 using BenchmarkTools
 
 using LinearAlgebra
+using Logging
 
 using Krylov
 using LinearOperators
 using MatrixMarket
+
+cg_logger = Logging.NullLogger()  # no output
 
 include("../test/get_div_grad.jl")
 include("../test/test_utils.jl")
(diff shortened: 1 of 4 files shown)


dpo:
cc @abelsiqueira, @Xx-fractale-xX

dpo:
It also makes the allocation tests fail:
```
Test Failed at /Users/travis/build/JuliaSmoothOptimizers/Krylov.jl/test/test_alloc.jl:14
  Expression: actual_cg_bytes ≤ 1.1expected_cg_bytes
   Evaluated: 23716512 ≤ 865075.2000000001
```

Xx-fractale-xX:
I think I see the problem. Each iteration you create a string with @sprintf macro. You need memory and time to do that. Whereas with the old version you display the content of the @sprintf with a @printf only when `verbose = true`.

Besides, you store all @sprintf in a variable `info_line` which need to do dynamic allocation every time we have a `*=`.
Replacing all `verbose && @printf(....)` by `@info @sprintf(...)` could work.

dpo:
You're right! That's it: https://gist.github.com/d0d03774882b157ff8039b1c7a450149
It's incredible how much memory those `@sprintf` were allocating!

dpo:
FYI: https://discourse.julialang.org/t/logging-performance/17389

amontoison:
@dpo Could you run benchmarks with Julia 1.3 when you have time to see if we still have performance regression?
In `cg` parameters maybe you can set a `NullLogger()` by default with :
```julia
verbose :: Bool = false,
logger :: AbstractLogger = verbose ? ConsoleLogger() : NullLogger()
```
You don't need to modify actual benchmarks with that.

amontoison:
superseded by #288

PR #86, opened 2018-11-11
Allocations tests for DQGMRES

Xx-fractale-xX:
I found a proper way to update p_m when we use an unsafe_opeye. I build it from p_{m-mem}.
(lines 118-129)


dpo:
Seems like this needs to be rebased.

Xx-fractale-xX:
Done. I will rebased `DIOM` after. Maybe we will have new conflicts.

dpo:
Thanks!

PR #87, opened 2018-11-11
Allocations tests for DIOM

Xx-fractale-xX:
Same comment that I did for DQGMRES. The modification between lines 122 and 131 allows us to use an unsafe_opEye.

### DIFF
--- a/test/test_alloc.jl
+++ b/test/test_alloc.jl
@@ -1,17 +1,33 @@
 include("test_utils.jl")
 
+A   = preallocated_LinearOperator(get_div_grad(32, 32, 32))
+n   = size(A, 1)
+b   = ones(n)
+M   = nonallocating_opEye(n)
+mem = 10
+
 # without preconditioner and with Ap preallocated, CG needs 3 n-vectors: x, r, p
 storage_cg(n) = 3 * n
 storage_cg_bytes(n) = 8 * storage_cg(n)
 
-A = preallocated_LinearOperator(get_div_grad(32, 32, 32))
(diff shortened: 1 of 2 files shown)


test/test_alloc.jl
```diff
@@ -1,18 +1,34 @@
...
+# - 1 mem-vector: L
+# - 1 (mem+2)-vector: H
+# - 1 mem-bitArray: p
+storage_diom(mem, n) = (2 * n) + (2 * n * mem) + (mem) + (mem + 2) + (mem / 64)
+storage_diom_bytes(mem, n) = 8 * storage_diom(mem, n)
+
+expected_diom_bytes = storage_diom_bytes(mem, n)
+diom(A, b, M=M, memory=mem)  # warmup
+actual_diom_bytes = @allocated diom(A, b, M=M, memory=mem)
+@test actual_diom_bytes ≤ 1.05 * expected_diom_bytes
```
dpo: Why 1.05 here while we use 1.1 everywhere else?

dpo:
Looks good. Only needs to be rebased.

test/test_alloc.jl
Xx-fractale-xX: Because for `DIOM` and `DQGMRES` we need around `22 n-vectors` when `mem = 10`. And if we used one more n-vector somewhere the test passed with a `10%` tolerance. 
For the others methods we use less vectors so `10%` is adequate.

dpo: Ok, thank you for explaining.

PR #91, opened 2018-12-14
An implementation of GMRES

amontoison:


Xx-fractale-xX:
Henceforth we can do a `GMRES(m)` (restarted GMRES).
I think it's possible to not store `V[iter+1]` when `iter` is the last iteration and economize one n-vector.
I will finish `FOM` (with permutations allow as keyword) tomorrow.
I have in mind to compare `GMRES` / `FOM` / `GLQ̀`. (residual, error, memory and time aspects).

Xx-fractale-xX:
I fixed a bug with `unsage_opEye` in `FOM` and `GMRES`.
I also find a way to don't allocate `V[iter+1]` when `iter` is the last iteration.
I don't think that I can improve them anymore.

Xx-fractale-xX:
- GMRES(m) and DQGMRES with memory m compute the same solution but their implementations are a little bit different.
`DQGMRES` computes x at each step (Direct version) whereas `GMRES` compute it only at the end. DQGMRES will use more memory for that (it will store Wn solution of `Rn * Wn = Vn` and update the solution of this system every iteration).
GMRES updates only the QR factorization of H and resolve `Rx = Qt * beta * e1` at the final step. It is also less expensive (cost of a backward substitution) and the storing of H is only O(m*m) coefficients against O(n*m) for Wn in DQGMRES.

- Good idea for the restart, I will add it!

Xx-fractale-xX:
I don't notice but when x₀ ≠ 0, I build the orthogonal basis of {r₀, Ar₀, A²r₀, ..., Aᵐ⁻¹r₀}. It's already the problem A * Δx = r₀ which is resolved.
I only add comments.

dpo:
Yes but solving A ∆x = r from x0=0 should be less sensitive to cancellation.

src/gmres.jl
```diff
@@ -0,0 +1,153 @@
...
+               reorthogonalization :: Bool=false, verbose :: Bool=false) where T <: AbstractFloat
+
+  m, n = size(A)
+  m == n || error("System must be square")
+  length(b) == m || error("Inconsistent problem size")
+  verbose && @printf("GMRES: system of size %d\n", n)
+
+  # Initial solution x₀ and residual r₀.
+  x = x0 # x₀
+  r₀ = A * x0
```
dpo: I wonder if there's an easy way to avoid this product when x0 = 0. Should we compute norm(x0)?

src/gmres.jl
```diff
@@ -0,0 +1,153 @@
...
+  @. r₀ = -r₀ + b
```
dpo: b - r0

src/gmres.jl
```diff
@@ -0,0 +1,153 @@
...
+  r₀ = M * r₀ # r₀ = M⁻¹(b - Ax₀)
+
+  # Compute β
+  rNorm = @knrm2(n, r₀) # β = ‖r₀‖₂
+  rNorm == 0 && return x, SimpleStats(true, false, [rNorm], T[], "x = 0 is a zero-residual solution")
```
dpo: Should the message be "x0 is a zero-residual solution" here?

dpo:
Looks great. In a revision, we should implement a stopping test based on the backward error `rNorm / (norm(A) * norm(x))`. The `norm(A)` can be estimated via the norm of H, which can be accumulated as we compute the H[i,j]. The additional cost would then be to compute `norm(x)`, but that's probably ok. This could surely be done in DQGMRES as well.

src/gmres.jl
amontoison: I'm actually updating it. I was thinking of something like this:
```julia
  # Check is GMRES is restarted
  restarted = (x0 == T[])
  restarted || x0 = x

# Initial residual r₀.
  if restarted
    r₀ = A * x0
    @. r₀ = b -r₀
  else
    r₀ = b
  end
  r₀ = M * r₀ # r₀ = M⁻¹(b - Ax₀)
```

src/gmres.jl
amontoison: Yes and we return `x0` instead of `x`.

src/gmres.jl
amontoison: And at the end of the resolution:
```julia
  # Form xₖ = x₀ + Vₖyₖ
  for i = 1 : iter
    Dᵢ = N * V[i] # Needed with right preconditioning
    @kaxpy!(n, y[i], Dᵢ, x)
  end
  restarted && @kaxpy!(n, 1.0, x0, x)
```

dpo: I think this `gmres` method should not take an `x0` as input. Instead, another `gmres` method should accept an `x0`, shift the problem, call this method, and then add ∆x to x0. That other method could also accept a `restart` argument that gives the number of iterations before a restart occurs. What do you think?

amontoison: Yes, good idea. I didn't know how to include a `restart` argument.

amontoison: `itmax` already give the number of iterations before that a potential restart occurs. We should use `restart` argument as MATLAB do, no?
 It's mean that `restart` is the number of time that we want to restart GMRES.

dpo: Ah yes, ok. We can use arguments similar to MATLAB.

dpo:
Revive this?

amontoison:
I begin to get `GMRES` back to life, but we should discuss what we want to do with this method. Actually I do dynamic allocation to increase the Krylov basis at each iteration. The implementation of restarted GMRES seems more relevant, we will allocate a small matrice with `basis_size` columns at the beginning and always reuse this memory.

dpo:
I agree that full GMRES doesn't make sense. So many people use restarted GMRES that it would be useful to have, even if it's not a great method. It's still useful when making comparisons. Thanks!

PR #92, opened 2018-12-18
An implementation of FOM

amontoison:


Xx-fractale-xX:
```julia
A = get_div_grad(16, 16, 16)
A = A - 5 * I
b = ones(size(A, 1))

(x1, stats1) = fom(A, b, pivoting=false)
n1 = length(stats1.residuals)
plot([1:n1;], stats1.residuals, yaxis=:log, xlabel = "iterations", ylabel = "‖rₖ‖₂", title = "Residuals of FOM without pivoting")

(x2, stats2) = fom(A, b, pivoting=true)
n2 = length(stats2.residuals)
plot([1:n2;], stats2.residuals, yaxis=:log, xlabel = "iterations", ylabel = "‖rₖ‖₂", title = "Residuals of FOM with pivoting")
```
plot1
plot2

Xx-fractale-xX:
I comment allocations test for the moment, I think that I have an wrong solution when I used non allocating preconditioner. I would like to find the problem. Even if tests passed, we should have the same problem with `GMRES`.

dpo:
Very interesting, thank you. Is this property of the residual documented somewhere? Also, it would probably be best to implement restarts here, wouldn't it?

Xx-fractale-xX:
I computed the real residual this time for the same problem. My expression of residual with pivoting is surely incomplete. I still have (small) oscillations.
plot3

plot4

I only see the concept of pivoting in the article that describe `DIOM`. I never see use of it somewhere else.
At first this main interest is to be able to continue the resolution even if `H` is singular.
The `LU` factorization is also more stable (division by the biggest number every step).
Also, the behavior of residual norm seems to be better.

Actually It's possible to do a restart FOM, we can give `x0` as parameter. But at the end, FOM return the last solution that it can compute (x_iter). It's not always the best solution.
A possible improvement is to choose `i` in [1,iter] such as the norm of `b - A * xi` is minimal and compute/return this xi.

dpo:
FOM / DIOM don't guarantee a monotonic residual. Have we determined what quantity FOM minimizes?

Xx-fractale-xX:
No, I still don't know what quantity FOM minimizes... FOM / DIOM don't guarantee a monotonic residual but maybe that pivoting could or at least limit the variations? I didn't find anywhere theoretical expression of the residual when we do a permutation...

dpo:
https://www.researchgate.net/publication/220656796_On_the_residual_norm_in_fom_and_GMRES

dpo:
This would be acceptable with a restart.

PR #94, opened 2019-02-04
Remove random tests

### LINKED ISSUE #5: CRMR tests sometimes pass, sometimes fail
Compare
https://travis-ci.org/JuliaOptimizers/Krylov.jl/builds/128475745
and
https://travis-ci.org/jagot/Krylov.jl/builds/128475739

which both test the same commit.

amontoison:
I didn't remove random data used to build `InverseLBFGSOperator`.
I used a fixed seed instead.
fix #5
#81

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,11 +1,11 @@
-using Test, Krylov, LinearOperators, LinearAlgebra, SparseArrays, Printf
+using Test, Krylov, LinearOperators, LinearAlgebra, SparseArrays, Printf, Random
+
+include("test_utils.jl")
+include("test_aux.jl")
 
 include("test_diom.jl")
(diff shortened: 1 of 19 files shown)


dpo:
This is great, many thanks! In order to avoid repetitions, I wonder if we could have all the test systems in a separate file and give them names. For example,
```julia
A, b = under_inconsistent()
A, b = under_consistent()
...
```
What do you think?

amontoison:
It's a good idea. I will add them in `test_utils.jl`.

dpo:
@amontoison It would be great if you found the time to separate out the tests at some point.

amontoison:
Sorry for the delay.

dpo:
Excellent, thank you!

PR #95, opened 2019-02-04
Allocations tests for CGNE

amontoison:
`A' * z` is allocated whereas it's not the case with `A.tprod(z)`
```julia
julia> @allocated A.tprod(p)
0
julia> @allocated A' * p
112
```
We had the same problem with `view`. A small allocation each step but at the end, around 150Mo is allocated only for that...

### DIFF
--- a/test/test_alloc.jl
+++ b/test/test_alloc.jl
@@ -1,9 +1,15 @@
 include("test_utils.jl")
 
-A   = preallocated_LinearOperator(get_div_grad(32, 32, 32))
-n   = size(A, 1)
+L   = get_div_grad(32, 32, 32)
+n   = size(L, 1)
+m   = div(n, 2)
+A   = preallocated_LinearOperator(L) # Dimension n x n
+Au  = preallocated_LinearOperator(L[1:m,:]) # Dimension m x n
+Ao  = preallocated_LinearOperator(L[:,1:m]) # Dimension n x m
 b   = ones(n)
+c   = ones(m)
 M   = nonallocating_opEye(n)
+N   = nonallocating_opEye(m)
 mem = 10
 
 # without preconditioner and with Ap preallocated, CG needs 3 n-vectors: x, r, p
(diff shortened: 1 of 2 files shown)


amontoison:
`A` is already an operator with preallocation. 
I think that julia compute `A'` first and do the product after.

dpo:
This seems like a bug. `A' * x` and `A.tprod(x)` should be the same thing. The whole point is to be able to write `A' * x`. Could you paste an example here where `A' * x` allocates?

amontoison:
Yes, I can reproduce a case where `A' * x` allocates with this example:
```julia
using LinearOperators, FastClosures

function preallocated_LinearOperator(A)
     (n, m) = size(A)
     Ap = zeros(n)
     prod = @closure p -> mul!(Ap, A, p)
     Atq = zeros(m)
     tprod = @closure q -> mul!(Atq, transpose(A), q)
     T = eltype(A)
     F1 = typeof(prod)
     F2 = typeof(tprod)
     LinearOperator{T,F1,F2,F2}(n, m, false, false, prod, tprod, tprod)
end

R = rand(1000,1000)
v = rand(1000)
A = preallocated_LinearOperator(R)

c = A' * v #warmup
@allocated A' * v
d = A.tprod(v) #warmup 
@allocated A.tprod(v)
```

test/test_alloc.jl
```diff
@@ -1,9 +1,15 @@
 include("test_utils.jl")
 
-A   = preallocated_LinearOperator(get_div_grad(32, 32, 32))
-n   = size(A, 1)
+L   = get_div_grad(32, 32, 32)
+n   = 32768
+m   = 5000
```
dpo: Could we do `n = size(A, 1)` and `m = n / 2` (or something similar?)

amontoison:
It was a good remark. With the fixed (n,m) values we had a false positive test. The allocations was right under `1.1 * expected_cgne_bytes`.

dpo:
Thank you. Hopefully, this kind of thing will go away when we revise the structure of linear operators.

PR #98, opened 2019-02-05
Allocations tests for CGLS

amontoison:
@abelsiqueira 
I will open an other pull request after this one with multi-precision `CG` and `CGLS`.

### DIFF
--- a/src/cgls.jl
+++ b/src/cgls.jl
@@ -50,7 +50,8 @@ function cgls(A :: AbstractLinearOperator, b :: AbstractVector{T};
   r = copy(b)
   bNorm = @knrm2(m, r)   # Marginally faster than norm(b);
   bNorm == 0 && return x, SimpleStats(true, false, [0.0], [0.0], "x = 0 is a zero-residual solution");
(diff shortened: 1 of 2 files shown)


src/cgls.jl
```diff
@@ -71,7 +72,8 @@ function cgls(A :: AbstractLinearOperator, b :: AbstractVector{T};
 
   while ! (solved || tired)
     q = A * p;
-    δ = @kdot(m, q, M * q)   # Faster than α = γ / dot(q, q);
+    Mq = M * q
+    δ = @kdot(m, q, Mq)   # Faster than α = γ / dot(q, q);
```
dpo: What's the difference here?

amontoison: If I used M * q inside a `@kdot` with M an unsafe_opEye() I had sometimes side effect in memory and the product was wrong.
I never had problems when I wrote the expression with two lines.

It was with `julia v0.6`, maybe it's safe now....

dpo:
Thank you!

PR #99, opened 2019-02-06
Multi-precision CG and CGLS

amontoison:
@abelsiqueira 
I only updated `CG` and `CGLS` for the moment. Abel needs them in the near future.

### DIFF
--- a/src/Krylov.jl
+++ b/src/Krylov.jl
@@ -3,34 +3,34 @@ module Krylov
 using LinearOperators, LinearAlgebra, SparseArrays, Printf
 
 "Abstract type for statistics returned by a solver"
-abstract type KrylovStats end;
+abstract type KrylovStats{T} end;
 
 "Type for statistics returned by non-Lanczos solvers"
-mutable struct SimpleStats <: KrylovStats
+mutable struct SimpleStats{T} <: KrylovStats{T}
   solved :: Bool
   inconsistent :: Bool
-  residuals :: Array{Float64,1}
-  Aresiduals :: Array{Float64,1}
+  residuals :: Array{T,1}
+  Aresiduals :: Array{T,1}
   status :: String
 end
 
-mutable struct LanczosStats <: KrylovStats
+mutable struct LanczosStats{T} <: KrylovStats{T}
   solved :: Bool
-  residuals :: Array{Float64}
+  residuals :: Array{T}
   flagged :: Union{Bool, Array{Bool,1}, BitArray{1}}
-  Anorm :: Float64
-  Acond :: Float64
+  Anorm :: T
+  Acond :: T
   status :: String
 end
 
-mutable struct SymmlqStats <: KrylovStats
+mutable struct SymmlqStats{T} <: KrylovStats{T}
   solved :: Bool
-  residuals :: Array{Float64}
-  residualscg :: Array{Float64}
-  errors :: Array{Float64}
-  errorscg :: Array{Float64}
-  Anorm :: Float64
-  Acond :: Float64
+  residuals :: Array{T}
+  residualscg :: Array{T}
+  errors :: Array{T}
+  errorscg :: Array{T}
+  Anorm :: T
+  Acond :: T
   status :: String
 end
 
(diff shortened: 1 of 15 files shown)


src/krylov_aux.jl
```diff
@@ -49,38 +49,38 @@ where q₂, q₁ and q₀ are real. Care is taken to avoid numerical
...
   # Case where q(x) is indeed quadratic.
-  rhs = sqrt(eps(Float64)) * q₁ * q₁
+  rhs = √eps(T) * q₁ * q₁
   if abs(q₀ * q₂) > rhs
-    ρ = q₁ * q₁ - 4.0 * q₂ * q₀
-    ρ < 0.0 && return Float64[]
-    d = -0.5 * (q₁ + copysign(sqrt(ρ), q₁))
+    ρ = q₁ * q₁ - 4 * q₂ * q₀
+    ρ < 0 && return T[]
+    d = -T(0.5) * (q₁ + copysign(sqrt(ρ), q₁))
```
dpo: Better divide by 2?

src/krylov_aux.jl
```diff
@@ -49,38 +49,38 @@ where q₂, q₁ and q₀ are real. Care is taken to avoid numerical
...
+      q₀ == zero(T) || (root = T[])
     else
       root = [-q₀ / q₁]
     end
     return root
   end
 
   # Case where q(x) is indeed quadratic.
-  rhs = sqrt(eps(Float64)) * q₁ * q₁
+  rhs = √eps(T) * q₁ * q₁
```
dpo: Because of this kind of operation, don't we need `T <: AbstractFloat`? Note that the vector `b` will need to be more general (Float U Complex).

test/test_mp.jl
```diff
@@ -0,0 +1,19 @@
...
+  n = 10
+  for fn in (:cg, :cgls)
+    @printf("%s ", string(fn))
+    for T in (Float16, Float32, Float64, BigFloat)
+      M = spdiagm(-1 => ones(T,n-1), 0 => 4*ones(T,n), 1 => ones(T,n-1))
+      A = LinearOperator(M)
+      b = ones(T, n)
+      (x, stats) = @eval $fn($A, $b)
+      atol = √eps(T)
+      rtol = T(1e-6)
```
dpo: Could be the same as `atol`.

dpo:
Looks great! Thank you!

dpo:
Oops.

src/krylov_aux.jl
amontoison: Yes, we need `T <: AbstractFloat` but we can extend `eps` function to complex number.
A complex number in Julia is a composite type with `real` and `imaginary` part of Float16, Float32, Float64, BigFloat...
So we can do something like this:
 ```julia
function eps(C :: Complex{T}) where T <: AbstractFloat
  return eps(T)
end
```
What do you think?

Comment (dpo) [reply src/krylov_aux.jl]: Yes, this seems necessary. But because you use square roots, it wouldn't make sense for `T` (in the function definition) to be, say, an integer type. I think we need `T <: Union{AbstractFloat, Complex}` or something similar.

File: src/cg.jl [hunk 4]
```diff
@@ -19,9 +19,9 @@ A preconditioner M may be provided in the form of a linear operator and is
 assumed to be symmetric and positive definite.
 """
 function cg(A :: AbstractLinearOperator, b :: AbstractVector{T};
-            M :: AbstractLinearOperator=opEye(), atol :: Float64=1.0e-8,
-            rtol :: Float64=1.0e-6, itmax :: Int=0, radius :: Float64=0.0,
-            verbose :: Bool=false) where T <: Number
+            M :: AbstractLinearOperator=opEye(), atol :: T=√eps(T),
+            rtol :: T=T(1.e-6), itmax :: Int=0, radius :: T=zero(T),
```
dpo: I think rtol could also be √eps(T) so it's related to T.

src/Krylov.jl
```diff
@@ -3,34 +3,34 @@ module Krylov
...
 
 "Type for statistics returned by non-Lanczos solvers"
-mutable struct SimpleStats <: KrylovStats
+mutable struct SimpleStats{T} <: KrylovStats{T}
   solved :: Bool
   inconsistent :: Bool
-  residuals :: Array{Float64,1}
-  Aresiduals :: Array{Float64,1}
+  residuals :: Array{T,1}
+  Aresiduals :: Array{T,1}
```
dpo: Let's change this to Vector{T}.

dpo:
Looking good. Almost there!

PR #104, opened 2019-02-11
Allocations tests for CRAIG

amontoison:


dpo:
Great, thank you. I guess I forgot to implement the stopping condition based on cond(A). I can easily add it later.

PR #106, opened 2019-02-12
Allocations tests for CG_LANCZOS_SHIFT_SEQ

amontoison:


test/test_alloc.jl
```diff
@@ -40,6 +43,19 @@ cg_lanczos(A, b)  # warmup
 actual_cg_lanczos_bytes = @allocated cg_lanczos(A, b)
 @test actual_cg_lanczos_bytes ≤ 1.1 * expected_cg_lanczos_bytes
 
+# with Ap preallocated, CG_LANCZOS_SHIFT_SEQ needs:
+# - 2 n-vectors: v, v_prev
+# - 2 (n*nshifts)-matrices: x, p
+# - 5 nshifts-vectors: σ, δhat, ω, γ, rNorms
+# - 2 nshifts-bitArray: indefinite, converged
+storage_cg_lanczos_shift_seq(n, nshifts) = (2 * n) + (2 * n * nshifts) + (5 * nshifts) + (2 * nshifts / 64)
```
dpo: Does 64 assume 64-bit integers here?

amontoison: `storage_cg_lanczos_shift_seq(n, nshifts)` counts the number of element in vectors and matrices. In general each coefficient are in double precision (64 bits). But with BitArrays, only 1 bit is used. So we need to divise by 64 to have the same measure.
Besides, `storage_cg_lanczos_shift_seq_bytes` multiplies everything by 8 because we suppose that each coefficient are written with 8 octets.

dpo: I was wondering if the 64 came from the fact that the underlying system uses 64-bit integers. Would it be 32 on 32-bit systems? I guess not: https://github.com/JuliaLang/julia/blob/master/base/bitarray.jl#L20.

dpo:
Thank you!

PR #109, opened 2019-02-13
Generalized Golub Kahan Bidiagonalization

amontoison:
#52

### DIFF
--- a/test/test_craig.jl
+++ b/test/test_craig.jl
@@ -88,8 +88,9 @@ function test_craig()
   (x, y, stats) = craig(A, b, M=M, N=N)
   show(stats)
   r = b - A * x
-  resid = norm(r) / norm(b)
+  resid = sqrt(dot(r, M * r)) / norm(b)
   @printf("CRAIG: Relative residual: %8.1e\n", resid)
+  @test(norm(x - N * A' * y) ≤ craig_tol * norm(x))
   @test(resid ≤ craig_tol)
   @test(stats.solved)
 end
--- a/test/test_craig.jl
+++ b/test/test_craig.jl
@@ -82,6 +82,16 @@ function test_craig()
   A, b = over_int()
   (x, y, stats) = craig(A, b)
   @test stats.solved
+
+  # Test with preconditioners
+  A, b, M, N = two_preconditioners()
+  (x, y, stats) = craig(A, b, M=M, N=N)
+  show(stats)
+  r = b - A * x
+  resid = norm(r) / norm(b)
+  @printf("CRAIG: Relative residual: %8.1e\n", resid)
+  @test(resid ≤ craig_tol)
+  @test(stats.solved)
 end
 
 test_craig()
--- a/test/test_utils.jl
+++ b/test/test_utils.jl
@@ -119,8 +119,19 @@ end
 function square_preconditioned(n :: Int=10)
   A = ones(n, n) + (n-1) * I
   b = n * [1:n;]
-  M = 1/n * opEye(n)
+  O = zeros(n, n)
+  M = PreallocatedLinearOperator(1/n * (O + I))
   return A, b, M
+end
+
+# Square problems with two preconditioners.
+function two_preconditioners(n :: Int=10, m :: Int=20)
+  A = ones(n, n) + (n-1) * I
+  b = ones(n)
+  O = zeros(n, n)
+  M = PreallocatedLinearOperator(1/√n * (O + I))
+  N = PreallocatedLinearOperator(1/√m * (O + I))
+  return A, b, M, N
 end
 
 # Random Ax = b with b == 0.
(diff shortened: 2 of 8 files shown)


test/test_craig.jl
```diff
@@ -82,6 +82,16 @@ function test_craig()
   A, b = over_int()
   (x, y, stats) = craig(A, b)
   @test stats.solved
+
+  # Test with preconditioners
+  A, b, M, N = two_preconditioners()
+  (x, y, stats) = craig(A, b, M=M, N=N)
+  show(stats)
+  r = b - A * x
+  resid = norm(r) / norm(b)
```
dpo: Residuals should be measured in M-inv norm here.

dpo: Also, you're regularizing, so you could also check that x = Ninv * A'r.

amontoison: Ok thanks, I updated it. 
For the regularization, it's `x = N⁻¹ * Aᵀ * y` that we should have or `x = N⁻¹ * Aᵀ * r`?

dpo: You're solving the SQD system
```
  [ M   A ] [ r ] = [ b ]
  [ A' -N ] [ x ]   [ 0 ]
```
so *x = N<sup>⁻¹</sup> A<sup>ᵀ</sup> r*.

amontoison: Thank you for the explanation. With `r = b - Ax` and  `M` / `N` permuted we have the same relations.

dpo: It would be best to improve the docstring to explain what M and N do.

amontoison: Ok, but I want to be sure to understand how generalized Golub-Kahan bidiagonalization works with `CRAIG` and `CRAIGMR` before.
I will update this pull request when it's done.

test/test_craig.jl
```diff
@@ -82,6 +82,17 @@ function test_craig()
   A, b = over_int()
   (x, y, stats) = craig(A, b)
   @test stats.solved
+
+  # Test with preconditioners
+  A, b, M, N = two_preconditioners()
+  (x, y, stats) = craig(A, b, M=M, N=N)
+  show(stats)
+  r = b - A * x
```
dpo: ps : il faut que Ax=b soit consistent pour CRAIG. Est-ce que ce test le garantit ?

test/test_utils.jl
```diff
@@ -119,10 +119,21 @@ end
...
   return A, b, M
 end
 
+# Square problems with two preconditioners.
+function two_preconditioners(n :: Int=10, m :: Int=20)
+  A = ones(n, n) + (n-1) * I
+  b = ones(n)
+  O = zeros(n, n)
+  M = PreallocatedLinearOperator(1/√n * (O + I))
+  N = PreallocatedLinearOperator(1/√m * (O + I))
```
dpo: Pourquoi `O + I`? Pour avoir quelque chose en Float64 ?

test/test_craig.jl
amontoison: Oui, c'est un cas particulier où A est carré, symétrique et définie positive. On peut ainsi utiliser ce test dans n'importe quelle méthode qui utilise deux préconditionneurs. Tout b non nul appartient donc `Im(A)` et en particulier `ones(n)`. Le système est donc toujours consistent.

test/test_utils.jl
amontoison: C'est juste pour avoir une matrice identité dense et non un `UniformScaling`.
Depuis julia 1.0, on ne peut plus utiliser `eye(n)`. À la place il faut faire `Matrix{Float64}(I, n, n)`
ou `O + I` avec `O = zeros(n, n)`.

dpo: Ok. This is very different from `1/√n * opEye()`, right? The latter operator could be confusing, and might break things as a preconditioner, right?

amontoison: `M₁ = 1/√n * opEye()` and `M₂ = PreallocatedLinearOperator(1/√n * (O + I))` represents the same matrix. The only difference is that `M₂ * p` is preallocated. It's more difficult to implement preconditioners with that constraint. But if it works with M₂, it works with any non-preallocated preconditioner including M₁.

With or without preconditioner(s) the solvers use the same amount of memory.

dpo: Actually, `1/√n * opEye()` is an invalid command.

amontoison: Sorry, I read `1/√n * OpEye(n)`, the previous preconditionner. Yes `1/√n * opEye()` is very different. Even if it works, it will break things beacause it will update `q` if with a `M * q` and we don't want that.

amontoison: It's because of that problem that we talked about a preallocated `OpEye`. But we could use it only as Jacobi preconditionner when all diagonal coefficient are the same, very special case.

amontoison:
I updated the documentation, but needs to be verified.

test/test_craig.jl
dpo: I tested your implementation of generalized CRAIG and I think it's correct. It's a bit ambiguous that M and N are swapped compared to LSQR and LSMR. We should make the docstrings and implementations uniform. Other than that, I think it's good to go. Thank you!

dpo:
Thank you.

PR #110, opened 2019-02-18
Allocations tests for LSMR

amontoison:


src/lsmr.jl
```diff
@@ -32,26 +32,26 @@ export lsmr
...
 
-  (A'A + λ² I) x = A'b
+  (AᵀA + λ² I) x = Aᵀb
 
 (and therefore to CRLS) but is more stable.
 
-LSMR produces monotonic residuals ‖r‖₂ and optimality residuals ‖A'r‖₂.
-rt is formally equivalent to CRLS, though can be slightly more accurate.
+LSMR produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᵀr‖₂.
+It is formally equivalent to CRLS, though can be slightly more accurate.
```
dpo: I think it would be fair to replace "slightly" with "substantially" here.

src/lsmr.jl
```diff
@@ -194,8 +195,9 @@ function lsmr(A :: AbstractLinearOperator, b :: AbstractVector{T};
     err_vec[mod(iter, window) + 1] = ζ
     iter >= window && (err_lbnd = @knrm2(window, err_vec))
 
-    # Update h, h_har, x.
-    hbar = h - (θbar * ρ / (ρold * ρbarold)) * hbar
+    # Update h, hbar and x.
+    δ = θbar * ρ / (ρold * ρbarold) # δₖ = θbarₖ * ρₖ / (ρₖ₋₁ * ρbarₖ₋₁)
+    @kaxpby!(n, 1.0, h, -δ, hbar)   # ĥₖ = hₖ - δₖ * ĥₖ₋₁
```
dpo: In the comment, ĥₖ is a \hat{h}. Could you make it a \bar{h}?

dpo: Maybe in the code too?!

amontoison: The only \bar{h} available is this one `ħ`. But It's difficult to distinguish with `h`.

dpo: OK.

dpo:
Thank you.

PR #111, opened 2019-02-19
update benchmarks

dpo:
Update benchmark scripts to use SolverBenchmarks.jl and SuiteSparseMatrixCollection.jl.


abelsiqueira:
Thanks

PR #112, opened 2019-02-19
Allocations tests for LSQR

amontoison:
I replace `slightly` by `substantially` like you recommanded.
Modifications are similar to `LSMR` ones. Except 2 other allocations mistakes.
- A vector `d` was allocated at each iteration and was not needed.
- A vector `var` was used but it didn't seem to have any special utility.

### DIFF
--- a/src/crls.jl
+++ b/src/crls.jl
@@ -33,7 +33,7 @@ applying MINRES to the normal equations
 This implementation recurs the residual r := b - Ax.
 
 CRLS produces monotonic residuals ‖r‖₂ and optimality residuals ‖A'r‖₂.
-It is formally equivalent to LSMR, though can be slightly less accurate,
+It is formally equivalent to LSMR, though can be substantially less accurate,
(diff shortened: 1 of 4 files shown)


amontoison:
@abelsiqueira, this method use now one more `m-vector` compared to `CGLS` with `m` the number of variable.
Maybe It could interested you for ill-conditioned problems...

src/cgls.jl
```diff
@@ -34,7 +34,7 @@ parameter. This method is equivalent to applying CG to the normal equations
 but is more stable.
 
 CGLS produces monotonic residuals ‖r‖₂ but not optimality residuals ‖A'r‖₂.
-It is formally equivalent to LSQR, though can be slightly less accurate,
+It is formally equivalent to LSQR, though can be substantially less accurate,
```
dpo: The change only applied to LSMR vs CRLS, not the others.

dpo:
Thank you!

PR #114, opened 2019-02-25
Allocations tests for LSLQ

amontoison:
I moved the transition to LSQR point at the end.
We could compute it only one time instead of updating it at each iteration.


dpo:
Great, thank you!

PR #115, opened 2019-02-25
Allocations tests for CRLS

amontoison:
Even with the `(A * u, Aᵀ * v)` vectors preallocated and used as storage for `(s, q)`, I used more memory compared to `LSMR` with least square problems.
Is it normal or `CRLS` can be more optimized?

### DIFF
--- a/src/crls.jl
+++ b/src/crls.jl
@@ -1,11 +1,11 @@
 # An implementation of CRLS for the solution of the
 # over-determined linear least-squares problem
 #
-#  minimize ‖Ax - b‖
+#  minimize ‖Ax - b‖₂
 #
 # equivalently, of the linear system
 #
-#  A'Ax = A'b.
+#  AᵀAx = Aᵀb.
 #
 # This implementation follows the formulation given in
 #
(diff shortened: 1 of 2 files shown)


dpo:
I guess it depends on m and n: 3m + 2n > 4m + n when n > m, i.e., for under-determined problems. However, your Ao is over-determined, so something doesn't seem right.

amontoison:
I understand where comes the problem. In a classic implementation of  `CRLS` the residual `rₖ = b - Axₖ` is not computed. If it's implemented without `rₖ`, one less `n-vector` is used.
`CRLS` and `LSMR` will use the same number of `n-vectors` but one less `m-vector` for `CRLS`.
We have exactly the same thing with `CGLS` and `LSQR`.
crls

amontoison:
But we can easily add the expression of `rₖ` like you did with the cost of an extra `n-vector`.
crls2

dpo:
Good catch, thank you! Let's leave the residual in there for now since the docstring mentions it explicitly. Later, we could add an option to avoid recurring it.

PR #116, opened 2019-02-27
Allocations tests for MINRES

amontoison:
- Commit `Test preconditioned MINRES` tests that without modification, preconditioned MINRES works.
- Commit `Allocations tests for MINRES` modified `copy(...)` into `@.` and add an allocation test for MINRES in `test_alloc`.
- Commit `Remove allocation for w in MINRES` computes wₖ from  wₖ₋₁ and wₖ₋₂ (w2 and w1) without allocation.
- Commit `Remove allocation for v in MINRES` uses the relation `M * r2 / β = v` to avoid the allocation of v. 
wₖ needs vₖ, so we need to compute it during the Lanczos process and before r2 contains `βₖ₊₁ M⁻¹ * vₖ₊₁` .

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -60,8 +60,8 @@ function minres(A :: AbstractLinearOperator, b :: AbstractVector{T};
   # Initialize Lanczos process.
   # β₁ M v₁ = b.
   r1 = copy(b)
-  y = M * r1
-  β₁ = @kdot(m, r1, y)
+  v = M * r1
+  β₁ = @kdot(m, r1, v)
   β₁ < 0.0 && error("Preconditioner is not positive definite")
   β₁ == 0.0 && return (x, SimpleStats(true, true, [0.0], [0.0], "x = 0 is a zero-residual solution"))
   β₁ = sqrt(β₁)
@@ -79,7 +79,6 @@ function minres(A :: AbstractLinearOperator, b :: AbstractVector{T};
   γmin = Inf
   cs = -1.0
   sn = 0.0
-  v = zeros(T, n)
   w1 = zeros(T, n)
   w2 = zeros(T, n)
   r2 = copy(r1)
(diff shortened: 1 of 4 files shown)


src/minres.jl
```diff
@@ -153,13 +152,26 @@ function minres(A :: AbstractLinearOperator, b :: AbstractVector{T};
...
+    # Compute w.
+    if iter == 1
+      w = w2
+    else
+      iter ≥ 3 && @kscal!(n, -oldϵ, w1)
+      w = w1
+      @kaxpy!(n, -δ, w2, w)
+    end
+    @kaxpy!(n, 1.0, v, w)
+    @kscal!(n, 1 / γ, w)
```
dpo: Smart!

src/minres.jl
```diff
@@ -153,13 +152,26 @@ function minres(A :: AbstractLinearOperator, b :: AbstractVector{T};
...
     # Update x.
-    @. w1 = w2
-    @. w2 = w
-    @. w = (v - oldϵ * w1 - δ * w2) / γ
     @kaxpy!(n, ϕ, w, x)  # x = x + ϕ * w
     xENorm² = xENorm² + ϕ * ϕ
 
+    # Update directions for x.
+    if iter ≥ 2
+      w1, w2 = w2, w
```
dpo: This creates a tuple, doesn't it? Could we do
```
w1 = w2
w2 = w
```

dpo:
This looks great. Thank you!

src/minres.jl
amontoison: `w1, w2 = w2, w` exchanges the pointers to the vectors in memory.
If we do 
```julia
w1 = w2
w2 = w
```
The three vectors represent the same elements in memory.
But we can do that :
```julia
waux = w1
w1 = w2
w2 = waux
```
But I found that less elegant.

amontoison:
```julia
n = 10^8
w2 = rand(n)
w = rand(n)
@time w1, w2 = w2, w
``` 
0.000032 seconds (7 allocations: 256 bytes)
```
julia> w1 = ones(5) 
5-element Array{Float64,1}:
 1.0
 1.0
 1.0
 1.0
 1.0

julia> w2 = 2 * ones(5)
5-element Array{Float64,1}:
 2.0
 2.0
 2.0
 2.0
 2.0

julia> w1, w2 = w2, w1
([2.0, 2.0, 2.0, 2.0, 2.0], [1.0, 1.0, 1.0, 1.0, 1.0])
```
You're right, this creates a tuple!

dpo: Yes, that's the only way it's able to swap two variables.

amontoison: I did a mistake in my previous message, 
```julia
waux = w1
w1 = w2
w2 = waux
```
has the same consequence that
```julia
w1 = w2
w2 = w
```
I don't know how to this trick otherwise.

dpo: > The three vectors represent the same element in memory

Sure, sorry, I typed that too fast. How about
```julia
macro swap(x,y)
   quote
      local tmp = $(esc(x))
      $(esc(x)) = $(esc(y))
      $(esc(y)) = tmp
    end
end

using BenchmarkTools

n = 10^8
w2 = rand(n)
w = rand(n)
@btime @swap $w $w2
  1.462 ns (0 allocations: 0 bytes)

@btime $w, $w2 = $w2, $w
  9.095 ns (1 allocation: 32 bytes)
```

dpo: (really splitting hair here, but hey)

dpo: Let's make it `@kwsap` !

amontoison: How did you find that??? I don't even know what `esc` function do.
Nice  `Metaprogramming` discovery! And it doesn't allocated anything!

amontoison: The swap is also 6 times faster. I suppose that this time was used for create a tuple.

dpo: Fun fact: `@kswap` is someone's handle on Github! ;-)

dpo:
Thank you!

PR #118, opened 2019-03-02
setup docs

dpo:


docs/src/index.md
```diff
@@ -0,0 +1,55 @@
+# [Krylov.jl documentation](@id Home)
+
+This package implements iterative methods for the solution of linear systems of equations
+<p align="center">
+  <b><i>Ax = b</i></b>,
+</p>
+and linear least-squares problems
```
abelsiqueira: You can use LaTeX ([NLPModels.jl](https://raw.githubusercontent.com/JuliaSmoothOptimizers/NLPModels.jl/master/docs/src/index.md))

.travis.yml
```diff
@@ -18,6 +18,16 @@ branches:
   only:
     - master
```
abelsiqueira: You need this to build on tags (and therefore have docs on tags):
```
- /^v\d+\.\d+(\.\d+)?(-\S*)?$/ # tags
```

dpo:
Any idea why the doc build fails? The error message is absolutely useless.

docs/make.jl
```diff
@@ -0,0 +1,21 @@
+using Documenter, Krylov
+
+makedocs(
+  modules = [Krylov],
+  doctest = true,
+  linkcheck = true,
+  strict = true,
```
abelsiqueira: This turns the warnings in the build into errors, and that's why it fails.
If you're not gonna document everything right away, better to comment out this line.

dpo: Thanks. I just found that out. It's ridiculous that it emits warnings and not errors.

PR #121, opened 2019-03-04
CRAIG backward error

### LINKED ISSUE #119: CRAIG should stop based on the backward error
The stopping tests should be as in LSQR, etc., and should include the backward error `rNorm / sqrt(β₁^2 + Anorm² * xNorm²)`.

dpo:
This PR
* closes #119 by implementing a stopping test based on the backward error
* implements a stopping test for ill-conditioned operators
* implements an option to transfer to the LSQR point at the end of the iterations
* removes old useless semicolons from the code.

It's easiest to examine each commit separately.

### DIFF
--- a/src/craig.jl
+++ b/src/craig.jl
@@ -50,7 +50,8 @@ multipliers of the least-norm problem
 In this implementation, both the x and y-parts of the solution are returned.
 """
 function craig(A :: AbstractLinearOperator, b :: AbstractVector{T};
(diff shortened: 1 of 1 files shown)


amontoison:
The error with `SmallNumber` is linked with : 
https://github.com/JuliaLang/julia/issues/29662
Maybe we shoud drop `julia 0.7`? 
With `julia 1.1`, `julia 0.7` is not maintained anymore.

dpo:
Yes, let's drop 0.7!

amontoison:
Done in #122

PR #123, opened 2019-03-07
 Allocations tests for SYMMLQ

### LINKED ISSUE #97: Allocations tests
We should check the allocations of each solver (old and future ones). It's a good way to find mistakes and verify that implementations perform well.
# Allocations tests roadmap
## Methods already in Krylov.jl
- [x] CG_LANCZOS
- [x] CG_LANCZOS_SHIFT_SEQ
- [x] CG
- [x] CR
- [x] CGLS
- [x] CRLS
- [x] CGNE
- [x] CRAIG
- [x] CRMR
- [x] CRAIGMR
- [x] LSLQ
- [x] LSMR
- [x] LSQR
- [x] MINRES
- [x] SYMMLQ
- [x] DQGMRES
- [x] DIOM
- [x] CGS
## Future methods
- [x] USYMLQ
- [x] USYMQR
- [x] TFQMR
- [x] GMRES
- [x] FOM

amontoison:
fix #97

### DIFF
--- a/src/symmlq.jl
+++ b/src/symmlq.jl
@@ -117,7 +117,7 @@ function symmlq(A :: AbstractLinearOperator, b :: AbstractVector{T};
                      0, rNorm, β, cold, sold, ANorm, Acond)
 
   iter = 0
-  itmax == 0 && (itmax = n)
+  itmax == 0 && (itmax = 2 * n)
 
   tol = atol + rtol * β₁
   status = "unknown"
--- a/src/symmlq.jl
+++ b/src/symmlq.jl
@@ -39,29 +39,33 @@ function symmlq(A :: AbstractLinearOperator, b :: AbstractVector{T};
   MisI = isa(M, opEye)
 
   ϵM = eps(T)
-  x = zeros(T, n)
-  ctol = conlim > 0.0 ? 1 ./ conlim : 0.0;
+  x_lq = zeros(T, n)
+  x_cg = zeros(T, n)
+  ctol = conlim > 0.0 ? 1 / conlim : 0.0;
 
   # Initialize Lanczos process.
   # β₁ M v₁ = b.
-  vold = copy(b)
-  p = M * vold
-  β₁ = @kdot(m, vold, p)
-  β₁ == 0.0 && return (x, x, SimpleStats(true, true, [0.0], [0.0], "x = 0 is a zero-residual solution"))
+  Mvold = copy(b)
+  vold = M * Mvold
+  β₁ = @kdot(m, vold, Mvold)
+  β₁ == 0.0 && return (x_lq, x_cg, SimpleStats(true, true, [0.0], [0.0], "x = 0 is a zero-residual solution"))
   β₁ = sqrt(β₁)
   β = β₁
-  @kscal!(m, 1 ./ β, vold)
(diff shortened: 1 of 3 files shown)


src/symmlq.jl
```diff
@@ -70,14 +74,11 @@ function symmlq(A :: AbstractLinearOperator, b :: AbstractVector{T};
   cold = 1.0
   sold = 0.0
 
-  wbar = vold
   ζold = 0.0
   ζbar = β₁/γbar
 
   ANorm² = α * α + β * β
 
-  xcg = ζbar*vold
```
dpo: I can't find where this line went.

src/symmlq.jl
```diff
@@ -260,9 +262,12 @@ function symmlq(A :: AbstractLinearOperator, b :: AbstractVector{T};
     solved = solved_mach | solved_lim | zero_resid_mach | zero_resid_lim | fwd_err
   end
 
+  # Compute CG point
+  @. x_cg = x_lq + ζbar * wbar
```
dpo: Do you only update x_cg once at the end?

amontoison: Yes

src/symmlq.jl
amontoison: I removed it. We can update `x_cg` only one time  at the end even if `iter=1`.

dpo:
Ok, thank you. Technically, we could do without x_cg. It's really there for information. We could update x_lq one last time and save a vector. As an intermediate measure, we could overwrite wbar with x_cg.

amontoison:
Maybe it's possible to use one less vector in the Lanczos process (`MINRES`, `CG_LANCZOS`, `SYMMLQ` methods)

Currently our implementation is :
one preallocated vector Av for Avₖ, one vector for vₖ and one vector for vₖ₋₁ for a total of three vectors.
Then we update Av with some @kaxpy! to have this relation : 
`βₖ₊vₖ₊₁ = Avₖ - δₖvₖ - βₖvₖ₋₁` with `δₖ = vₖᵀ Avₖ`

But I think we can build `vₖ₊₁` from `vₖ₋₁`, It could be possible to do a @kscal! to compute βₖvₖ₋₁
and then compute βₖvₖ₋₁ + Avₖ directly. Like this for example `vₖ₋₁[i] = βₖvₖ₋₁[i] * sum(A[i,j] * vₖ[j] for j=1:n)` (basic form without BLAS operation).

It's still possible to compute δₖ because vₖᵀ ( βₖvₖ₋₁ + Avₖ) =  vₖᵀ Avₖ and vₖ₋₁ ⟂ vₖ.
We finish to update vₖ₊₁ with `Avₖ - βₖvₖ₋₁ - δₖvₖ` and after we do a `@kswap` to swich pointers.

We can use only two vectors with that, what do you think?

amontoison:
I like the idea to overwrite `wbar` with `x_cg`.
```julia
@kaxpy!(n, 1.0, x_lq, ζbar, wbar)
x_cg = wbar
```
You already merge but we can do it in the future.

dpo:
Feel free to open a new PR. It will be simple, and we won't have to give up x_cg. The storage is already there!

amontoison:
Yes, you're right. I can do now, it's only two lines to write.

PR #125, opened 2019-03-09
Minor improvements to CRAIG and LSQR

dpo:


dpo:
```
Simple stats
  solved: true
  inconsistent: false
  residuals:  [  2.0e+01  1.2e+01  1.7e+17  5.6e+02 ]
  Aresiduals: [ ]
  status: solution good enough for the tolerances given
Test Failed at /Users/travis/build/JuliaSmoothOptimizers/Krylov.jl/test/test_craig.jl:30
  Expression: stats.inconsistent
```
This is a perverse effect of a test based on the backward error. On an inconsistent system, norm(x) becomes so large that the convergence test is triggered. We need better infeasibility detection.

PR #127, opened 2019-03-24
Don't allocate w in SYMMLQ and LSLQ

amontoison:
A little trick that I found for `USYMLQ`, `BiLQ` and `LNLQ`.
@dpo It could interested you, it's possible to avoid the allocation of `p` and only use `p̅` in your `USYMLQR`.


dpo:
Thank you!

PR #138, opened 2019-08-22
SYMMLQ improvements

amontoison:
I modified SYMMLQ to be able to compute `x_lq` when `x_cg` doesn't exist.
`ζ = ζbar * c` is equal to `Inf` if `ζbar = Inf`  / `γbar = 0`.
The `η` variable has been introduce to solve this problem.

### DIFF
--- a/src/symmlq.jl
+++ b/src/symmlq.jl
@@ -25,7 +25,8 @@ A preconditioner M may be provided in the form of a linear operator and is
 assumed to be symmetric and positive definite.
 """
 function symmlq(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
-                M :: AbstractLinearOperator=opEye(), λ :: T=zero(T),
+                M :: AbstractLinearOperator=opEye(),
+                λ :: T=zero(T), transfer_to_cg :: Bool=true,
                 λest :: T=zero(T), atol :: T=√eps(T), rtol :: T=√eps(T),
                 etol :: T=√eps(T), window :: Int=0, itmax :: Int=0,
                 conlim :: T=1/√eps(T), verbose :: Bool=false) where T <: AbstractFloat
@@ -39,7 +40,7 @@ function symmlq(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
   MisI = isa(M, opEye)
 
   ϵM = eps(T)
-  x_lq = zeros(T, n)
+  x = zeros(T, n)
   ctol = conlim > 0 ? 1 / conlim : zero(T)
 
   # Initialize Lanczos process.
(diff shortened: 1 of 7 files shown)


src/symmlq.jl
```diff
@@ -82,7 +85,7 @@ function symmlq(A :: AbstractLinearOperator, b :: AbstractVector{T};
   ANorm = 0.0
   Acond = 0.0
   rNorm = β₁
-  rcgNorm = β₁
+  rcgNorm = β * abs(ζbar)
```
dpo: Can't we still initialize this to beta1 with a comment that it is equal to beta * abs(zetabar)?

src/symmlq.jl
```diff
@@ -172,10 +173,11 @@ function symmlq(A :: AbstractLinearOperator, b :: AbstractVector{T};
     γbar = δbar * s - α * c
     ϵ = β * s
     δbar = -β * c
-    ζbar = -(ϵold * ζold + δ * ζ)/γbar
+    η = -ϵold * ζold - δ * ζ
+    ζbar = η / γbar
 
     rNorm = sqrt(γ * γ * ζ * ζ + ϵold * ϵold * ζold * ζold)
-    rcgNorm = abs(rcgNorm*s*cold/c)
+    rcgNorm = β * abs(s * ζ - c * ζbar)
```
dpo: Could you provide more information on these two changes?

dpo: Should we check if gammabar = 0 ?

amontoison:
I rebase this pull request.
Contrary to `BiLQ` or `USYMLQ`, I don't check that `γbar ≠ 0` when I compute `ζbar`. We need it to compute `CG` residual norms and error norms bounds that are returned in a `SymmlqStats` at the end.

dpo:
I'm not thrilled about having Infs in the code when gammabar = 0. I think we should test like in the other solvers?!

amontoison:
My problem is what should we do with `rcgNorm` and `errorscg` if `gammabar = 0`?
A solution could be to not update them, but the drawback is that we loose the link between `iter` and `rcgNorm[iter]`. We don't anymore at which iteration this error norm or residual norm correspond.

dpo:
How do you do it in BiLQ? We could insert a `missing` value for the CG residual and error, which would avoid using Infs and NaNs.

amontoison:
In BiLQ and USYMLQ, residual norms are computed only if CG point exists (δbarₖ ≠ 0). With those methods, a SimpleStats is returned at the end. So it's not a problem if rcgNorms are not computed because we don't need to store them.
With SYMMLQ, more elaborated statistics are returned. They include residuals and errors norms of CG points. Using missing value if a CG point doesn't exist is a good idea.

dpo:
> In BiLQ and USYMLQ, residual norms are computed only if CG point exists (δbarₖ ≠ 0). With those methods, a SimpleStats is returned at the end. So it's not a problem if rcgNorms are not computed because we don't need to store them.

BiLQ and USYMLQ could (should) also return CG residuals, even if they don't form the CG iterate until the end.

amontoison:
I think we could generalize `SymmlqStats` to `LQStats` in an other pull request for all LQ methods (LSLQ, LNLQ, SYMMLQ, BiLQ, USYMLQ).

dpo:
sounds good.

dpo:
Thank you.

PR #139, opened 2019-09-16
Add @kref! macro

amontoison:


amontoison:
I open small pull requests to understand why Travis doesn't work anymore with my account.

amontoison:
The problem came for Travis, my old username Xx-fractale-xX was flagged for abuse...

dpo:
:-) Great that you sorted it out!

src/krylov_aux.jl
```diff
@@ -151,6 +151,16 @@ function krylov_axpby!(n :: Int, s :: T, x :: AbstractVector{T}, dx :: Int, t ::
   return y
 end
 
+function krylov_ref!(n :: Int, x :: AbstractVector{T}, y :: AbstractVector{T}, c :: T , s :: T) where T <: Number
```
dpo: Should we add an increment `dx` here as in the other functions? The macros are just a way to use `dx=1` by default.

amontoison: I didn't implement an increment `dx` because this function doesn't have his `BLAS` version like the other ones. We could easily add `dx` and `dy` it in the future if Julia add a `BLAS` version of it.
For the moment, `BLAS` only implement a `krot!` (same spirit but with a Givens rotation) and it isn't interfaced in Julia...

dpo: Right, but it would be trivial to add an increment even though there's no BLAS function.

dpo:
Thank you!

PR #142, opened 2019-09-19
An implementation of MINRES-QLP

amontoison:
For the moment, this `MINRES-QLP` isn't optimized for singular inconsistent systems.
We could easily add conditions for returning the minimum-norm solution.

For the same amount of memory and few more operations,  this `MINRES-QLP` is a robust alternative of `MINRES`.


src/minres_qlp.jl
```diff
@@ -226,13 +229,20 @@ function minres_qlp(A :: AbstractLinearOperator, b :: AbstractVector{T};
...
     rNorm = abs(ζbarₖ₊₁)
     push!(rNorms, rNorm)
 
+    # Update ‖Arₖ₋₁‖ estimate
+    # ‖ Arₖ₋₁ ‖ = |ζbarₖ| * √((λₖ)² + (γbarₖ)²)
+    ArNorm = abs(ζbarₖ) * √(λₖ^2 + (cₖ * βₖ₊₁)^2)
+    push!(ArNorms, ArNorm)
+
     # Update stopping criterion.
+    iter == 1 && (ϟ = atol + rtol * ArNorm)
```
amontoison: @dpo Is it a relevant stopping condition ? I reused what we do with the residual norm, but I'm not sure that we should do the same thing with Aresidual norm...

dpo: What is that weird symbol? A lightning strike? Does it represent the current ArNorm?

amontoison: `\koppa`, a `k` in old grec. I found it when I wanted to write `\kappa`. It's a nice (grec) symbol. It represents `atol + rtol * Ar₀`. It's similar to `ε = atol + rtol * r₀`.

dpo: It's a fine stopping criterion, but I'm not crazy about that symbol. I like \kappa because it reminds us of "KKT".

amontoison: Ok, I just need to verify that `‖ Arₖ₋₁ ‖ = |ζbarₖ| * √((λₖ)² + (γbarₖ)²)` or `‖ Arₖ₋₁ ‖ = |ζbarₖ| * √((λbarₖ)² + (γbarₖ)²)` and change my lightning stike symbol.

amontoison: There was a mistake in `‖ Arₖ₋₁ ‖`, I corrected it. I finished what I wanted to do with `MINRES-QLP`.

dpo:
🤘 🚀

PR #146, opened 2019-10-09
An implementation of BiLQ

amontoison:


src/bilq.jl
```diff
@@ -0,0 +1,216 @@
...
+  iter = 0
+  itmax == 0 && (itmax = 2*n)
+
+  rNorms = [rNorm;]
+  ε = atol + rtol * bNorm
+  verbose && @printf("%5d  %7.1e\n", iter, rNorm)
+
+  # Set up workspace.
+  bᵗc = @kdot(n, b, c)      # ⟨b,c⟩
+  βₖ = √(abs(bᵗc))          # β₁γ₁ = bᵀc
```
dpo: Shouldn't there be a check for b'c = 0 here?

src/bilq.jl
```diff
@@ -0,0 +1,216 @@
...
+  cₖ₋₁ = cₖ = -one(T)       # Givens cosines used for the LQ factorization of Tₖ
+  sₖ₋₁ = sₖ = zero(T)       # Givens sines used for the LQ factorization of Tₖ
+  d̅ = zeros(T, n)           # Last column of D̅ₖ = Vₖ(Qₖ)ᵀ
+  ζₖ₋₁ = ζbarₖ = zero(T)    # ζₖ₋₁ and ζbarₖ are the last components of z̅ₖ = (L̅ₖ)⁻¹β₁e₁
+  ζₖ₋₂ = ηₖ = zero(T)       # ζₖ₋₂ and ηₖ are used to update ζₖ₋₁ and ζbarₖ
+  δbarₖ₋₁ = δbarₖ = zero(T) # Coefficients of Lₖ₋₁ and L̅ₖ modified during two iterations
+  norm_vₖ = bNorm / βₖ     # ‖vₖ‖ used for residual norm estimates
+
+  # Stopping criterion.
+  solved = rNorm ≤ ε
```
dpo: You could use solved_mach and solved_lim as in LSQR and others.

src/bilq.jl
```diff
@@ -0,0 +1,216 @@
...
+  uₖ₋₁ = zeros(T, n)        # u₀ = 0
+  vₖ = b / βₖ               # v₁ = b / β₁
+  uₖ = c / γₖ               # u₁ = c / γ₁
+  cₖ₋₁ = cₖ = -one(T)       # Givens cosines used for the LQ factorization of Tₖ
+  sₖ₋₁ = sₖ = zero(T)       # Givens sines used for the LQ factorization of Tₖ
+  d̅ = zeros(T, n)           # Last column of D̅ₖ = Vₖ(Qₖ)ᵀ
+  ζₖ₋₁ = ζbarₖ = zero(T)    # ζₖ₋₁ and ζbarₖ are the last components of z̅ₖ = (L̅ₖ)⁻¹β₁e₁
+  ζₖ₋₂ = ηₖ = zero(T)       # ζₖ₋₂ and ηₖ are used to update ζₖ₋₁ and ζbarₖ
+  δbarₖ₋₁ = δbarₖ = zero(T) # Coefficients of Lₖ₋₁ and L̅ₖ modified during two iterations
+  norm_vₖ = bNorm / βₖ     # ‖vₖ‖ used for residual norm estimates
```
dpo: Align comment with the previous ones?

src/bilq.jl
```diff
@@ -0,0 +1,216 @@
...
+  vₖ₋₁ = zeros(T, n)        # v₀ = 0
+  uₖ₋₁ = zeros(T, n)        # u₀ = 0
+  vₖ = b / βₖ               # v₁ = b / β₁
+  uₖ = c / γₖ               # u₁ = c / γ₁
+  cₖ₋₁ = cₖ = -one(T)       # Givens cosines used for the LQ factorization of Tₖ
+  sₖ₋₁ = sₖ = zero(T)       # Givens sines used for the LQ factorization of Tₖ
+  d̅ = zeros(T, n)           # Last column of D̅ₖ = Vₖ(Qₖ)ᵀ
+  ζₖ₋₁ = ζbarₖ = zero(T)    # ζₖ₋₁ and ζbarₖ are the last components of z̅ₖ = (L̅ₖ)⁻¹β₁e₁
+  ζₖ₋₂ = ηₖ = zero(T)       # ζₖ₋₂ and ηₖ are used to update ζₖ₋₁ and ζbarₖ
+  δbarₖ₋₁ = δbarₖ = zero(T) # Coefficients of Lₖ₋₁ and L̅ₖ modified during two iterations
```
dpo: Normally, we use two spaces before a comment.

src/bilq.jl
```diff
@@ -0,0 +1,216 @@
...
+      rNorm_BiLQ = sqrt(μₖ^2 * norm_vₖ^2 + ωₖ^2 * norm_vₖ₊₁^2 + 2 * μₖ * ωₖ * dot_vₖ_vₖ₊₁)
+    end
+
+    # Compute BiCG residual norm
+    # ‖rₖ‖ = |ρₖ| * ‖vₖ₊₁‖.
+    ρₖ = βₖ₊₁ * (sₖ * ζₖ₋₁ - cₖ * ζbarₖ)
+    rNorm_BiCG = abs(ρₖ) * norm_vₖ₊₁
+
+    rNorm = min(rNorm_BiLQ, rNorm_BiCG)
+    push!(rNorms, rNorm)
```
dpo: Not sure I like this. Only the BiLQ iterate is guaranteed to exist (unless there's a breakdown). `rNorms` should only contain `rNorm_BiLQ`. If one of the two residuals is smaller than the stopping tolerance, we stop. If `xCG` exists and `rNorm_BiCG` is smaller than `rNorm_BiLQ`, we transfer.

src/bilq.jl
```diff
@@ -0,0 +1,216 @@
...
+  verbose && @printf("\n")
+
+  # Compute BICG point
+  # (xᶜ)ₖ ← (xᴸ)ₖ₋₁ + ζbarₖ * d̅ₖ
+  @kaxpby!(n, one(T), x_lq, ζbarₖ, d̅)
+  x_cg = d̅
+
+  status = tired ? "maximum number of iterations exceeded" : "solution good enough given atol and rtol"
+  stats = SimpleStats(solved, false, rNorms, T[], status)
+  return (x_lq, x_cg, stats)
```
dpo: We should have a way to indicate which of `x_lq` and `x_cg` triggered the stopping condition (and we should do the same in SYMMLQ, LSLQ and LNLQ).

src/bilq.jl
```diff
@@ -0,0 +1,216 @@
...
+    end
+
+    # Compute ⟨vₖ,vₖ₊₁⟩ and ‖vₖ₊₁‖
+    dot_vₖ_vₖ₊₁ = @kdot(n, vₖ₋₁, vₖ)
+    norm_vₖ₊₁ = @knrm2(n, vₖ)
+
+    # Compute BiLQ residual norm
+    # ‖rₖ‖ = √((μₖ)²‖vₖ‖² + (ωₖ)²‖vₖ₊₁‖² + 2μₖωₖ⟨vₖ,vₖ₊₁⟩)
+    if iter == 1
+      rNorm_BiLQ = bNorm
```
dpo: How about `rNorm_lq` and `rNorm_cg` for consistency with `x_lq` and `x_cg`?

src/bilq.jl
amontoison: `solved_math` is for undeflow?

amontoison:
I update the code with your comments.
There is two remaining aspects to add / improve : 
- `solved_mach`
- verbose mode with more output

src/bilq.jl
dpo: `solved_mach` is for the current floating-point precision, e.g., fl(1 + ||r||) = fl(1).

dpo:
🤘

PR #150, opened 2019-10-16
Update variants to not convert unless necessary

### LINKED ISSUE #126: Precision is lost when using matrices
Precision is lost when using matrices.
```
A = rand(BigFloat, 3, 3)
A = A * A'
b = rand(BigFloat, 3)
x = cg(A, b)[1]
eltype(x)
```
returns `Float64`. I couldn't fix, because just adding types in `variants.jl` breaks other places.

abelsiqueira:
Closes #126


src/variants.jl
```diff
@@ -1,41 +1,80 @@
...
+        S = promote_type(S, Float64)
+      end
+      $fn(preallocated_LinearOperator(A, S), convert(Vector{S}, b), args...; kwargs...)
+    end
+
+    $fn(A :: AbstractMatrix{T}, b :: SparseVector{T}, args...; kwargs...) where T <: AbstractFloat =
+      $fn(preallocated_LinearOperator(A, T), convert(Vector{T}, b), args...; kwargs...)
+
+    $fn(A :: AbstractMatrix{T}, b :: AbstractVector{T}, args...; kwargs...) where T <: AbstractFloat =
+      $fn(preallocated_LinearOperator(A, T), b, args...; kwargs...)
```
abelsiqueira: This prevents downgrading `BigFloat`.

src/variants.jl
```diff
@@ -1,41 +1,80 @@
 using LinearAlgebra
 
 # Define a generic linear operator with preallocation
-function preallocated_LinearOperator(A)
+function preallocated_LinearOperator(A :: AbstractMatrix, T)
```
dpo: AbstractMatrix{T}?

abelsiqueira: I separated the type from the Matrix because I'm not converting the matrix - so it could be `eltype(A) = Int` - and `T` is the promoted type. I need the allocation to have a promoted type, but the matrix doesn't need to be converted.

test/test_cg.jl
```diff
@@ -70,6 +70,13 @@ function test_cg()
...
   @test(resid <= cg_tol);
   @test(stats.solved);
+
+  # Test that precision is not lost (#126)
+  A = rand(BigFloat, 3, 3)
+  A = A * A'
+  b = rand(BigFloat, 3)
+  x = cg(A, b)[1]
+  @test eltype(x) == BigFloat
 end
```
amontoison: Maybe you could update `test_mp.jl` ? We can verify that it works for each method and precision with that.

src/variants.jl
amontoison: Yep, we cannot use `PreallocatedLinearOperator` because of `Int` matrices.

test/test_cg.jl
amontoison:
```julia
function test_mp()
  @printf("Tests of multi-precision methods:\n")
  n = 5
  for fn in (:cg, :cgls, :usymqr, :cgne, :cgs, :crmr, :cg_lanczos,
             :dqgmres, :diom, :cr, :lslq, :lsqr, :lsmr, :craig,
             :craigmr, :crls, :symmlq, :minres, :cg_lanczos_shift_seq,
             :bilq, :minres_qlp, :qmr)
    @printf("%s ", string(fn))
    for T in (Float16, Float32, Float64, BigFloat)
      A = spdiagm(-1 => -ones(T,n-1), 0 => 3*ones(T,n), 1 => -ones(T,n-1))
      b = ones(T, n)
      c = - ones(T, n)
      λ = zero(T)
      if fn == :cg_lanczos_shift_seq
        shifts = [λ]
        xs = @eval $fn($A, $b, $shifts)[1]
        x = xs[1]
      elseif fn == :usymqr
        x = @eval $fn($A, $b, $c)[1]
      else
        x = @eval $fn($A, $b)[1]
      end
      atol = √eps(T)
      rtol = √eps(T)
      if T == Float16
        @test norm(A * x - b) ≤ 10 * (atol + norm(b) * rtol)
      else
        @test norm(A * x - b) ≤ atol + norm(b) * rtol
      end
      @test eltype(x) == T
    end
    @printf("✔\n")
  end
  @printf("\n")
end

test_mp()
```

amontoison: I also change
```julia
M = spdiagm(-1 => -ones(T,n-1), 0 => 3*ones(T,n), 1 => -ones(T,n-1))
A = LinearOperator(M)
```
in
```julia
A = spdiagm(-1 => -ones(T,n-1), 0 => 3*ones(T,n), 1 => -ones(T,n-1))
```

abelsiqueira:
True. Updated.

src/variants.jl
```diff
@@ -1,41 +1,80 @@
...
-      $fn(preallocated_LinearOperator(A), convert(Vector{Float64}, b), args...; kwargs...)
+    function $fn(A :: AbstractMatrix{TA}, b :: AbstractVector{Tb}, args...; kwargs...) where {TA, Tb <: Number}
+      S = promote_type(TA, Tb)
+      if S <: Integer || S <: Complex{<: Integer}
+        S = promote_type(S, Float64)
+      end
+      $fn(preallocated_LinearOperator(A, S), convert(Vector{S}, b), args...; kwargs...)
+    end
+
+    $fn(A :: AbstractMatrix{T}, b :: SparseVector{T}, args...; kwargs...) where T <: AbstractFloat =
```
amontoison: Good catch! It will improve performance of some methods with sparse `b`. We allocate some vectors with `v = b / β` but `v` should not be sparse. We can keep `b` sparse in the future but we need to check methods one by one that some vectors aren't initialize from `b`....

amontoison:
@abelsiqueira / @dpo I think you can merge this pull request. I will open a new pull request that remove `A.tprod` and it could be great if we avoid some rebase.

dpo:
Thank you both!

amontoison:
:metal:

PR #152, opened 2019-11-05
Add documentations

amontoison:


docs/src/api.md
```diff
@@ -1,23 +1,17 @@
...
 
 ```@docs
-KrylovStats
-SimpleStats
-LanczosStats
-SymmlqStats
+Krylov.KrylovStats
+Krylov.SimpleStats
+Krylov.LanczosStats
+Krylov.SymmlqStats
```
dpo: Why is it necessary to qualify everything like this?

src/Krylov.jl
```diff
@@ -23,6 +24,7 @@ mutable struct LanczosStats{T} <: KrylovStats{T}
   status :: String
 end
 
+"Type for statistics returned by SYMMLQ solver"
```
dpo: "... by the SYMMLQ solver" (or simply "... by SYMMLQ").

src/krylov_utils.jl
```diff
@@ -1,6 +1,7 @@
-# Display an array in the form
-# [ -3.0e-01 -5.1e-01  1.9e-01 ... -2.3e-01 -4.4e-01  2.4e-01 ]
-# with ndisp/2 elements on each side.
+"""Display an array in the form
+[ -3.0e-01 -5.1e-01  1.9e-01 ... -2.3e-01 -4.4e-01  2.4e-01 ]
+with ndisp/2 elements on each side.
```
dpo: (ndisp - 1)/2

docs/src/api.md
amontoison: Because we don't export them in the package. In the `REPL`, we cannot use `SimpleStats` or call
`sym_givens` without the prefix for example.

amontoison:
Ok, thanks for the review. I also changed some indentations and add a comment about the `AbstractLinearOperator` required for A.
If you want to see what the documentation looks like before merging the pull request: [build.zip](https://github.com/JuliaSmoothOptimizers/Krylov.jl/files/3812258/build.zip)

dpo:
Thank you!

PR #153, opened 2019-11-06
Update README.md

matbesancon:
Add `import Pkg` for julia >= 1.x


dpo:
Thank you! (Appveyor is so slow!!!)

PR #159, opened 2019-11-14
BiLQR & TriLQR

amontoison:
- BiLQR and TriLQR methods for solving adjoint systems Ax=b and Aᵀt = c
- ODE and PDE adjoint problems used as tests
- Generalized TriLQR for rectangular matrices
- Allows inconsistent dual systems in TriLQR (solved by USYMQR)
- Stop primal or dual system when it's solved and continues the other one
- Improve verbose mode and statuses in both methods
- Use BiLQ, QMR, USYMLQ and USYMQR residual norm estimates described in the BiLQR / TriLQR article

PR #151, opened 2019-10-25
Remove A.tprod

amontoison:
close #135 

I remove `A.tprod` but I don't understand something.
In the `REPL` with the following code :
```julia
using LinearOperators
M = rand(100,100)
x = rand(100)
A = PreallocatedLinearOperator(M)
A2 = A'
@allocated A2 * x
A3 = A'
@allocated A3 * x
```
Nothing is allocated when I compute `A3 * x` (cost of compilation of `*` method) whereas `Aᵀ * x` seems to be recompiled inside methods even if the method was compiled before.
And `Aᵀ = A'` is not allocating if the function has been "warm up"...

PR #160, opened 2019-11-20
Version 0.3.1

abelsiqueira:
We have to limit `LinearOperators` version, but also there are 69 commits since the last release.
Otherwise, we could choose an older point to make release 0.3.1.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,6 @@
 name = "Krylov"
 uuid = "ba0b0d4f-ebba-5204-a429-3ac8c609bfb7"
+version = "0.3.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"


dpo:
We would like to merge #138 and #151 before the new release.

amontoison:
If pull requests https://github.com/JuliaSmoothOptimizers/LinearOperators.jl/pull/128
is merge and https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/151 are merged we don't need to limit `LinearOperators`.
After #138  and https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/159 we wanted to release 0.4.0.

But I agree that we need a new release.

abelsiqueira:
Since it's currently breaking if simply `add Krylov`, because it tries to use `LinearOperators 0.7.0`, I'd like to limit an earlier version and release as `0.3.1`. It can be any version after `0.3.0`. I would create a branch `release-0.3` just for that, essentially. I'm doing the same with NLPModels.jl [here](https://github.com/JuliaSmoothOptimizers/NLPModels.jl/tree/release-0.9).

If you agree, where in the tree should `release-0.3` be created?

dpo:
This is as good a place as any.

PR #169, opened 2019-12-18
Poisson equation in polar coordinate

amontoison:
Example used  in the BiLQ article.

### DIFF
--- a/src/diom.jl
+++ b/src/diom.jl
@@ -29,7 +29,7 @@ function diom(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
               M :: AbstractLinearOperator=opEye(),
               N :: AbstractLinearOperator=opEye(),
               atol :: T=√eps(T), rtol :: T=√eps(T), itmax :: Int=0,
-              memory :: Int=20, verbose :: Bool=false) where T <: AbstractFloat
+              memory :: Int=20, pivoting :: Bool=false, verbose :: Bool=false) where T <: AbstractFloat
 
   m, n = size(A)
   m == n || error("System must be square")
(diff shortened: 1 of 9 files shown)


dpo:
Great! Thanks! DIOM and DQGMRES would also be good candidates.

amontoison:
You're right, I also added `DIOM` and `DQGMRES`.

I also tested CGS and the method is unable to solve it.
-> CGS: Relative residual:  6.8e+27
The condition number of A is squared with this method and with this problem the 
condition number of A is already `5 * 10^5`...

An other proof that `BiLQ` is relevant :+1:

dpo:
Of course if the BiCG iterate sometimes fails to exist, what could possibly be said about CGS?!?!? 💣 💥 💣 

That raises the question of whether a transpose-free BiLQ would be useful.

amontoison:
I should have a look first to see if it's possible. :thinking: 
It's weird that no transpose-free variants were developed for QMR and a multitude for BiCG (CGS, BiCGSTAB, TFQMR, ...).

dpo:
You'll want to read https://link.springer.com/article/10.1023/A:1011637511962.

dpo:
You know what to do :-)

amontoison:
Tests failed because of DIOM. I forgot that I used a LU factorization with row pivoting by default in DIOM, I was able to make it working but I did a mistake in the residual norm estimate that I never solved.
By default we should use a LU factorization without row pivoting such that DIOM is equivalent to CG with reorthogonalization.

dpo:
Hmmm why did the 32 bit Windows build fail and why did only one Cirrus build run?

dpo:
Actually, there's a failure on Linux too: https://travis-ci.org/JuliaSmoothOptimizers/Krylov.jl/jobs/626565071#L320

amontoison:
I increased the memory to see if it works on windows 32 bits.

dpo:
Why is only one FreeBSD running???

amontoison:
It was only a bug.

dpo:
Thank you!

PR #174, opened 2020-02-10
Add examples of SYMMLQ and MINRES-QLP

amontoison:


amontoison:
This pull request was forgotten.

examples/test_minres_qlp.jl
```diff
@@ -0,0 +1,15 @@
...
+b = ones(n)
+b_norm = norm(b)
+
+# MINRES-QLP returns the minimum-norm solution of symmetric, singular and inconsistent systems
+(x, stats) = minres_qlp(A, b);
+r = b - A * x;
+
+@printf("Residual r: %s\n", Krylov.vec2str(r))
+@printf("Relative residual norm ‖r‖: %8.1e\n", norm(r) / b_norm)
+@printf("Solution x: %s\n", Krylov.vec2str(x))
```
dpo: Let's check that x is indeed the minimum-norm solution.

examples/test_symmlq.jl
```diff
@@ -0,0 +1,15 @@
...
+b = A * ones(n)
+b_norm = norm(b)
+
+# SYMMLQ returns the minimum-norm solution of symmetric, singular and consistent systems
+(x, stats) = symmlq(A, b, transfer_to_cg=false);
+r = b - A * x;
+
+@printf("Residual r: %s\n", Krylov.vec2str(r))
+@printf("Relative residual norm ‖r‖: %8.1e\n", norm(r) / b_norm)
+@printf("Solution x: %s\n", Krylov.vec2str(x))
```
dpo: Here too.

examples/test_minres_qlp.jl
amontoison: How can I do that? A \ b returns with [1.0; 1.0; 1.0; Inf].

examples/test_symmlq.jl
amontoison: A \ b returns [1.0; 1.0; 1.0; NaN] with this problem.

examples/test_minres_qlp.jl
dpo: Your system isn't consistent. What did MINRES-QLP return here?

examples/test_symmlq.jl
dpo: Set `b = [1.0; 2.0; 3.0; 0.0]`. Then `x = [1.0; 1.0; 1.0; 0.0]`. By the way, CG should also identify the same solution.

examples/test_minres_qlp.jl
amontoison: MINRES-QLP returns [1.0; 1.0; 1.0; 0.0]. The minimum-norm solution that minimize ||b - Ax||.

dpo: Yes, that seems right. Let's just check that that's the value returned.

dpo:
Thank you.

PR #178, opened 2020-02-28
Suitable residual norm estimate in CR when a preconditioner is supplied

### LINKED ISSUE #120: Preconditioned CR
With #84, preconditioned CR doesn't work anymore even if we are not in a linesearch or trust-region context : `lineseach = false` and `radius = 0`.
#54

amontoison:
fix #120 
The problem came from the initialization of `rNorm`.


dpo:
GReat, thank you!

PR #181, opened 2020-03-11
Wrap automatically preconditioners in a LinearOperator

amontoison:
@abelsiqueira 
- Preconditioners are wrapped automatically in a LinearOperator
- `cgs`, `diom` and `dqgmres` don't allocate anymore a vector for products with transpose(A)

### DIFF
--- a/src/variants.jl
+++ b/src/variants.jl
@@ -1,13 +1,23 @@
 using LinearAlgebra
 
-# Variants where A is a matrix without specific properties
-for fn in (:cgls, :cgne, :craig, :craigmr, :crls, :crmr, :lslq, :lsmr, :lsqr, :dqgmres, :diom, :cgs, :bilq, :qmr)
+# Wrap preconditioners in a linear operator with preallocation
+function wrap_preconditioners(kwargs)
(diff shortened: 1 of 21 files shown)


src/variants.jl
```diff
@@ -1,30 +1,46 @@
 using LinearAlgebra
 
-# Define a generic linear operator with preallocation
-function preallocated_LinearOperator(A :: AbstractMatrix, T)
+# Define a basic linear operator with preallocation for A * p products
+function preallocated_basic_LinearOperator(A :: AbstractMatrix, T)
```
dpo: Can't we deduce T from the matrix A as a parameter?

amontoison: You asked the same question last year ^^
https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/150

-> Problems arrize when A is an integer matrix.

dpo: There isn't much sense in applying Krylov methods to integer matrices. I wonder if we should restrict T <: AbstractFloat or T <: Complex.

amontoison: Right, but the problem comes from the user after all. Matrice-vector products can be computed in Float64. Methods will be less efficient.

amontoison:
Maybe a could do a minor release `0.4.1` after this pull request?

dpo:
Sure. Thank you.

PR #182, opened 2020-03-13
Drop support for integer matrices

amontoison:
+ Drop support for integer matrices (#181) 
+ I also removed the variant for A and b with different type. If it's the case, we should not always convert `b` to a `Vector`. It could be a `CuVector` for example.
-> The user should give A and b with the same type.


amontoison:
Ready to be merged.

amontoison:
@dpo Can you merge this pull request? I want to rebase #181 and test it with CuArrays.

dpo:
Yes. I was waiting yesterday for the Appveyor builds to finish, which was never happening... Thank you for this.

PR #186, opened 2020-03-23
Fix missing docstring source links

amontoison:
- Private discussion with @abelsiqueira :crossed_fingers: 
-> https://github.com/JuliaDocs/Documenter.jl/issues/834#issuecomment-602854108


abelsiqueira:
Thanks.

PR #187, opened 2020-03-23
Only preallocated one vector for preconditioners

amontoison:


src/variants.jl
```diff
@@ -4,7 +4,7 @@ using LinearAlgebra
 function wrap_preconditioners(kwargs)
   if (haskey(kwargs, :M) && typeof(kwargs[:M]) <: AbstractMatrix) || (haskey(kwargs, :N) && typeof(kwargs[:N]) <: AbstractMatrix)
     k = keys(kwargs)
-    v = Tuple(typeof(arg) <: AbstractMatrix ? PreallocatedLinearOperator(arg) : arg for arg in values(kwargs))
+    v = Tuple(typeof(arg) <: AbstractMatrix ? PreallocatedLinearOperator(arg, symmetric=true) : arg for arg in values(kwargs))
```
dpo: With GMRES & co., preconditioners could be non-symmetric.

amontoison: Yes, but we only need `M⁻¹ * v` products. I did the same thing with `A` for `CGS`, `DIOM` and `DQGMRES` even if `A` is not symmetric.

dpo: I see. It's worth adding a comment around your last change to explain that.

amontoison: With the option `symmetric=true`, we preallocate matrix-vector products with A and A' with the same vector. If we don't use / need products with A' it's the good thing to do even if A is not symmetric.

src/variants.jl
```diff
@@ -4,6 +4,7 @@ using LinearAlgebra
 function wrap_preconditioners(kwargs)
   if (haskey(kwargs, :M) && typeof(kwargs[:M]) <: AbstractMatrix) || (haskey(kwargs, :N) && typeof(kwargs[:N]) <: AbstractMatrix)
     k = keys(kwargs)
+    # Matrix-vector products with Mᵀ and Nᵀ are not required, we can safely use one vector for products with M / Mᵀ and N / Nᵀ
```
dpo: Please explain that that's why you set `symmetric=true` somewhere. Otherwise, I'm sure we'll forget and puzzle over this in 6 months time.

dpo:
Thank you.

PR #190, opened 2020-04-02
Check for inconsistency in CG

### LINKED ISSUE #189: cg returns NaN for singular problem
```julia
Q = [1.0 0.0; 0.0 0.0]
g = -Q * ones(2) + [0.0; 1.0]
x, ks = Krylov.cg(Q, -g)
```
Output:
```
([NaN, NaN], 
Simple stats
  solved: false
  inconsistent: false
  residuals:  [ 1.4e+00  1.4e+00      NaN      NaN      NaN ]
  Aresiduals: []
  status: maximum number of iterations exceeded
)
```

amontoison:
close #189

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -49,7 +49,9 @@ function cg(A, b :: AbstractVector{T};
   iter = 0
   itmax == 0 && (itmax = 2 * n)
 
+  pAp = zero(T)
   rNorm = sqrt(γ)
+  pNorm² = γ
   rNorms = [rNorm;]
   ε = atol + rtol * rNorm
   verbose && @printf("%5d  %8.1e  ", iter, rNorm)
@@ -59,15 +61,14 @@ function cg(A, b :: AbstractVector{T};
   inconsistent = false
   on_boundary = false
   zero_curvature = false
-  pAp = zero(T)
 
   status = "unknown"
 
   while !(solved || tired || zero_curvature)
     Ap = A * p
     pAp = @kdot(n, p, Ap)
-    if (pAp ≤ 0) && (radius == 0)
-      if pAp == 0
+    if (pAp ≤ eps(T) * pNorm²) && (radius == 0)
+      if abs(pAp) ≤ eps(T) * pNorm²
         zero_curvature = true
         inconsistent = !linesearch
       end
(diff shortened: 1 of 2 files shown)


src/cg.jl
```diff
@@ -47,13 +47,17 @@ function cg(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
...
+  inconsistent = false
   on_boundary = false
   status = "unknown"
 
-  while ! (solved || tired)
+  while ! (solved || tired || inconsistent)
     Ap = A * p
     pAp = @kdot(n, p, Ap)
 
+    inconsistent = (pAp < eps(T)) && (radius == 0)
```
dpo, line 58: I think you can say ≤ 0 here.

dpo:
Thank you!

src/cg.jl
amontoison: `pAp = 1e-32` with the inconsistent sytem used in the tests. So it's not triggered if I use ≤ 0.
pAp values :
```julia
1.385
7013
0.00441
4652784
0.00023
64999719
9.0
1.560003254953661e-32
2.25
4.563542160821626e31
3.7023852539727136e63
3.00373176926378e95
6.092291808370931e126
1.2356635788235097e158
1.0024893935205019e190
2.0332900502699166e221
6.598403462816631e253
NaN
NaN
NaN
NaN
NaN
NaN
NaN
NaN
NaN
```

dpo: I see. The problem is that simply eps(T) doesn't scale. It should be something like eps(T) * norm(p)^2, then. We don't have norm(p) but I think it should be possible to recur it.

amontoison: If `pAp` is near 0, I compute ||p|| and I check that `pAp ≤ eps(T) * norm(p)^2` We avoid the compute of pNorm at each iteration with that.

dpo:
How about the following recurrence?

[Note Apr 1 2020 at 23_46_56.pdf](https://github.com/JuliaSmoothOptimizers/Krylov.jl/files/4418818/Note.Apr.1.2020.at.23_46_56.pdf)

dpo:
One thing we did with Marie-Ange is pass ||p|| to `to_boundary()` to avoid computing it.

amontoison:
Your notes seem right! I just need to check what should I modify if a preconditioner is provided.

amontoison:
The same relation can be used if a preconditioner is provided. We just compute ||p||_M instead of ||p||_2. I just have a problem with `to_boundary` function. If we use a preconditioner, we should compute ||x||_2 with ||p||_2 or ||x||_M with |p||_M ?
Before this pull request, everything was computed with euclidian norm.

dpo:
If there's a preconditioner.. we should use ||p||_M because the preconditioner determines the elliptical norm that defines the trust region. So it seems all good.

dpo:
Thank you!

PR #191, opened 2020-04-03
Add linesearch to CG

### LINKED ISSUE #100: Flag for enforcing positive definiteness or linesearch

amontoison:
close #100

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -25,7 +25,7 @@ function cg(A, b :: AbstractVector{T};
             itmax :: Int=0, radius :: T=zero(T), linesearch :: Bool=false,
             verbose :: Bool=false) where T <: AbstractFloat
 
-  linesearch && (radius > 0) && error("'linesearch' set to 'true' but trust-region radius > 0")
+  linesearch && (radius > 0) && error("`linesearch` set to `true` but trust-region radius > 0")
 
   n = size(b, 1)
   (size(A, 1) == n & size(A, 2) == n) || error("Inconsistent problem size")
@@ -55,17 +55,17 @@ function cg(A, b :: AbstractVector{T};
   tired = iter ≥ itmax
   inconsistent = false
   on_boundary = false
-  null_curvature = false
+  zero_curvature = false
   pAp = zero(T)
 
   status = "unknown"
 
-  while !(solved || tired || null_curvature)
+  while !(solved || tired || zero_curvature)
     Ap = A * p
     pAp = @kdot(n, p, Ap)
     if (pAp ≤ 0) && (radius == 0)
       if pAp == 0
-        null_curvature = true
+        zero_curvature = true
         inconsistent = !linesearch
       end
       if linesearch
@@ -73,7 +73,7 @@ function cg(A, b :: AbstractVector{T};
         solved = true
       end
     end
-    (null_curvature || solved) && continue
+    (zero_curvature || solved) && continue
 
     α = γ / pAp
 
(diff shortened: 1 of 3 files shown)


src/cg.jl
```diff
@@ -47,13 +51,22 @@ function cg(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
...
   on_boundary = false
+  nonpositive_curvature = false
   status = "unknown"
 
-  while ! (solved || tired)
+  while !(solved || tired || nonpositive_curvature)
     Ap = A * p
     pAp = @kdot(n, p, Ap)
 
+    if (pAp ≤ 0) && (radius == 0)
```
amontoison, line 63: I will improve it with #190

src/cg.jl
```diff
@@ -23,7 +23,11 @@ M also indicates the weighted norm in which residuals are measured.
 function cg(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
             M :: Preconditioner{T}=opEye(), atol :: T=√eps(T),
             rtol :: T=√eps(T), itmax :: Int=0, radius :: T=zero(T),
-            verbose :: Bool=false) where T <: AbstractFloat
+            linesearch :: Bool=false, verbose :: Bool=false) where T <: AbstractFloat
+
+  if linesearch && (radius > 0)
+    error("'linesearch' set to 'true' but trust-region radius > 0")
```
dpo, line 29: Let's use backticks here: `linesearch` and `true`.

src/cg.jl
```diff
@@ -47,13 +51,22 @@ function cg(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
...
   status = "unknown"
 
-  while ! (solved || tired)
+  while !(solved || tired || nonpositive_curvature)
     Ap = A * p
     pAp = @kdot(n, p, Ap)
 
+    if (pAp ≤ 0) && (radius == 0)
+      nonpositive_curvature = true
+      iter == 0 && linesearch && (x = r)
```
dpo, line 65: Should this be `x .= r`?

src/cg.jl
```diff
@@ -47,13 +51,22 @@ function cg(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
...
+      inconsistent = !linesearch
```
dpo, line 66: On second thought, we have to be careful when declaring a system inconsistent. If you solve an SQD system, you will have negative curvature but it's consistent. There must be a smarter test. Maybe if ||x|| grows without bound?

dpo: Actually, it should be ok if you check pAp in absolute value.

dpo: But we need a test that the stopping condition isn't triggered on a singular but consistent system.

src/cg.jl
```diff
@@ -60,9 +58,9 @@ function cg(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
     Ap = A * p
     pAp = @kdot(n, p, Ap)
 
-    if (pAp ≤ 0) && (radius == 0)
+    if (pAp == 0) && (radius == 0)
```
dpo, line 61: Well this will never happen. We need |pAp| ≤ eps * ||p||^2. Something like this: https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/src/cr.jl#L72

src/cg.jl
```diff
@@ -47,12 +49,27 @@ function cg(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
...
+  pAp = zeros(T)
+
   status = "unknown"
 
-  while ! (solved || tired)
+  while !(solved || tired || null_curvature)
     Ap = A * p
     pAp = @kdot(n, p, Ap)
+    if (pAp ≤ 0) && (radius == 0)
+      if pAp == 0
```
dpo, line 63: We need to merge this PR with the one where ||p|| is computed because this test should be `abs(pAp) ≤ √eps(T) * pNorm^2` (or some other tolerance).

PR #192, opened 2020-04-09
Remove restriction on A, M and N parameters types

amontoison:
@dpo @pkofod

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -43,9 +43,9 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᵀr
 A preconditioner M may be provided in the form of a linear operator and is
 assumed to be symmetric and positive definite.
 """
-function minres(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
-                M :: Preconditioner{T}=opEye(), λ :: T=zero(T),
-                atol :: T=√eps(T)/100, rtol :: T=√eps(T)/100, etol :: T=√eps(T),
+function minres(A, b :: AbstractVector{T};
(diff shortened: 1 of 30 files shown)


dpo:
Thank you for all this. Just a thought: is `opEye` still relevant here or could we just use `I`?

src/minres.jl
```diff
@@ -43,9 +43,9 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᵀr
 A preconditioner M may be provided in the form of a linear operator and is
 assumed to be symmetric and positive definite.
 """
-function minres(A :: AbstractLinearOperator{T}, b :: AbstractVector{T};
-                M :: Preconditioner{T}=opEye(), λ :: T=zero(T),
-                atol :: T=√eps(T)/100, rtol :: T=√eps(T)/100, etol :: T=√eps(T),
+function minres(A, b :: AbstractVector{T};
+                M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
```
dpo, line 47: Side question: why are atol and rtol so much smaller in minres than in other methode?

amontoison:
`opEye` is still relevant here because each products with `I` are allocating a new vector.

src/minres.jl
amontoison: Because in all solvers `atol = 1e-8` / `rtol=1e-6` before that I updated them for multiple precision support. MINRES was an exception, `rtol` and `atol` were fixed to `1e-12`, and without smaller tolorance, tests weren't passed anymore.

amontoison:
Tests with FreeBSD failed but the problems seem to come from Cirrus and not Krylov.jl

dpo:
Thank you!

PR #193, opened 2020-04-11
build on travis/FreeBSD

dpo:


amontoison:
FreeBSD is available with Travis ?! Good to know.

PR #198, opened 2020-04-20
Improve verbose mode of CG

amontoison:


amontoison:
I used `%8.1e` instead of of `%7.1e` because the value of `pAp`, `α` or `σ` can be negative with SQD systems. It's also relevant to know what represents each column.

dpo:
Thank you.

PR #203, opened 2020-07-02
Fix verbose mode for SYMMLQ and MINRES

amontoison:


src/minres.jl
```diff
@@ -208,7 +208,7 @@ function minres(A, b :: AbstractVector{T};
     Acond = γmax / γmin
 
     verbose && @printf("%5d  %7.1e  %7.1e  %7.1e  %8.1e  %8.1e  %7.1e  %7.1e\n",
-                       iter, test1, test2, β, cs, sn, ANorm, Acond)
+                       iter, rNorm, ArNorm, β, cs, sn, ANorm, Acond)
```
dpo, line 211: What's wrong with this? It prints the stopping conditions.

amontoison: It is supposed to be the value of `‖r‖` and  `‖Aᵀr‖`.
-> https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/203/files#diff-6fe66634bf178bd0b47ce2bf07f1f31eR105-L106
-> https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/203/files#diff-9be212bb8b10dfd4ef543ba8fb930f19L133-L134

dpo: I know but that's not very useful. It's helpful to see that something became very small when the solver stops.

amontoison: With a preconditioner provided, the value of `test1` and `test2` don't help me to know the behaviour of MINRES.

dpo: In what sense? rNorm is the preconditioned norm.

amontoison: `test1` oscillated between high range values. With MINRES, `‖r‖` should be monotonic and I thought that something was wrong with the preconditioning.

dpo: It's normal for the backward error to oscillate. We could print all test1, test2, rNorm and ArNorm.

dpo: Would you mind updating what's printed here?

dpo:
Thank you.

PR #204, opened 2020-07-24
Take the absolute value of \gamma_next when taking its square root in cg

ajozefiak:
I am using Krylov's cg in a package and in a small example for a unit test I am getting that `\gamma_next`, lines 101-102 of `src/cg.jl` is negative, about `-1e-35`. This value arrises due to machine error as `\gamma_next` in my example is essentially zero at this particular iteration of the cg. A sensible fix that would avoid this machine error causing problems would be to take the absolute value of `\gamma_next` before taking its square root.

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -50,7 +50,7 @@ function cg(A, b :: AbstractVector{T};
   itmax == 0 && (itmax = 2 * n)
 
   pAp = zero(T)
-  rNorm = sqrt(γ)
+  rNorm = sqrt(abs(γ))
   pNorm² = γ
   rNorms = [rNorm;]
   ε = atol + rtol * rNorm
@@ -99,7 +99,7 @@ function cg(A, b :: AbstractVector{T};
     @kaxpy!(n, -α, Ap, r)
     z = M * r
     γ_next = @kdot(n, r, z)
-    rNorm = sqrt(γ_next)
+    rNorm = sqrt(abs(γ_next))
     push!(rNorms, rNorm)
 
     solved = (rNorm ≤ ε) || on_boundary


dpo:
That's often an indication of a pretty ill-conditioned problem and/or preconditioner. Is that the case for you? What stopping tolerances are you using?

ajozefiak:
My stopping condition is a relative error of 1e-6, but my example converges in a single iteration. Additionally, the example is very small, a system of about 6 equations.

dpo:
Is there any chance you could share that system?

ajozefiak:
My Matrix is: 
```
6×6 SparseMatrixCSC{Float64,Int64} with 18 stored entries:
  [1, 1]  =  2.0
  [5, 1]  =  -1.0
  [6, 1]  =  -1.0
  [2, 2]  =  2.0
  [5, 2]  =  -2.0
  [3, 3]  =  2.0
  [5, 3]  =  -1.0
  [6, 3]  =  -1.0
  [4, 4]  =  2.0
  [6, 4]  =  -2.0
  [1, 5]  =  -1.0
  [2, 5]  =  -2.0
  [3, 5]  =  -1.0
  [5, 5]  =  4.0
  [1, 6]  =  -1.0
  [3, 6]  =  -1.0
  [4, 6]  =  -2.0
  [6, 6]  =  4.0
```

and my right hand side is:
```
 -0.08164965809277261
  0.0
  0.08164965809277261
  0.08164965809277261
  0.0
 -0.08164965809277261
```
As for the preconditioner, I am extracting the preconditioner from the `approxchol` solver from `Laplacians.jl` which is a sparse approximate cholesky factorization of my system. A single multiplication of the precondtioner solves this small system. I could try to provide the preconditioner, but it is a matrix-free factorization.

dpo:
@ajozefiak The worry is that simply taking the absolute value will ignore situations where the preconditioner isn't positive definite, which technically isn't permitted with CG.

A possible fix would be something like this (not tested):
```julia
if gamma < 0
  if gamma > -eps(T) * norm(r) * norm(z)   # eps(T) might have to be adjusted here but √eps(T) is too large
    gamma = abs(gamma)
  else
    error("Preconditioner is not positive definite")
  end
end
```

amontoison:
@ajozefiak Your matrix A is only positive semidefinite.
```julia
eigvals(Matrix(A))

6-element Array{Float64,1}:
 2.717535850685321e-16
 0.7639320225002099
 1.9999999999999996
 2.0
 5.236067977499791
 6.0
```
A has 0 as an eigenvalue.

The sparse approximate cholesky factorization should not be positive definite either.

dpo:
Closing this as I don't think it's a robust solution. Feel free to open a new PR or an issue to discuss ideas.

PR #206, opened 2020-08-02
Add regularization in CRAIGMR

amontoison:
I found a way to update x without computing Aᵀy, even if λ > 0 in CRAIGMR! With this pull request, #171 and #128 everything is now generalized for saddle-point / SQD systems.

### DIFF
--- a/src/craigmr.jl
+++ b/src/craigmr.jl
@@ -5,7 +5,7 @@
 #
 # The method seeks to solve the minimum-norm problem
 #
-#  min ‖x‖²  s.t. Ax = b,
+#  min ‖x‖  s.t.  Ax = b,
 #
 # and is equivalent to applying the conjugate residual method
 # to the linear system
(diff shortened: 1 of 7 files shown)


src/craigmr.jl
```diff
@@ -31,38 +31,45 @@ export craigmr
...
 Preconditioners M⁻¹ and N⁻¹ may be provided in the form of linear operators and are
 assumed to be symmetric and positive definite.
-Afterward CRAIGMR solves the symmetric and quasi-definite system
+If `sqd = true`, CRAIGMR solves the symmetric and quasi-definite system
 
     [ -N   Aᵀ ] [ x ]   [ 0 ]
     [  A   M  ] [ y ] = [ b ],
 
-which is equivalent to applying MINRES to (M + AN⁻¹Aᵀ)y = b.
+which is equivalent to applying MINRES to `(AN⁻¹Aᵀ + M)y = b` with `Nx = Aᵀy`.
```
amontoison, line 57: @mtanneau is it the normal equation that you use in `Tulip.jl` ?

mtanneau: If you mean "does `(AN⁻¹Aᵀ + M)y = b` have the same structure as the normal equations?", then yes.
In Tulip, `N` and `M` will be diagonal matrices with positive elements (for LP problems).

However, the right-hand side of the 2x2 system here is different from the one in Tulip, which can have non-zeros in the upper block. As a consequence, the relation `Nx = Aᵀy` does not hold in general. Instead, one has  `Nx = Aᵀy + c` where `c` is the RHS' upper block.

amontoison: Oh you're right. I forgot that the right hand side is dense in your case.

dpo: > If you mean "does (AN⁻¹Aᵀ + M)y = b have the same structure as the normal equations?", then yes.

NB: It doesn't. Normal equations would have a rhs of the form Ab.

mtanneau: Right. I looked only at the left-hand matrix.

For completeness, in Tulip:
The 2x2 augmented system has the form
```
[ -N  Aᵀ ] [x] = [c]
[  A  M  ] [y]   [b]
```
where `b` and `c` are arbitrary (dense) vectors.
After substituting `x`, the normal equations system writes
```
(AN⁻¹Aᵀ + M) y = b + AN⁻¹c
             x = N⁻¹(Aᵀy - c)
```

I hope this clarifies any confusion.

amontoison:
bump

amontoison:
It was the last rebase of this PR...
I merge the modifications.

PR #211, opened 2020-08-31
Update CI scripts

amontoison:


dpo:
Thanks. Should we at least keep 1.3 since that's the minimum version [in Project.toml](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/Project.toml#L13)?

amontoison:
Yes, good idea.

dpo:
Any idea why this errors out only on AMD? https://travis-ci.org/github/JuliaSmoothOptimizers/Krylov.jl/jobs/722902656#L558

amontoison:
No :thinking: I modified the `check_min_norm` function a little bit to solve this problem.

test/check_min_norm.jl
```diff
@@ -7,7 +7,8 @@ function check_min_norm(A, b, x; λ=0.0)
     AI = A
     xI = x
   end
-  xmin = AI' * ((AI * AI') \ b)
+  QR = qr(AI * AI')
```
dpo, line 10: Yes, the problem is probably that Julia tries to compute the LU of AA', which is singular. But we only need the QR of A', not of AA'.

amontoison: You're right !

dpo:
Still failing. How about `x = Q * (R' \ b)`?

amontoison:
I forgot to swap `R` and `Rᵀ`...
With `QR = Aᵀ`, `AAᵀ = RᵀR` and I should use the following code:
```julia
QR = qr(AI')
xmin = AI' * (QR.R \ (QR.R' \ b))
```

dpo:
Sure, although multiplying by Q should be less prone to errors than another triangular solve and multiplying by A'.

dpo:
🎉

PR #213, opened 2020-09-02
TriCG and TriMR examples

amontoison:


examples/README.md
```diff
@@ -5,3 +5,4 @@ Examples in this folder require other Julia packages:
 * [MatrixMarket.jl](https://github.com/JuliaSparse/MatrixMarket.jl)
 * [HarwellRutherfordBoeing.jl](https://github.com/JuliaSparse/HarwellRutherfordBoeing.jl)
 * [DistributedArrays.jl](https://github.com/JuliaParallel/DistributedArrays.jl)
+* [LDLFactorizations.jl](https://github.com/JuliaSmoothOptimizers/LDLFactorizations.jl)
```
dpo, line 8: It looks like there should be a `Project.toml` in this folder.

examples/test_trimr.jl
```diff
@@ -0,0 +1,64 @@
...
+K = [eye(m) A; A' -eye(n)]
+B = [b; c]
+r = B - K * [x; y]
+resid = norm(r)
+@printf("TriMR: Relative residual: %8.1e\n", resid)
+
+# [M   A] [x] = [b]
+# [Aᵀ -N] [y]   [c]
+ldlt_M = ldl(M)
+ldlt_N = ldl(N)
```
dpo, line 53: Do you need `ldl()` or is `cholesly()` sufficient for the example?

amontoison: `cholesky()` is sufficient but you need to allocate a new vector at each iteration when systems with sparse `M` and `N` are solved because `lmul!` is not implemented for `SuiteSparse.CHOLMOD` structure.

amontoison: I give an example with both because of that.

dpo:
Thank you!

PR #220, opened 2020-10-16
[CGS] Add features and improvements

amontoison:
With the help of BiCGSTAB references, I understand a little bit more the CGS method. I updated the code to add some functionalities:
- CGS is based on the Lanczos biorthogonalization process and the user can now provide his own second vector `c`. It could be also easier to compare with BiLQ or QMR.
- Use the `N` parameter instead of `M` for a right preconditioner. It allows us to stay consistent with the other Krylov methods.
- Add the possibility to use a left preconditioner `M`.
- Add an example to show how an ILU factorization can be used as a split preconditioner.


dpo:
I think in CGS we use c = r implicitly to avoid the user providing a vector c such that rho = 0.

amontoison:
c is not argument but a keyword argument. By default, `r̅₀=c=b` such that `rho  ≠ 0`.

examples/test_cgs.jl
```diff
@@ -0,0 +1,29 @@
...
+A = spdiagm(
+      -1 => fill(-1.0, n - 1), 
+       0 => fill(3.0, n), 
+       1 => fill(-2.0, n - 1)
+    )
+Id = sparse(1.0I, n, n)
+A = kron(A, Id) + kron(Id, A)
+A = kron(A, Id) + kron(Id, A)
+x = ones(N)
+b = A * x
```
amontoison, line 15: After #221, I will replace these lines by
```julia
include("../test/test_utils.jl")
A, b = kron_unsymmetric()
```

dpo: I merged #221.

dpo (approved):

dpo:
Thanks.

PR #223, opened 2020-10-21
Replace identity operator modeled with UniformScaling by opEye

### LINKED ISSUE #219: High memory allocation when using preconditioners
I have noticed some higher-than-expected memory allocations when using a preconditioner other than `opEye()`.

I only tested `minres` and `cg` so far, and the culprit seems to be a matrix-vector product with `M`, namely, in [cg](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/cd6b0512d73937c1aa4f2da92829005e00ecef10/src/cg.jl#L100) and [minres](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/cd6b0512d73937c1aa4f2da92829005e00ecef10/src/minres.jl#L146).
If I replace, e.g., `v = M * r2` in `minres` by `mul!(v, M, r2)`, the allocations are more reasonable (see example below).

```julia
using Krylov, LinearAlgebra, SparseArrays, Random
using BenchmarkTools

Random.seed!(0)
m = 2^10
A = sprand(m, m, 0.01)
S = A+A'
b = ones(m)

M1 = Krylov.opEye()
M2 = 1.0I
M3 = Diagonal(ones(m))

# Warm-up will be done during calibration
@btime minres($S, $b)         # No preconditioner
@btime minres($S, $b, M=$M1)  # M = opEye
@btime minres($S, $b, M=$M2)  # M = I (uniform scaling)
@btime minres($S, $b, M=$M3)  # M = Diagonal(1.0, ..., 1.0)
```

Current version
```
  19.906 ms (34 allocations: 114.13 KiB)
  19.902 ms (34 allocations: 114.13 KiB)
  21.170 ms (1059 allocations: 8.24 MiB)
  20.397 ms (54 allocations: 122.78 KiB)
```
and if I replace with `mul!(v, M, r2)`
```
  20.147 ms (34 allocations: 114.13 KiB)
  20.006 ms (34 allocations: 114.13 KiB)
  20.654 ms (35 allocations: 122.25 KiB)
  20.344 ms (54 allocations: 122.78 KiB)
```
Same behavior with `cg`.

EDIT: making the above modification to `minres` causes some tests to fail

amontoison:
close #219

### DIFF
--- a/src/variants.jl
+++ b/src/variants.jl
@@ -1,12 +1,27 @@
 using LinearAlgebra
 
+# Replace identity operator modelized with UniformScaling by opEye
+function wrap_uniformscaling(U)
+  if U.λ == 1
+    return opEye()
+  else
+    @printf("-------------------------------------------------------------------------------------------\n")
+    @printf("Diagonal preconditioners modeled with UniformScaling are not recommended.\n")
+    @printf("They allocate a new vector at each operator-vector products, which is not memory efficient.\n")
+    @printf("-------------------------------------------------------------------------------------------\n")
+    return U
+  end
+end
+
 # Wrap preconditioners in a linear operator with preallocation
 function wrap_preconditioners(kwargs, S)
-  if (haskey(kwargs, :M) && typeof(kwargs[:M]) <: AbstractMatrix) || (haskey(kwargs, :N) && typeof(kwargs[:N]) <: AbstractMatrix)
+  wrapM = haskey(kwargs, :M) && (typeof(kwargs[:M]) <: Union{AbstractMatrix, UniformScaling})
+  wrapN = haskey(kwargs, :N) && (typeof(kwargs[:N]) <: Union{AbstractMatrix, UniformScaling})
+  if (wrapM || wrapN)
     k = keys(kwargs)
     # Matrix-vector products with Mᵀ and Nᵀ are not required, we can safely use one vector for products with M / Mᵀ and N / Nᵀ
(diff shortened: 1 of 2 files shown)


src/variants.jl
```diff
@@ -1,12 +1,27 @@
 using LinearAlgebra
 
+# Replace identity operator modelized with UniformScaling by opEye
+function wrap_uniformscaling(U)
+  if U.λ == 1
+    return opEye()
+  else
+    @printf("-------------------------------------------------------------------------------------------\n")
+    @printf("Diagonal preconditioners modelized with UniformScaling are not recommended.\n")
```
dpo, line 9: modelized -> modeled

src/variants.jl
```diff
@@ -1,12 +1,27 @@
...
+    @printf("They allocate a new vector at each operator-vector products, which is not memory efficient.\n")
+    @printf("-------------------------------------------------------------------------------------------\n")
```
dpo, line 11: It would be useful if the message told the user to use `opEye` instead.

amontoison: We cannot modeled `λI` with `OpEye()`. We can use `λ * opEye(n)` but It also allocates a new vector at each operator-vector product.

dpo: Tell the user what to do then ;-)

amontoison: I will add a section about preconditioners in the documentation soon.
I will explain how to build an optimized Jacobi preconditioner. :)

dpo:
Thank you.

PR #225, opened 2020-10-22
An implementation of BICGSTAB

amontoison:
BICGSTAB(ℓ) is not relevant because it's quite expensive when ℓ increases and the combination of BiCG with restarted GMRES(ℓ) is worst than DQGMRES from my point of view.

I implemented the "basic" version of BICGSTAB, developed by Van Der Vorst.


dpo (approved):

dpo:
Thanks!

PR #228, opened 2020-10-30
Version 0.5.4

amontoison:


test/test_utils.jl
```diff
@@ -100,7 +100,7 @@ end
 function almost_singular(n :: Int=16)
   A = get_div_grad(n, n, n)
   A = A - 5 * I
-  b = randn(size(A, 1))
```
amontoison, line 103: It's the culprit of some random fails with Travis and AppVeyor !

dpo: I think we should eliminate all random tests.

PR #232, opened 2020-11-10
Add an attribute iterations for all variants of KrylovStats

amontoison:
- The number of iterations used by a Krylov solver is now returned.

### DIFF
--- a/src/Krylov.jl
+++ b/src/Krylov.jl
@@ -7,6 +7,7 @@ abstract type KrylovStats{T} end
 
 """
 Type for statistics returned by the majority of Krylov solvers, the attributes are:
+- iterations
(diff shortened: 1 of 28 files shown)


dpo:
What additional information does this provide that we didn't already have?

amontoison:
We don't already have the number of iterations. We can only compute it with `iterations = length(stats.residuals) - 1`.

dpo:
It can also be deduced from the number of operator-vector products. But more importantly, the number of iterations doesn't convey any relevant information in my opinion. It doesn't measure the amount of work performed. That's why we use the number of operator-vector products instead.

amontoison:
PR not relevant

PR #237, opened 2020-11-18
[BICGSTAB] Add a frequency option for the verbose mode

amontoison:
I will update the other solvers if you're fine with this syntax.
https://ci.appveyor.com/project/dpo/krylov-jl/builds/36383369/job/he4b75f2pcfveayk#L80

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -18,7 +18,7 @@ export bicgstab
 """
     (x, stats) = bicgstab(A, b::AbstractVector{T}; c::AbstractVector{T}=b,
                           M=opEye(), N=opEye(), atol::T=√eps(T), rtol::T=√eps(T),
-                          itmax::Int=0, verbose::Bool=false) where T <: AbstractFloat
+                          itmax::Int=0, verbose::Int=0) where T <: AbstractFloat
 
 Solve the square linear system Ax = b using the BICGSTAB method.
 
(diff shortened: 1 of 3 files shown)


dpo:
Just `frequency` is enough. Or `verbose` could be an int.

amontoison:
I like the second idea.
`verbose=0` -> no display
`verbose=p` with `p>=1` -> display at each multiple of p iterations

dpo:
Ok. Better say `verbose ≤ 0` -> no display.

amontoison:
I updated the PR.

src/bicgstab.jl
```diff
@@ -29,18 +29,20 @@ If BICGSTAB stagnates, we recommend DQGMRES and BiLQ as alternative methods for
 BICGSTAB stops when `itmax` iterations are reached or when `‖rₖ‖ ≤ atol + ‖b‖ * rtol`.
 `atol` is an absolute tolerance and `rtol` is a relative tolerance.
 
-Additional details can be displayed if the `verbose` mode is enabled.
+Additional details can be displayed if the verbose mode is enabled (verbose > 0).
```
dpo, line 32:
```suggestion
Additional details can be displayed if verbose mode is enabled (verbose > 0).
```

src/bicgstab.jl
```diff
@@ -29,18 +29,20 @@ If BICGSTAB stagnates, we recommend DQGMRES and BiLQ as alternative methods for
...
+The frequency of the display can be tuned, it is at each multiple of `verbose` iterations.
```
dpo, line 33:
```suggestion
Information will be displayed every `verbose` iterations.
```

dpo (changes requested):

dpo:
Thank you.

PR #238, opened 2020-11-18
Add values of default arguments in docstrings

frapac:
That allows to read the default values of the arguments with the `help?>`  function, directly in Julia's REPL. 

*Note:* By default, `maxit` is set to `2 * length(b)` whenever `maxit=0`. Should we specify this value directly in the args?

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -16,7 +16,9 @@
 export bicgstab
 
 """
-    (x, stats) = bicgstab(A, b; c, M, N, atol, rtol, itmax, verbose)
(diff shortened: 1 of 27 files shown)


amontoison:
If the matrix is rectangular `itmax=n+m` with `(n,m) = size(A)`. But we should add in the doc that `itmax=0`means `itmax=2n` if the matrix is square, otherwise `itmax=n+m`.

frapac:
I see. I will update the PR accordingly.

dpo:
Thank you!

Issue #239, opened 2020-11-18
Allow more flexibility in verbose option


frapac:
Currently, `Krylov.jl` allows to control the verbosity using the `verbose` option, common to each method. If set to `true`, the algorithm prints out the trace at each iteration. 
I am wondering if we could allow finer-grained control to the verbose option? E.g. to print the trace every 10 or 50 iterations? I see two ways forward: 
- add a new argument `verbose_it::Int` specifying how often we should print out the trace if `verbose=true`. 
- update `verbose` to allow it to take non-negative integer values (that would incur a breaking change, though)

dpo:
I would say `verbose` and `verbose_freq` are redundant: if `verbose_freq = 0`, no output should be produced. If `verbose_freq = k`, we print a line every `k` iterations.

Breaking changes are ok if they're good.

I'm interested in removing the `verbose` parameter to use logging in the future. Unfortunately, my experience is that logging slows things down, even when it's deactivated. It's worth revisiting though.

amontoison:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/237 ?

amontoison:
#237 and #240 are merged. @frapac can you close the issue?

PR #240, opened 2020-11-18
Update verbose mode

amontoison:
close  #239 
BICGSTAB was not modified to avoid conflicts with #237


dpo:
Thank you.

amontoison:
@dpo Do you know why the documentation is not updated?
I found that but I'm sure that it's a problem : https://github.com/JuliaSmoothOptimizers/Krylov.jl/runs/1425470203#step:5:23

PR #241, opened 2020-11-20
Improve performance on GPU

amontoison:
I only updated `BICGSTAB` and  `CGS` for the moment.
In the future, we should remove all occurrence of broadcast in the other solvers.

```julia
using Random, LinearAlgebra, CUDA, Krylov, BenchmarkTools
CUDA.allowscalar(false)
Random.seed!(0)
n = 2^12
A = Matrix(Symmetric(rand(n, n)))
cA = CuArray(A)
b = ones(n)
cb = CuArray(b)
@benchmark bicgstab($cA, $cb)
```

Before this pull request:
```julia
BenchmarkTools.Trial: 
  memory estimate:  121.50 MiB
  allocs estimate:  2654329
  --------------
  minimum time:     6.909 s (0.18% GC)
  median time:      6.909 s (0.18% GC)
  mean time:        6.909 s (0.18% GC)
  maximum time:     6.909 s (0.18% GC)
  --------------
  samples:          1
  evals/sample:     1
```

Now: 
```julia
BenchmarkTools.Trial: 
  memory estimate:  6.25 MiB
  allocs estimate:  368761
  --------------
  minimum time:     6.460 s (0.00% GC)
  median time:      6.460 s (0.00% GC)
  mean time:        6.460 s (0.00% GC)
  maximum time:     6.460 s (0.00% GC)
  --------------
  samples:          1
  evals/sample:     1
 ```

PR #242, opened 2020-11-23
Add benchmarks with CUDA.jl

amontoison:
I used it to see the differences with #241 
-> https://gist.github.com/amontoison/ac560351915821faa5271273ec42a9b6
-> https://gist.github.com/amontoison/3ecd4f56386b4528a31b30051709106c

I propose to merge #241 and release the version `0.6.0`.


benchmark/run_benchmarks.jl
```diff
@@ -1,7 +1,8 @@
 using GitHub, JSON, PkgBenchmark
 
-commit = benchmarkpkg("Krylov")  # current state of repository
-master = benchmarkpkg("Krylov", "master")
+filename = "gpu.jl"  # gpu.jl, cg_bmark.jl, ...
```
amontoison, line 3: If we want to use another benchmark file, we need to add a commit because `PkgBenchmark̀` cannot allow the package to be dirty when it changes branch. A solution could be to use a environment variable (ENV["BENCHMARK"] = "gpu.jl").

benchmark/gpu.jl
```diff
@@ -0,0 +1,47 @@
...
+
+using CUDA
+using CUDA.CUSPARSE
+
+using Krylov
+using LinearOperators
+using SuiteSparseMatrixCollection
+
+ufl_cg = filter(p -> p.structure == "unsymmetric" && p.type == "real" && (10000 ≤ p.rows ≤ 20000), ssmc)
+ufl_bicgstab = filter(p -> p.structure == "symmetric" && p.posDef == "yes" && p.type == "real" && (10000 ≤ p.rows ≤ 20000), ssmc)
```
dpo, line 16: These don't need to be pos.def. or even symmetric. Just square.

benchmark/run_benchmarks.jl
dpo: Yes or `run_benchmarks.jl` could be a script that takes a command-line argument.

benchmark/gpu.jl
amontoison: Oh, I swapped `ufl_cg` and `ufl_bicgstab`.

amontoison:
https://gist.github.com/amontoison/3aa8c1fd1b2ff33aa9bdbcbc35e7f9cc

dpo:
Great!

PR #247, opened 2020-11-27
Update the documentation about preconditioners on GPU

amontoison:
CUDA.jl will change the syntax for triangular solves in the near future (https://github.com/JuliaGPU/CUDA.jl/pull/575).
I updated the documentation for the future release of CUDA and I also added what should be modified if we want to use ILU and Incomplete Cholesky preconditioners with `CuSparseMatrixCSR` type.
@frapac

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -51,9 +51,9 @@ P = ic02(A_gpu, 'O')
 
 # Solve Py = x
 function ldiv!(y, P, x)
-  y .= x
-  sv2!('T', 'U', 1.0, P, y, 'O')
-  sv2!('N', 'U', 1.0, P, y, 'O')
+  copyto!(y, x)                        # Variant for CuSparseMatrixCSR
+  sv2!('T', 'U', 'N', 1.0, P, y, 'O')  # sv2!('N', 'L', 'N', 1.0, P, y, 'O')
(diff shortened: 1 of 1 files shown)


docs/src/gpu.md
```diff
@@ -76,9 +76,9 @@ P = ilu02(A_gpu, 'O')
 
 # Solve Py = x
 function ldiv!(y, P, x)
-  y .= x
-  sv2!('N', 'L', 1.0, P, y, 'O')
-  sv2!('N', 'U', 1.0, P, y, 'O', unit_diag=true)
+  copyto!(y, x)                        # Variant for CuSparseMatrixCSR
```
frapac, line 79: According to the comments, both functions are variant for `CuSparseMatrixCSR`. Is it intended?

amontoison: No,  with `CuSparseMatrixCSR`, `ilu02` returns `P = LU ≈ A` with `L` a unit lower triangular matrix and `U` an upper triangular matrix.
With `CuSparseMatrixCSC`, `ilu02` returns `P = UᵀLᵀ ≈ A` and the computation is done with `Aᵀ`. `Uᵀ` is a lower triangular matrix and `Lᵀ` a unit upper triangular matrix.

dpo:
Cool!

PR #256, opened 2021-01-06
Fix the regularization in MINRES and MINRES-QLP

### LINKED ISSUE #255: Sign of the regularization in minres
I suspect there is a bug with the sign of the regularization in `minres`. Following the doc, it should solve the shifted linear system

    (A + λ I) x = b

Taking A = I, b = 1, and λ=0.5 should return a vector with 2/3. However, it returns 2, which would correspond to put -λ.

```
using Krylov, LinearAlgebra
In = diagm(0 => ones(3))
b = ones(3)
(x, stats) = minres(In, b, λ = 0.5) # return ≈2*ones(3)
(x, stats) = minres(In, b, λ = -0.5) # return ≈2/3*ones(3)
```

amontoison:
Fixes #255

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -127,7 +127,7 @@ function minres(A, b :: AbstractVector{T};
 
     # Generate next Lanczos vector.
     y = A * v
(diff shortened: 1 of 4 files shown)


dpo:
Thank you!

PR #261, opened 2021-02-02
Discretization of poisson equation in cartesian coordinates

amontoison:
- Fix `PDE` function when `m ≠ n` (number of subdivisions in x / y)
- Fix the right-hand side `b` generated by `polar_poisson` function
- Add the function `cartesian_poisson`, a discretization of the poisson equation in cartesian coordinates with finite differences


test/test_cg.jl
```diff
@@ -92,6 +92,15 @@ function test_cg()
   A, b = square_inconsistent()
   x, stats = cg(A, b)
   @test stats.inconsistent
+
+  # Poisson equation in cartesian coordinates.
+  A, b = cartesian_poisson()
+  (x, stats) = cg(-A, -b)
```
amontoison, line 98: positive semi-definite matrix because `-A` is diagonal dominant.

dpo: We should be able to call `cg(A,b)` directly, no?

dpo:
Did this bug make it into the BiLQ test?

amontoison:
> Did this bug make it into the BiLQ test?

No, I verified and the mistake is only here in Krylov.jl.
Code used for the BiLQ test :
```julia
prec = Float64

# Discretization
M = 50
N = 50
R = one(prec)

# Poisson's equation
f(r, θ) = -3 * cos(θ)
g(r, θ) = 0

b = zeros(prec, N * M)
for j = 1 : M
  for i = 1 : N-1
      b[N*(j-1) + i] = Δr * Δr * f(r[i], θ[j])
  end
  b[N*j] = Δr * Δr * f(r[N], θ[j]) - (one(prec) + λ[N]) * g(r[N+1], θ[j])
end
```

test/test_cg.jl
amontoison: Yes, you're right. Negative curvature is allowed for symmetric quasi-definite systems. Therefore we can solve linear systems with negative definite matrices.

dpo:
> No, I verified and the mistake is only here in Krylov.jl.

🥵

> `cartesian_poisson` is the solution of an MTH8211 exercise.
> We will solve it in two weeks 🤫

🤫

> It is great to have that kind of problems in Krylov.jl.

Yes, indeed. Thank you!

PR #269, opened 2021-03-15
in-place minres

geoffroyleconte:
@amontoison


src/krylov_solvers.jl
```diff
@@ -0,0 +1,35 @@
+export KrylovSolver, MinresSolver
+
+"Abstract type for using Krylov solvers in-place"
+abstract type KrylovSolver{S, T} end
```
amontoison, line 4:
```suggestion
abstract type KrylovSolver{S} end
```

src/krylov_solvers.jl
```diff
@@ -0,0 +1,35 @@
...
+- x 
+- r1 
+- r2 
+- w1 
+- w2 
+- err_vec 
+
+The outer constructor 
+
+    solver = MinresSolver(b :: AbstractVector{T}, window :: Int=5) where T <: AbstractFloat
```
amontoison, line 18:
```suggestion
    solver = MinresSolver(A, b)
```

src/krylov_solvers.jl
```diff
@@ -0,0 +1,35 @@
...
+
+"""
+Type for storing the vectors required by the in-place version of MINRES.
+The attributes are:
+- x 
+- r1 
+- r2 
+- w1 
+- w2 
+- err_vec
```
amontoison, line 14:
```suggestion
- x
- r1
- r2
- w1
- w2
```

src/krylov_solvers.jl
```diff
@@ -0,0 +1,35 @@
...
+may be used in order to create these vectors.
+"""
+mutable struct MinresSolver{S, T} <: KrylovSolver{S, T}
+  x       :: S
+  r1      :: S
+  r2      :: S
+  w1      :: S 
+  w2      :: S
+  err_vec :: Vector{T}
+end
```
amontoison, line 29:
```suggestion
mutable struct MinresSolver{S} <: KrylovSolver{S}
  x       :: S
  r1      :: S
  r2      :: S
  w1      :: S 
  w2      :: S
end
```

src/krylov_solvers.jl
```diff
@@ -0,0 +1,35 @@
...
+
+function MinresSolver(b :: AbstractVector{T}, window :: Int=5) where T <: AbstractFloat
+  n = length(b)
+  S = typeof(b)
+  return MinresSolver{S, T}(kzeros(S, n), kzeros(S, n), kzeros(S, n), kzeros(S, n), kzeros(S, n), zeros(T, window))
+end
```
amontoison, line 35:
```suggestion
function MinresSolver(A, b)
  m, n = size(A)
  S = typeof(b)
  return MinresSolver{S}(S(undef, n), S(undef, n), S(undef, n), S(undef, n), S(undef, n))
end
```

src/minres.jl
```diff
@@ -49,15 +49,27 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᵀr
...
 
+It is also possible to use MINRES in-place
+
+    (x, stats) = minres!(A, b :: AbstractVector{T}, solver :: MinresSolver{S, T};
+                         M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
+                         rtol :: T=√eps(T)/100, etol :: T=√eps(T),
+                         itmax :: Int=0, conlim :: T=1/√eps(T),
+                         verbose :: Int=0) where {S, T <: AbstractFloat}
+
+where `solver` is a [`Krylov.MinresSolver`](@ref) used to store the vectors used by `minres!`.
```
amontoison, line 60:
```suggestion
    (x, stats) = minres!(A, b :: AbstractVector{T}, solver :: MinresSolver{S};
                         M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
                         rtol :: T=√eps(T)/100, etol :: T=√eps(T),
                         itmax :: Int=0, conlim :: T=1/√eps(T),
                         verbose :: Int=0) where {S, T <: AbstractFloat}

where `solver` is a [`Krylov.MinresSolver`](@ref) used to store the vectors used by `minres!`.
```

src/minres.jl
```diff
@@ -68,16 +80,17 @@ function minres(A, b :: AbstractVector{T};
...
 
-  # Determine the storage type of b
-  S = typeof(b)
+  # get data from solver
+  x, r1, r2, w1, w2, err_vec = solver.x, solver.r1, solver.r2, solver.w1, solver.w2, solver.err_vec
+  window = length(err_vec)
 
   ϵM = eps(T)
-  x = kzeros(S, n)
+  x .= 0
```
amontoison, line 88:
```suggestion
  x .= zero(T)
```

src/minres.jl
```diff
@@ -110,7 +123,7 @@ function minres(A, b :: AbstractVector{T};
 
   xENorm² = zero(T)
   err_lbnd = zero(T)
-  err_vec = zeros(T, window)
+  err_vec .= 0
```
amontoison, line 126:
```suggestion
  err_vec .= zeros(T, window)
```

src/minres.jl
```diff
@@ -49,15 +49,27 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᵀr
...
 #### Reference
 
 * C. C. Paige and M. A. Saunders, *Solution of Sparse Indefinite Systems of Linear Equations*, SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
 """
-function minres(A, b :: AbstractVector{T};
-                M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
-                rtol :: T=√eps(T)/100, etol :: T=√eps(T),
-                window :: Int=5, itmax :: Int=0, conlim :: T=1/√eps(T),
-                verbose :: Int=0) where T <: AbstractFloat
+minres(A, b :: AbstractVector{T}; window :: Int=5, kwargs...) where T <: AbstractFloat = minres!(A, b, MinresSolver(b, window); kwargs...)
```
amontoison, line 66:
```suggestion
@inline minres(A, b :: AbstractVector{T}; kwargs...) where T <: AbstractFloat = minres!(A, b, MinresSolver(A, b); kwargs...)
```

src/minres.jl
```diff
@@ -49,15 +49,27 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᵀr
...
+
+function minres!(A, b :: AbstractVector{T}, solver :: MinresSolver{S, T};
+                 M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
+                 rtol :: T=√eps(T)/100, etol :: T=√eps(T),
+                 itmax :: Int=0, conlim :: T=1/√eps(T),
+                 verbose :: Int=0) where {S, T <: AbstractFloat}
```
amontoison, line 72:
```suggestion
function minres!(A, b :: AbstractVector{T}, solver :: MinresSolver{S};
                 M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
                 rtol :: T=√eps(T)/100, etol :: T=√eps(T),
                 itmax :: Int=0, conlim :: T=1/√eps(T),
                 verbose :: Int=0) where {S, T <: AbstractFloat}
```

amontoison (changes requested):
Great PR !!!

amontoison:
if you want to rebase : 
```
git remote add upstream https://github.com/JuliaSmoothOptimizers/Krylov.jl.git
git fetch upstream
git rebase upstream/master
git push -f
```

amontoison:
@geoffroyleconte You were right, we need both `S` and `T` for `KrylovSolver{S, T}`. For some methods we cannot avoid the storage of vectors on CPU (`DIOM`, `DQGMRES`, ...). We can keep `err_vec` after all for `minres` but we need to check at the beginning that `length(solver.err_vec) == windows`.

amontoison (approved):

amontoison:
@dpo It's fine for me. What do you think about it ?

src/krylov_solvers.jl
```diff
@@ -0,0 +1,34 @@
...
+abstract type KrylovSolver{T, S} end
+
+"""
+Type for storing the vectors required by the in-place version of MINRES.
+The attributes are:
+- x 
+- r1 
+- r2 
+- w1 
+- w2
```
dpo, line 13: This doesn’t seem important/useful to document since the user is never supposed to touch this structure. I would remove lines 9-13 and add comments inside the struct to guide the developer.

test/test_minres.jl
```diff
@@ -66,6 +66,17 @@ function test_minres()
...
+  # in-place minres (minres!) with Jacobi (or diagonal) preconditioner
+  A, b, M = square_preconditioned()
+  solver = MinresSolver(b)
+  (solver.x, stats) = minres!(A, b, solver, M=M)
+  show(stats)
+  r = b - A * solver.x
+  resid = norm(r) / norm(b)
+  @printf("MINRES: Relative residual: %8.1e\n", resid)
+  @test(resid ≤ minres_tol)
+  @test(stats.solved)
```
dpo, line 79: It would be great to add a test showing that `minres!(A, b, solver)` does not allocate.

dpo (changes requested):

src/krylov_solvers.jl
geoffroyleconte: The user may need to access `x` no? And maybe other parameters if we create the same struct for other Krylov solver?

test/test_minres.jl
geoffroyleconte: I think this is not possible yet because there are still allocs done with the Residuals and Aresiduals vectors (see https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/272) . Even after the PR there will still be allocs because we create empty vectors for Residuals and Aresiduals and we do not fill them.

dpo:
I think we need to separate cleanly the Minres storage from the stats structure. I think x should be in the stats structure so that the user never needs to look into the solver private storage. We should have an option to not accumulate the arrays of rNorms and ArNorms and be able to check that when that option is turned on, the solver doesn’t allocate.

geoffroyleconte:
If x is in the stats structure, we will need to add it in the arguments of minres! . Do you mean something like
```minres!(A, b, solver, x; ....)``` ?

File: src/krylov_solvers.jl (hunk 11 above)
Comment (amontoison) [reply src/krylov_solvers.jl:13]: `x` is always returned by a Krylov method and it's already a pointer to a vector inside a `KrylovSolver`.

**Comment by amontoison** [comment]
Can you rebase your branch Geoffroy ?

**Comment by amontoison** [comment]
It's not possible to remove all the allocation because we have some pointers that must be stored but you can add a test here https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/test/test_alloc.jl#L45-L48
that check `@allocated minres!(solver, A, b) <= 1000)`

**Comment by dpo** [comment]
Store the stats structure inside MinresSolver and x inside the stats. Then just return the stats instead of (x, stats).

If the operators are preallocated, it should be possible to remove all allocations.

**Comment by amontoison** [comment]
@dpo It will break every code that use `Krylov.jl` and stats.rNorms and stats.ArNorms have a different length at each call of a Krylov method if `history = true`.
I don't think that it's a good idea.

**Comment by dpo** [comment]
> It will break every code that use Krylov.jl

I don’t think so. If you want, you can return (stats.x, stats) for the time being. What will break? At any rate, that’s what version numbers are for. Overall I think it’s an improvement.

> stats.rNorms and stats.ArNorms have a different length at each call of a Krylov method if history = true.

I think the objective of this should be to remove allocations. What I’m saying is that if history = false and the operators preallocate, we should be able to have no allocations in the solver itself.

**Comment by geoffroyleconte** [comment]
@dpo , with @amontoison we think that it might be wise to keep this PR as it is for now, and to focus on removing the Residuals and Aresiduals in another PR. The ultimate goal is, as you said, to remove all allocations with `minres!`, maybe in version 0.8 .

**Comment by amontoison** [comment]
@dpo We should think about a generic `stats` for all solvers. It will also solve #166. 
Let's do it step by step.

File: src/minres.jl [hunk 13]
```diff
@@ -49,15 +49,28 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᵀr
 A preconditioner M may be provided in the form of a linear operator and is
 assumed to be symmetric and positive definite.
 
+It is also possible to use MINRES in-place
+
+    (x, stats) = minres!(A, b :: AbstractVector{T}, solver :: MinresSolver{S, T};
```
amontoison, line 54:
```suggestion
    (x, stats) = minres!(solver :: MinresSolver{S, T}, A, b :: AbstractVector{T};
```

dpo:
> We should think about a generic stats for all solvers.

That’s a different topic. We’re just doing this for MINRES for now. Generic stats can come later (maybe).

amontoison:
> > We should think about a generic stats for all solvers.
> 
> That’s a different topic. We’re just doing this for MINRES for now. Generic stats can come later (maybe).

Yes, sure. But if you want to add `x` in `stats` like you asked, Geoffroy needs to update `SimpleStats` (which require to update all the other methods that use it) or create a new `stats` object and implement its `show` function.
We should discuss about it together friday.

dpo:
Sure, let’s duplicate SimpleStats to make one just for MINRES for now. This is our proof of concept. It’s not that difficult. No need to implement show at this point. The main point is: can we get rid of all allocations in the second call?

amontoison:
@geoffroyleconte 
We need to add `minres!` here : https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/src/variants.jl#L57-L65

amontoison:
> Sure, let’s duplicate SimpleStats to make one just for MINRES for now. This is our proof of concept. It’s not that difficult. No need to implement show at this point. The main point is: can we get rid of all allocations in the second call?

Yes, it will be the case for MINRES (and all methods that do not require A').

dpo:
Let’s do it so @geoffroyleconte can use it in his code.

amontoison:
@geoffroyleconte
You can continue without me.

test/test_alloc.jl
```diff
@@ -47,6 +47,11 @@ minres(A, b)  # warmup
 actual_minres_bytes = @allocated minres(A, b)
 @test actual_minres_bytes ≤ 1.1 * expected_minres_bytes
 
+opA, solver = MinresSolver(A, b)
+minres!(solver, opA, b)  # warmup
+inplace_minres_bytes = @allocated minres!(solver, opA, b)
+@test inplace_minres_bytes == 240
```
amontoison, line 53: @dpo @geoffroyleconte 
The last allocation is the `KrylovStats`.

dpo: 240 bytes!!! Much better!!! Thanks guys. Conclusion: anything is possible.

Let’s see if we can get to zero by putting the `SimpleStats` inside `MinresSolver`, and `x` inside `SimpleStats` (define a copy `MinresStats` of `SimpleStats` that also contains an `x`).

test/test_alloc.jl
```diff
@@ -47,6 +47,11 @@ minres(A, b)  # warmup
 actual_minres_bytes = @allocated minres(A, b)
 @test actual_minres_bytes ≤ 1.1 * expected_minres_bytes
 
+opA, solver = MinresSolver(A, b)
+minres!(solver, opA, b)  # warmup
+inplace_minres_bytes = @allocated minres!(solver, opA, b)
+@test inplace_minres_bytes == 32
```
geoffroyleconte, line 53: I am now down to 32 bytes but it seems I cannot go lower

amontoison: Comment this part of code https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/269/files/ac17b26c9081c7a7fb3168452aeca9976c41e679..98caea1db74a85ec478f6f9ec4dbea26a08df282#diff-af32b9d734e2fe1b702cc5c0d7fc714c5c978e2ee4fc0bfb449a68d837b63453R67-R80 to see if the 32 bytes comes from the `wrap_preconditioners` function.

amontoison:
A last issue that we have is how to reuse the vector preallocated by `PreallocatedLinearOperator`.
The only easy solution that I see is stored the `y` in the `MinresSolver` and use `mul!(y, A, v)`.
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/src/minres.jl#L133

Otherwise Geoffroy will need to create a new `PreallocatedLinearOperator` (which allocates 1 vector) at each iteration of its IPM method.

dpo:
Couldn't he simply preallocate a vector before his main loop and create a new PreallocatedLinearOperator inside the loop that uses that vector as storage?

test/test_alloc.jl
geoffroyleconte: If you mean to comment L67-L81 of variants.jl, it does not remove the 32 bytes

amontoison:
> Couldn't he simply preallocate a vector before his main loop and create a new PreallocatedLinearOperator inside the loop that uses that vector as storage?

You're right, I forgot that we have that kind of constructor : https://github.com/JuliaSmoothOptimizers/LinearOperators.jl/blob/master/src/PreallocatedLinearOperators.jl#L82

geoffroyleconte:
Could we replace `*` from LinearOperators by `mul!` ? This way we would not have to create every time a new LinearOperator if `A` is a matrix.

dpo:
@geoffroyleconte That's a great idea and we've discussed that in the past. I think LinearOperators could use some attention and be based on `mul!()`. @amontoison Didn't you concoct a proof of concept at some point?

geoffroyleconte:
There is nothing to change in LinearOperators for this I think. If `A` is a matrix then `MinresSolver` preallocates 1 vector, and `minres!` uses the normal `mul!` . Otherwise, it uses this function: https://github.com/JuliaSmoothOptimizers/LinearOperators.jl/blob/ea38e7539ab036e748f126b98a2c0076131a798c/src/operations.jl#L20 and stores the preallocated vector by the `PreallocatedLinearOperator` in `MinresSolver` . 
The only issue is if `A` is a LinearOperator without allocation (like OpEye), but maybe we can find something to make it work?

amontoison:
Yes, I did a macro such that `@.. y = A * x` dispatch to `mul!(y,  A, x)` but the main problem was the implementation of `mul!` in `LinearOperators`.

geoffroyleconte:
This PR is getting more complicated than it should be. I think `minres!` should not force the user to work with a `LinearOperator{T}` , maybe even the whole package. It should also be able to work with an `AbstractMatrix{T}` as input, and use the  `mul!` function from LinearAlgebra. 
I suggest we modify this PR so that `minres!` works with `AbstractMatrix{T}` as input (otherwise I will have to write my own `minres!` function and add it to RipQP because I do not need LinearOperators for now). Then we pause it, and make some modifications to LinearOperators to implement the `mul!` function and deal with allocations issues.
This way, we would have a `minres!` function that takes any `AbstractMatrix{T}` or `LinearOperator{T}` as input, and would only specialise when using `mul!`.

src/krylov_solvers.jl
```diff
@@ -0,0 +1,37 @@
...
+function MinresSolver(A, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
+  n, m = size(A)
+  S = typeof(b)
+  x  = S(undef, n)
+  r1 = S(undef, n)
+  r2 = S(undef, n)
+  w1 = S(undef, n)
+  w2 = S(undef, n)
+  err_vec = zeros(T, window)
+  opA = typeof(A) <: AbstractMatrix ? PreallocatedLinearOperator(A, storagetype=S, symmetric=true) : A
```
dpo, line 34: you don't want to do that because you want to be able to call `minres!()` later with a different operator and a different `b` that have the same dimensions.

src/krylov_solvers.jl
```diff
@@ -0,0 +1,37 @@
...
+  w1      :: S 
+  w2      :: S
+  err_vec :: Vector{T}
+  stats   :: SimpleStats{T}
+end
+
+function MinresSolver(A, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
+  n, m = size(A)
+  S = typeof(b)
+  x  = S(undef, n)
```
dpo, line 28: I think this should be `x = similar(b, n)` because it could result in an error:
```julia
julia> b1 = rand(10); b = view(b1, 1:2:10)
julia> S = typeof(b)
SubArray{Float64,1,Array{Float64,1},Tuple{StepRange{Int64,Int64}},true}

julia> S(undef, 5)
ERROR: MethodError: no method matching SubArray{Float64,1,Array{Float64,1},Tuple{StepRange{Int64,Int64}},true}(::UndefInitializer, ::Int64)
Closest candidates are:
  SubArray{Float64,1,Array{Float64,1},Tuple{StepRange{Int64,Int64}},true}(::Any, ::Any, ::Any, ::Any) where {T, N, P, I, L} at subarray.jl:19
```

src/krylov_solvers.jl
```diff
@@ -22,16 +22,14 @@ mutable struct MinresSolver{T, S} <: KrylovSolver{T, S}
...
 
-function MinresSolver(A, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
-  n, m = size(A)
-  S = typeof(b)
-  x  = S(undef, n)
-  r1 = S(undef, n)
-  r2 = S(undef, n)
-  w1 = S(undef, n)
-  w2 = S(undef, n)
+function MinresSolver(opA :: PreallocatedLinearOperator{T}, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
```
amontoison, line 25: If someone wants to use `LinearMaps` or its own linear operator it should be able to do it. It's for that reason that we removed the type restrictions on `A` , `M` and `N` for all methods.

src/krylov_solvers.jl
```diff
@@ -22,16 +22,14 @@ mutable struct MinresSolver{T, S} <: KrylovSolver{T, S}
...
+  m, n = opA.nrow, opA.ncol
```
amontoison, line 26: Keep the generic `m, n = size(A)`

amontoison (changes requested):

src/krylov_solvers.jl
amontoison, line 25:
```suggestion
function MinresSolver(A, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
```

amontoison:
I like this implementation. It will break nothing !

src/krylov_solvers.jl
```diff
@@ -0,0 +1,35 @@
...
+
+"Abstract type for using Krylov solvers in-place"
+abstract type KrylovSolver{T, S} end
+
+"""
+Type for storing the vectors required by the in-place version of MINRES.
+
+The outer constructor 
+
+    opA, solver = MinresSolver(A, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
```
amontoison, line 11:
```suggestion
    solver = MinresSolver(A, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
```

dpo (changes requested):

test/test_alloc.jl
dpo: https://docs.julialang.org/en/v1/manual/profile/#Memory-allocation-analysis

src/krylov_solvers.jl
geoffroyleconte: OK, but in this case it shall be expected that the user does not give an `AbstractMatrix` to `MinresSolver`.

amontoison: Why ?

geoffroyleconte: Sorry for `minres!` I should say, otherwise the vectors created by the `PreallocatedLinearOperator` cannot be reused when using `minres!` multiple times

amontoison: Yes you're right, but it's the problem of the user in that case.

dpo: Exactly. It's the user's responsibility (at least for now).

amontoison:
I found the culprit, Julia cannot interfered the type of a `MinresSolver` when allocation tests are performed in the REPL. To avoid this issue I wrapped the tests in a function `test_alloc()`.

dpo:
Thank you. Tests should always be in functions It seems like it isn’t working with Julia 1.3 though.

geoffroyleconte:
Thank you! Should we upgrade the Julia compat?

amontoison:
I think that I can help a little bit Julia 1.3 and type inference. I will have a look ASAP.

amontoison:
allocs

We cannot remove the 32 bits with Julia < 1.5. I propose to update the allocation test like this https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/280/files#diff-8da4f5f3a63cefac6bb5115d0b873b95246ef210751fb7a8269141807e98870dR141

I can rebase the PR tomorrow if it's fine for you. I also found some minor issues if `history=true` with the last commits of Geoffroy. I can solve it in the same time.

geoffroyleconte:
It is fine for me, thank you!

dpo:
Instead of zero bytes, I suggest measuring the amount allocated with Julia 1.5 and using that as the benchmark (after making sure it matches the "theory").

geoffroyleconte:
@dpo I don't understand

dpo:
I was replying to @amontoison’s proposal.

amontoison:
> Instead of zero bytes, I suggest measuring the amount allocated with Julia 1.5 and using that as the benchmark (after making sure it matches the "theory").

I'm sure to understand the differences with what I proposed and I did for BICGSTAB ? 
https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/280/files#diff-8da4f5f3a63cefac6bb5115d0b873b95246ef210751fb7a8269141807e98870dR129-R141

geoffroyleconte:
@dpo @amontoison can we merge this if I check that the Julia version is 1.5 for the allocs test with `minres!` ? It is very inconvenient to have to rebase every time... We can open an issue to make allocs tests for other versions later.

dpo:
> I don't understand. What are the differences with what I proposed and I did for BICGSTAB ?

I didn’t realize that you simply deactivated the test for Julia ≤ 1.5. That’s ok I guess.

dpo:
Also it looks like the coverage went down. I think we just need to call the variants in the tests: A = dense matrix, b = sparse vector, etc.

dpo:
@geoffroyleconte Could you deactivate the MINRES allocation test for Julia ≤ 1.5 as @amontoison did? The tests should pass with Julia 1.3.

test/test_variants.jl
```diff
@@ -27,6 +27,13 @@ function test_variants()
                 @eval $fn($transpose($A), $b, $c)
                 @eval $fn($adjoint($A), $b, $c)
               end
+            elseif fn == :minres! 
+              if b != b_sparse
+                solver = eval(Symbol(uppercasefirst(string(fn))[1:end-1], "Solver"))(A, b)
+                @eval $fn($solver, $A, $b)
+                @eval $fn($solver, $transpose($A), $b)
+                @eval $fn($solver, $adjoint($A), $b)
```
dpo, line 35: See the annotations in `variants.jl` to see which are not covered by the current tests.

geoffroyleconte: There is an issue when `b` is a `SparseVector`, I did not find how to fix it. But anyway the function converts `b`to a `Vector` so this is not really an issue, the user just has to give a `Vector`instead of a `SparseVector`. All these issues will be fixed when we change LinearOperators

dpo: I understand, but what we’re trying to fix now is coverage. What’s the problem with SparseVector?

dpo: Coverage is completely separate from the in-place solver. We need to test the variants simply to ensure that they are callable and behave correctly. It's understood that they will allocate. But allocations are not what we're testing. If you're having an issue with `SparseVector`s, then it's an indication that there's a bug and we should fix it.

Could you paste the error here?

geoffroyleconte: There is no issue with `minres`, only with `minres!`. It looks like the coverage is increased. We are loosing time on minor issues that will be fixed with the changes that will be made on LinearOperators.jl. We should say that for now the user has to use PreallocatedLinearOperators. It is actually the only way he can avoid allocations with the current implementation. I can add this to the documentation if needed.

Here is the error:

`minres! ERROR: LoadError: LoadError: TypeError: in typeassert, expected SparseVector{Float32,Int32}, got a value of type Array{Float32,1}
Stacktrace:
 [1] *(::PreallocatedLinearOperator{Float32}, ::SparseVector{Float32,Int32}) at C:\Users\Geoffroy Leconte\.julia\packages\LinearOperators\qFko7\src\operations.jl:7
 [2] minres!(::MinresSolver{Float32,SparseVector{Float32,Int32}}, ::PreallocatedLinearOperator{Float32}, ::Array{Float32,1}; M::opEye, λ::Float32, atol::Float32, rtol::Float32, ratol`

geoffroyleconte:
I added annotations so that the user is forced to use `PreallocatedLinearOperators` with `minres!`. 

I can raise an issue to deal with `b` when it is a `SparseVector`. This may be solved with https://github.com/JuliaSmoothOptimizers/LinearOperators.jl/pull/172 , but if it works it will change all the Krylov variants with `b::SparseVector{T}` so we should deal with this in another PR.

src/krylov_solvers.jl
```diff
@@ -22,7 +23,7 @@ mutable struct MinresSolver{T, S} <: KrylovSolver{T, S}
   stats   :: SimpleStats{T}
 end
 
-function MinresSolver(A, b :: AbstractVector{T}; window :: Int=5) where T <: AbstractFloat
+function MinresSolver(A :: PreallocatedLinearOperator{T}, b :: Vector{T}; window :: Int=5) where T <: AbstractFloat
```
amontoison, line 26: We cannot use anymore `minres` on GPU :(

dpo: I thought the annotation should go in `minres!()`, not here. Would that also break the GPU?

amontoison: `b` must be an `AbstractVector{T}` to keep support with `CuVector{T}`. (GPU storage).
The parametrization of A is not needed too : https://juliasmoothoptimizers.github.io/Krylov.jl/dev/matrix-free/

But one thing should be explained somewhere, if the user want to use `minres!`, he should wrap itselft A::AbstractMatrix in a PreallocatedLinearOperator. It's only done automatically with `minres`.
If the user already uses its linear operator (`LinearOperators.jl`, `LinearMaps.jl`, ...) he don't need to do anything.

geoffroyleconte: OK, I removed annotations and documented that the user has to use PreallocatedLinearOperators to avoid allocations.

amontoison (approved):

dpo:
Great! Thank you!

Issue #264, opened 2021-03-15
Preallocate workspace for solvers


dpo:
### Krylov methods

- [x] CG
- [x] CR
- [x] SYMMLQ
- [x] CG-LANCZOS
- [x] CG-LANCZOS-SHIFT-SEQ
- [x] MINRES
- [x] MINRES-QLP
- [x] DQGMRES
- [x] DIOM
- [x] USYMLQ
- [x] USYMQR
- [x] TRICG
- [x] TRIMR
- [x] TRILQR
- [x] CGS
- [x] BICGSTAB
- [x] BILQ
- [x] QMR
- [x] BILQR
- [x] CGLS
- [x] CRLS
- [x] CGNE
- [x] CRMR
- [x] LSLQ
- [x] LSQR
- [x] LSMR
- [x] LNLQ
- [x] CRAIG
- [x] CRAIGMR

For MINRES, this could take the form (not tested)
```julia
function minres(A, b :: AbstractVector{T}; kwargs...) where T <: AbstractFloat
  x = similar(b)
  r1 = similar(b)
  window = kwargs.get(:window, 5)
  err_vec = Vector{T}(undef, window)
  minres!(A, b, x, r1, err_vec; kwargs...)
end

function minres!(A, b, x, r1, err_vec;
                M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
                rtol :: T=√eps(T)/100, etol :: T=√eps(T),
                window :: Int=5, itmax :: Int=0, conlim :: T=1/√eps(T),
                verbose :: Int=0) where T <: AbstractFloat
  ...
  x .= 0
  r1 .= b
  err_vec .= 0
  ...
end
```
The objective is for `minres!()` to not allocate at all. We could also define
```julia
abstract type KrylovSolver end
mutable struct MinresSolver <: KrylovSolver
  x
  r1
  err_vec
  # maybe default parameter values...
  function MinresSolver(A, b)
    new(...)
  end
end
```
and pass an instance of `MinresSolver` around.

@amontoison @geoffroyleconte

amontoison:
Speak of the devil ! We were talking about it this afternoon with Geoffroy.

dpo:
😈

amontoison:
I have something in mind, I will open a pull request when it's ready. It should not break GPU support, which a little bit tricky.

amontoison:
@geoffroyleconte 
I let you try to add this feature for MINRES.
[...]

PR #271, opened 2021-03-16
Add a KrylovSolver for each method

amontoison:
#264 
@geoffroyleconte After #269 I will rebase this PR and we could easily add in-place method for all krylov solvers :rocket: 

For all solver, the syntax is the same and we can do
```
KnameSolver(A, b)
kname!(solver, A, b)
```
to reuse the storage, which is quite user-friendly.

### DIFF
--- a/src/Krylov.jl
+++ b/src/Krylov.jl
@@ -3,6 +3,7 @@ module Krylov
 using LinearOperators, LinearAlgebra, SparseArrays, Printf
 
 include("krylov_stats.jl")
+include("krylov_solvers.jl")
 include("krylov_utils.jl")
 
 include("cg.jl")
--- a/src/krylov_solvers.jl
+++ b/src/krylov_solvers.jl
@@ -1,0 +1,631 @@
+abstract type KrylovSolver{T,S} end
+
+mutable struct CgSolver{T,S} <: KrylovSolver{T,S}
+  x :: S
+  r :: S
+  p :: S
+end
+
+function CgSolver(A, b)
+  n, m = size(A)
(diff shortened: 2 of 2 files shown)


amontoison:
Some benchmarks are needed to see if `struct` is not more relevant than `mutable struct`.

dpo:
Before launching ahead, we should coordinate with @abelsiqueira and https://github.com/JuliaSmoothOptimizers/SolverCore.jl/pull/6, where he’s proposing an `AbstractSolver` struct that we should be able to import here as a base solver.

amontoison:
@abelsiqueira For the solver revamp we only need to work on the KrylovSolvers : https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/a4c91b8c7f3f37e5de9eac9e4b7edf725e9c687c/src/krylov_solvers.jl

amontoison:
@tmigot, @dpo, @geoffroyleconte 
The PR is ready to be review, don't be afraid of the number of modifications.

**Note** : I defined a new function called `ktypeof` that returned the type of a vector `v` if it's dense or a dense storage type if it's sparse (`SparseVector` / `CuSparseVector`) or a `SubArray` (#290). With this new function we don't need to convert the rhs `b` (or `c`) anymore and it removes some variants in `src/variants.jl`.

amontoison:
> Thank you for all these changes! Maybe you could add the `stats` to the `Solver` type in order to remove more allocations, like it was done with MINRES?

I would like to use a generic `KrylovStats` for all Krylov methods before and avoid to modify all files 2 times.

PR #272, opened 2021-03-16
History MINRES parameter to return empty residuals and Aresiduals

geoffroyleconte:
Removing empty vectors may speed up MINRES, especially if it is used in a loop in-place.
When history = false, we have to return empty vectors because they are required by the `SimpleStats` type.


amontoison (approved):
The last commit is perfect !
We just need to rebase this PR after #269.
Can you do the same modification with the other solvers in a separate branch ?

PR #279, opened 2021-03-31
MINRES stopping conditions

geoffroyleconte:
Hello, I found that MINRES was stopping early when I tested it with RipQP, and the residuals were not close to 0, because of the `test1` test.

I changed the test so that MINRES does not stop too early, however I still have to normalize `b` when solving `Ax=b`. I left the `test1` values displayed by the logger in case they are needed, but I can remove them.

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -25,7 +25,8 @@ export minres
 """
     (x, stats) = minres(A, b::AbstractVector{T};
                         M=opEye(), λ::T=zero(T), atol::T=√eps(T)/100,
-                        rtol::T=√eps(T)/100, etol::T=√eps(T),
+                        rtol::T=√eps(T)/100, ratol :: T=√eps(T), 
+                        rrtol :: T=√eps(T), etol::T=√eps(T),
                         window::Int=5, itmax::Int=0, conlim::T=1/√eps(T),
                         verbose::Int=0, history::Bool=false) where T <: AbstractFloat
 
@@ -54,8 +55,8 @@ assumed to be symmetric and positive definite.
 * C. C. Paige and M. A. Saunders, *Solution of Sparse Indefinite Systems of Linear Equations*, SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
 """
 function minres(A, b :: AbstractVector{T};
-                M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
-                rtol :: T=√eps(T)/100, etol :: T=√eps(T),
+                M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100, rtol :: T=√eps(T)/100, 
+                ratol :: T=√eps(T), rrtol :: T=√eps(T), etol :: T=√eps(T),
                 window :: Int=5, itmax :: Int=0, conlim :: T=1/√eps(T),
                 verbose :: Int=0, history :: Bool=false) where T <: AbstractFloat
 
@@ -89,7 +90,6 @@ function minres(A, b :: AbstractVector{T};
   δbar = zero(T)
   ϵ = zero(T)
   rNorm = β₁
-  rNorm0 = rNorm
   rNorms = history ? [β₁] : T[]
   ϕbar = β₁
   rhs1 = β₁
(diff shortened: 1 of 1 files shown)


src/minres.jl
```diff
@@ -233,7 +234,7 @@ function minres(A, b :: AbstractVector{T};
     tired = iter ≥ itmax
     ill_cond_lim = (one(T) / Acond ≤ ctol)
     solved_lim = (test2 ≤ tol)
-    zero_resid_lim = (test1 ≤ tol)
+    zero_resid_lim = (rNorm ≤ ϵA + ϵr * rNorm0)
```
amontoison, line 237:
```suggestion
    zero_resid_lim = isa(M, opEye) && (test1 ≤ tol)
```

amontoison: The backward error is only correctly mesured when `M = I`. We can keep the original test `(test1 ≤ tol)` in that case.

dpo: I requested different changes.

geoffroyleconte: I have made the changes but I don't know about the default values

src/minres.jl
```diff
@@ -233,7 +235,7 @@ function minres(A, b :: AbstractVector{T};
     tired = iter ≥ itmax
     ill_cond_lim = (one(T) / Acond ≤ ctol)
     solved_lim = (test2 ≤ tol)
-    zero_resid_lim = (test1 ≤ tol)
+    zero_resid_lim = (test1 ≤ tol) || (rNorm ≤ rNormtol)
```
dpo, line 238: It should be a new test in two parts:
```julia
resid_decrease_lim = rNorm ≤ rNormtol
...
resid_decrease_mach = rNorm + one(T) ≤ one(T)
```

src/minres.jl
```diff
@@ -25,7 +25,8 @@ export minres
 """
     (x, stats) = minres(A, b::AbstractVector{T};
                         M=opEye(), λ::T=zero(T), atol::T=√eps(T)/100,
-                        rtol::T=√eps(T)/100, etol::T=√eps(T),
+                        rtol::T=√eps(T)/100, ratol :: T=√eps(T), 
+                        rrtol :: T=√eps(T), etol::T=√eps(T),
```
dpo, line 29: Set them to zero by default so as not to change the current behavior.

dpo (changes requested):
Thanks. Almost there.

src/minres.jl
geoffroyleconte: I don't understand the use of `resid_decrease_mach`

dpo: It’s there as a fallback in case the user sets tolerances that are too small and rNorm becomes so small that in floating-point arithmetic, fl(rNorm + 1) == 1. It will be the case as soon as rNorm ≤ eps(T). There’s a “mach” test for the other stopping conditions too.

dpo:
Great, thanks. Does this work for you?

dpo (approved):

dpo:
Hmmm 🤔 
```
signal (4): Illegal instruction: 4
366
in expression starting at /Users/runner/work/Krylov.jl/Krylov.jl/test/test_lnlq.jl:101
367
dlapy2_64_ at /Users/runner/hostedtoolcache/julia/nightly/x64/lib/julia/libopenblas64_.0.3.13.dylib (unknown line)
```
https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/279/checks?check_run_id=2281601177#step:6:365

geoffroyleconte:
Do we care about nightly?

dpo:
The failures are not due to this PR: #282

amontoison:
> Hmmm thinking
> 
> ```
> signal (4): Illegal instruction: 4
> in expression starting at /Users/runner/work/Krylov.jl/Krylov.jl/test/test_lnlq.jl:101
> dlapy2_64_ at /Users/runner/hostedtoolcache/julia/nightly/x64/lib/julia/libopenblas64_.0.3.13.dylib (unknown line)
> ```
> 
> https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/279/checks?check_run_id=2281601177#step:6:365

They are some issues with the last version of OpenBLAS compiled with `Yggdrasil` but It only concerns Julia 1.7.

dpo:
Thank you @geoffroyleconte !

PR #286, opened 2021-04-21
Update CI, TagBot and documentation workflows

JSOBot:


amontoison:
The docstrings for Krylov methods and the other functions are not in the same section.
This pull request is not relevant for this repository.

dpo:
The idea is to make the docs automatic and uniform. I would rather not go and modify the docs by hand each time. The references are supposed to be just that: a reference.

dpo:
We could set `strict=false` in `make.jl` so Documenter doesn’t complain that docstrings are included multiple times. Would that be better?

amontoison:
Yes, I like this idea. Let's get back `api.md` and set `strict=false` in the `make.jl` file.

PR #287, opened 2021-04-23
Add testsets

amontoison:
@dpo 
I removed all outputs in the tests and each method is tested within a `testset`.

### DIFF
--- a/test/check_min_norm.jl
+++ b/test/check_min_norm.jl
@@ -10,6 +10,5 @@ function check_min_norm(A, b, x; λ=0.0)
   QR = qr(AI')
   xmin = QR.Q * (QR.R' \ b)
   xmin_norm = norm(xmin)
-  @printf("‖x - xmin‖ = %7.1e\n", norm(xI - xmin))
   return (xI, xmin, xmin_norm)
 end
--- a/test/test_aux.jl
+++ b/test/test_aux.jl
@@ -97,4 +97,6 @@ function test_aux()
   Krylov.kones(Vector{Float64}, 10) == ones(10)
 end
 
-test_aux()
+@testset "aux" begin
+  test_aux()
+end
--- a/test/test_aux.jl
+++ b/test/test_aux.jl
@@ -1,4 +1,4 @@
-function test_aux()
+@testset "aux" begin
   # test Givens reflector corner cases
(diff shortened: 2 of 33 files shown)


test/check_min_norm.jl
```diff
@@ -10,6 +10,5 @@ function check_min_norm(A, b, x; λ=0.0)
   QR = qr(AI')
   xmin = QR.Q * (QR.R' \ b)
   xmin_norm = norm(xmin)
-  @printf("‖x - xmin‖ = %7.1e\n", norm(xI - xmin))
```
dpo, line 13: We should test that they’re close.

test/test_aux.jl
```diff
@@ -97,4 +97,6 @@ function test_aux()
   Krylov.kones(Vector{Float64}, 10) == ones(10)
 end
 
-test_aux()
+@testset "aux" begin
+  test_aux()
```
dpo, line 101: Do we still need a function here? Why not paste the contents of `test_aux()` inside the testset?

amontoison: For the allocation tests we need a function but in that case it's not required. The advantage of using functions is that all variables are in a local scope.

test/check_min_norm.jl
amontoison: We already test that they're close : 
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/test/test_lnlq.jl#L32
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/test/test_craig.jl#L23
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/master/test/test_cgne.jl#L23

I only removed the `@printf` message.

test/test_aux.jl
dpo: Same with testsets I think since they are between `begin .. end`.

amontoison: I updated the PR, I only kept a function for allocation tests. The type of `MinresSolver` is not correcly infered within a `begin ... end`.

dpo:
OMG! So much better!!! Thanks!

I think we lost a little coverage because we’re no longer exercising the `show` functions. Abel has a way of testing those with `IOBuffer`s. I’ll open an issue for later.

PR #288, opened 2021-04-23
Logging

amontoison:
@dpo 
superseded #85


dpo:
Here it would be useful to see benchmarks with logging activated and deactivated.

src/bicgstab.jl
```diff
@@ -48,7 +48,7 @@ function bicgstab(A, b :: AbstractVector{T}; c :: AbstractVector{T}=b,
   n, m = size(A)
   m == n || error("System must be square")
   length(b) == m || error("Inconsistent problem size")
-  (verbose > 0) && @printf("BICGSTAB: system of size %d\n", n)
+  (verbose > 0) && @info @sprintf("BICGSTAB: system of size %d\n", n)
```
dpo, line 51: No need for newlines if using logging.

dpo: Also we no longer need the verbose flag since there’s a mechanism to turn off logging.

amontoison: Good catch for newlines :+1: 
The verbose flag is used to avoid the computation of `@sprintf` that is called even if logging is turned off. It's a solution to avoid allocations.

amontoison:
> Here it would be useful to see benchmarks with logging activated and deactivated.

Do you want to benchmark methods with `verbose=0` and `verbose=1` ?

amontoison:
The current `verbose` mode with an integer parameter works fine and is efficient.
We should keep it like this to avoid performance loss.

PR #293, opened 2021-05-12
Use mul! function

amontoison:
@geoffroyleconte, @dpo
What do you think about this trick ? It could solve our issue with `mul!`.
`kmul!` is `*` for LinearOperators and `mul!` for matrices and other linear operators (LinearMaps, etc...).
 
Some vectors must not be allocated by default, like the ones needed by the preconditioners or the other ones used by `LS` and `LN` methods for the regularization. Even if we update `LinearOperators.jl`, we will need that :
```
!isa(M, OpEye) && isempty(solver.v) && solver.v = kzeros(S, n)
sqd && isempty(solver.v) && solver.v = kzeros(S, n)
```
 
The idea here is that we can do the same thing for the vectors needed for `A * v` and `A' * u`.

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -105,16 +105,16 @@ function bicgstab!(solver :: BicgstabSolver{T,S}, A, b :: AbstractVector{T}; c :
     iter = iter + 1
     ρ = next_ρ
 
-    y = N * p                            # yₖ = N⁻¹pₖ
-    q = A * y                            # qₖ = Ayₖ
-    Mq = M * q; @kcopy!(n, Mq, v)        # vₖ = M⁻¹qₖ
+    mul!(y, N, p)                        # yₖ = N⁻¹pₖ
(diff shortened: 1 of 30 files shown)


dpo:
We won’t need to do this with @geoffroyleconte ’s revamp of LinearOperators, but I think it’s a good idea to preallocate vectors for the result of mul!() inside the solver structure.

geoffroyleconte:
Yes I agree with @dpo , if you want you can test the revamp here: https://github.com/geoffroyleconte/LinearOperatorsRevamp.jl (There way be issues with special operators and `opDiagonal` is not implemented yet). If I am correct the only thing that will have to change is that I am using preallocated Vectors inside the `LinearOperator` struct and you can reuse them in the KrylovSolvers (otherwise if you work with matrices instead of `LinearOperator` you just have to create new Vectors).

amontoison:
Ok, I will only use `mul!` and remove `src/variants.jl` such that we can test the revamp of `LinearOperators.jl`.

PR #295, opened 2021-05-13
Don't use ~1.2 in compat

abelsiqueira:
`~1.2` means `[1.2.0,1.3.0)`.
`1.2` means `[1.2.0,2.0.0)`.
Using `^` is the same as not using anything.
Ref.: https://pkgdocs.julialang.org/v1/compatibility/


amontoison (approved):

dpo:
Thanks

PR #296, opened 2021-05-13
Update the documentation of Krylov.jl

amontoison:
@dpo 
It's not the best division because some solvers have multiple applications but It can help the user to find an appropriate solver for its problem.


src/minres.jl
```diff
@@ -52,14 +52,14 @@ assumed to be symmetric and positive definite.
...
+    (x, stats) = minres!(solver :: MinresSolver{T,S}, A, b :: AbstractVector{T};
                          M=opEye(), λ :: T=zero(T), atol :: T=√eps(T)/100,
                          rtol :: T=√eps(T)/100, ratol :: T=zero(T), 
                          rrtol :: T=zero(T), etol :: T=√eps(T),
                          itmax :: Int=0, conlim :: T=1/√eps(T),
-                         verbose :: Int=0) where {S, T <: AbstractFloat}
+                         verbose :: Int=0) where {T <: AbstractFloat, S <: DenseVector{T}}
 
-where `solver` is a [`Krylov.MinresSolver`](@ref) used to store the vectors used by `minres!`.
+where `solver` is a `Krylov.MinresSolver` used to store the vectors used by `minres!`.
```
geoffroyleconte, line 62: You don't want to keep the link here?

dpo:
I don’t understand why we get these failures:
```
cr: Test Failed at /home/runner/work/Krylov.jl/Krylov.jl/test/test_cr.jl:38
  Expression: resid ≤ cr_tol
   Evaluated: 0.004739611328922038 ≤ 1.0e-6
 ```

dpo:
Otherwise the docs look good to me.

amontoison:
> I don’t understand why we get these failures:
> 
> ```
> cr: Test Failed at /home/runner/work/Krylov.jl/Krylov.jl/test/test_cr.jl:38
>   Expression: resid ≤ cr_tol
>    Evaluated: 0.004739611328922038 ≤ 1.0e-6
> ```

A random rhs is generated in `test/cr.jl` before the ramdom seed is fixed.
I solved this issue with #297

src/minres.jl
amontoison: It doesn't work anymore because Dominique allows duplicate documentation with #286 and Documenter.jl doesn't know which one it should use for the link.

After LA21, I will finish #293 and #288 and add a section about in-place methods.

PR #298, opened 2021-05-13
Fix issues with Julia 1.7

amontoison:


dpo:
Great. I was wondering about those failures. There’s still something related to LinearOperators though :-(

amontoison:
Let's wait the official release 1.7 to see if we still have the same issues.

amontoison:
> Great. I was wondering about those failures. There’s still something related to LinearOperators though :-(

The issue is related to the function `size` of `SparseMatrixCSC` that returns `Int32` now. I think that it's a bug because we still have `Int64` with Float32 dense matrices.

amontoison:
I found the culprit : https://github.com/JuliaLang/julia/pull/39645

amontoison:
`nrow` and `ncol` attributes of LinearOperators should be `Integer` instead of `Int`.
@geoffroyleconte Can you take this into account for the revamp ?

dpo (approved):

PR #300, opened 2021-05-19
Add a new constructor for all KrylovSolvers

### LINKED ISSUE #299: Constructors KrylovSolver
Thanks for the amazing work including the KrylovSolver.

I just find having `MinresSolver(A, b)` as a constructor not practical since only the size of A and the type of b is needed.
For instance, in an optimization solver, we typically want to initialize the KrylovSolver at the initialization step where the size and the type is known but we don't have access yet to the matrix and the rhs.

Maybe we could have a `MinresSolver(m, n, S)` by default and an additional one with `MinresSolver(A, b)`. What do you think?

amontoison:
fix #299 
@tmigot @dpo

### DIFF
--- a/src/krylov_solvers.jl
+++ b/src/krylov_solvers.jl
@@ -10,8 +10,9 @@ abstract type KrylovSolver{T,S} end
 """
 Type for storing the vectors required by the in-place version of MINRES.
 
-The outer constructor
+The outer constructors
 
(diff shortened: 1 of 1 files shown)


tmigot (approved):
I tested it with the code I am working on, and it's perfect, thanks !

PR #330, opened 2021-05-28
Remove spaces with special character

amontoison:


dpo:
Hm. What was happening here?

amontoison:
I updated Sublime Text and I remarked that some spaces used a special character.
With a sed command, I removed all the "special" spaces.

PR #334, opened 2021-05-29
Update Project.toml and remove dependencies of LinearOperators.jl

amontoison:
@geoffroyleconte @abelsiqueira @dpo

Vectors inside KrylovSolvers used for the preconditioners or the regularization are `Union{S, Nothing}`.
Sometimes 16 extra bits are required for type inference.
To help Julia I added some `::S`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,12 +4,10 @@ version = "0.7.2"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
-LinearOperators = "5c8ed15e-5a4c-59e4-a42b-c7e8811fb125"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
-LinearOperators = "1.2"
 julia = "^1.3.0"
 
 [extras]
--- a/src/Krylov.jl
+++ b/src/Krylov.jl
@@ -1,6 +1,6 @@
 module Krylov
 
-using LinearOperators, LinearAlgebra, SparseArrays, Printf
+using LinearAlgebra, SparseArrays, Printf
 
 include("krylov_stats.jl")
 include("krylov_utils.jl")
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -17,7 +17,7 @@ export bicgstab, bicgstab!
 
 """
     (x, stats) = bicgstab(A, b::AbstractVector{T}; c::AbstractVector{T}=b,
-                          M=opEye(), N=opEye(), atol::T=√eps(T), rtol::T=√eps(T),
+                          M=I, N=I, atol::T=√eps(T), rtol::T=√eps(T),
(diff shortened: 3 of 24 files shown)


geoffroyleconte:
Maybe you should try to replace in the solvers `Union{nothing, S}` by `S` and allocate an empty vector when you don't need it? I'm not sure that adding `Union` in types is a good idea because there is no specialization but I may be mistaken

amontoison:
> Maybe you should try to replace in the solvers `Union{nothing, S}` by `S` and allocate an empty vector when you don't need it? I'm not sure that adding `Union` in types is a good idea because there is no specialization but I may be mistaken

You're right Geoffroy. I used `nothing` because I wanted to avoid the 80 bits required for `S[]` but it will avoid type inference issues.

geoffroyleconte:
This may also work
```
mutable struct CgSolver{T,S,S2} <: KrylovSolver{T,S}
  x  :: S
  r  :: S
  p  :: S
  Ap :: S
  z  :: S2
end

CgSolver{T,S}(x::S, r::S, p::S, Ap::S, z::S2) where {T,S,S2} = CgSolver{T,S,S2}(x, r, p, Ap, z)

function CgSolver{T,S}(n, m) where {T,S}
  x  = S(undef, n)
  r  = S(undef, n)
  p  = S(undef, n)
  Ap = S(undef, n)
  z  = nothing
  solver = CgSolver{T,S}(x, r, p, Ap, z)
  return solver
end

test = CgSolver{Float64, Vector{Float64}}(5,5)
```

amontoison:
> x::S, r::S, p::S, Ap::S, z::S2

Can you change the type of `z` in that case with same solver ?
`Nothing` -> `S` ?

amontoison:
We allocate 80 or 160 extra bits for some KrylovSolvers now but it solves our type inference issues.
What do you think of this solution ?

geoffroyleconte:
For your 1st question: no, but you could create another solver that reuses all the memory of the previous solver (you could create a function to do this).

But yes I think that it is not a problem to allocate a bit more in the solver if the in-place function has 0 allocations.

amontoison:
> For your 1st question: no, but you could create another solver that reuses all the memory of the previous solver (you could create a function to do this).

I don't like this solution because the user will not be able to have a pointer to the new solver if we create it inside Krylov methods with the previous solver.

> But yes I think that it is not a problem to allocate a bit more in the solver if the in-place function has 0 allocations.

Ok, let's keep this solution if Dominique and Abel like it too.

dpo:
Looks good to me but all the S stuff is unrelated to the subject of this PR. It would really be best to keep each PR focused.

amontoison:
> Looks good to me but all the S stuff is unrelated to the subject of this PR. It would really be best to keep each PR focused.

Ok, I will open a PR for the S stuff and I will rebase this one after.

abelsiqueira:
> 
> > But yes I think that it is not a problem to allocate a bit more in the solver if the in-place function has 0 allocations.
> 
> Ok, let's keep this solution if Dominique and Abel like it too.

I like it too.

tmigot:
Hey @amontoison ! Any idea why DCISolver.jl is breaking with this PR?

PR #205, opened 2020-08-02
Update the DOI

amontoison:
doi

PR #348, opened 2021-06-07
Fix .zenodo.json

amontoison:
Follows the recommendation of Abel (#205)

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -2,7 +2,6 @@
   "description": "A Julia basket of hand-picked Krylov methods",
   "title": "Krylov.jl",
   "upload_type": "software",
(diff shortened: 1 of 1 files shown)


dpo:
Other packages have an explicit date, e.g., "publication_date": "2019-02-07"

amontoison:
> Other packages have an explicit date, e.g., "publication_date": "2019-02-07"

I wanted to remove it like `NLPModels.jl` but it's also an issue: https://zenodo.org/record/3692025#.YL5hZjopA1g. 
The new releases are not added. Let's keep the original release date.

dpo:
Yes. I think it doesn't really matter. Or we could update it when we bump the version.

Issue #265, opened 2021-03-15
Estimates of the error in the LNLQ method


amontoison:

amontoison:
fixed by #350

PR #350, opened 2021-06-19
Estimate of the error in LNLQ

tmigot:
This would fix Issue #265 .

The modification is heavily inspired by the Matlab implementation: [https://github.com/restrin/LinearSystemSolvers/blob/master/LNLQ/lnlq.m](https://github.com/restrin/LinearSystemSolvers/blob/master/LNLQ/lnlq.m).

A couple of comments:
- We use the CRAIG-bound when `transfer_to_craig` is set as true.
- Due to numerical issues, the `sqrt` in the bound might not be defined. When this happens, there is a boolean `complex_error_bnd` set as `true` and we stop computing this bound.
- The extra allocations are due to the use of `LNLQStats` instead of `SimpleStats`.

### DIFF
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -74,18 +74,18 @@ mutable struct AdjointStats{T} <: KrylovStats{T}
 end
 
 """
-Type for statistics returned by adjoint systems solvers BiLQR and TriLQR, the attributes are:
+Type for statistics returned by the LNLQ method, the attributes are:
 - solved
-- inconsistent
 - residuals
+- error_with_bnd
 - error_bnd_x
 - error_bnd_y
 - status
 """
 mutable struct LNLQStats{T} <: KrylovStats{T}
   solved :: Bool
-  inconsistent :: Bool
   residuals :: Vector{T}
+  error_with_bnd :: Bool
   error_bnd_x :: Vector{T}
   error_bnd_y :: Vector{T}
   status :: String
@@ -100,7 +100,7 @@ special_fields = Dict(
   :Acond => "κ₂(A)",
 )
 
-for f in ["Simple", "Lanczos", "Symmlq", "Adjoint"]
+for f in ["Simple", "Lanczos", "Symmlq", "Adjoint", "LNLQ"]
   T = Meta.parse("Krylov." * f * "Stats{S}")
   @eval function show(io :: IO, stats :: $T) where S
     s  = $f * " stats\n"
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -73,6 +73,24 @@ mutable struct AdjointStats{T} <: KrylovStats{T}
   status :: String
 end
 
+"""
+Type for statistics returned by adjoint systems solvers BiLQR and TriLQR, the attributes are:
+- solved
+- inconsistent
+- residuals
+- error_bnd_x
+- error_bnd_y
+- status
+"""
+mutable struct LNLQStats{T} <: KrylovStats{T}
+  solved :: Bool
+  inconsistent :: Bool
+  residuals :: Vector{T}
+  error_bnd_x :: Vector{T}
+  error_bnd_y :: Vector{T}
+  status :: String
+end
+
 import Base.show
 
 special_fields = Dict(
--- a/src/lnlq.jl
+++ b/src/lnlq.jl
@@ -26,8 +26,8 @@ export lnlq, lnlq!
 
 """
     (x, y, stats) = lnlq(A, b::AbstractVector{T};
-                         M=I, N=I, sqd::Bool=false, λ::T=zero(T),
-                         atol::T=√eps(T), rtol::T=√eps(T), itmax::Int=0,
+                         M=I, N=I, sqd::Bool=false, λ::T=zero(T), σ::T=zero(T),
+                         atol::T=√eps(T), rtol::T=√eps(T), etolx::T=√eps(T), etoly::T=√eps(T), itmax::Int=0,
                          transfer_to_craig::Bool=true, verbose::Int=0, history::Bool=false) where T <: AbstractFloat
 
 Find the least-norm solution of the consistent linear system
@@ -61,8 +61,8 @@ In this case, M can still be specified and indicates the weighted norm in which 
 In this implementation, both the x and y-parts of the solution are returned.
 
 `etolx` and `etoly` are tolerances on the upper bound of the distance to the solution ‖x-xₛ‖ and ‖y-yₛ‖, respectively.
-The bound is valid if λ>0 or σₐ>0 where σₐ should be strictly smaller than the smallest positive singular value.
-For instance σₐ:=(1-1e-7)σₘᵢₙ .
+The bound is valid if λ>0 or σ>0 where σ should be strictly smaller than the smallest positive singular value.
+For instance σ:=(1-1e-7)σₘᵢₙ .
 
 #### Reference
 
@@ -74,7 +74,7 @@ function lnlq(A, b :: AbstractVector{T}; kwargs...) where T <: AbstractFloat
 end
 
 function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
-               M=I, N=I, sqd :: Bool=false, λ :: T=zero(T), σₐ :: T=zero(T),
+               M=I, N=I, sqd :: Bool=false, λ :: T=zero(T), σ :: T=zero(T),
                atol :: T=√eps(T), rtol :: T=√eps(T), etolx :: T=√eps(T), etoly :: T=√eps(T), itmax :: Int=0,
                transfer_to_craig :: Bool=true, verbose :: Int=0, history :: Bool=false) where {T <: AbstractFloat, S <: DenseVector{T}}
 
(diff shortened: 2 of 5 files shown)


src/krylov_stats.jl
```diff
@@ -73,6 +73,24 @@ mutable struct AdjointStats{T} <: KrylovStats{T}
   status :: String
 end
 
+"""
+Type for statistics returned by adjoint systems solvers BiLQR and TriLQR, the attributes are:
```
amontoison, line 77:
```suggestion
Type for statistics returned by the LNLQ method, the attributes are:
```

src/krylov_stats.jl
```diff
@@ -73,6 +73,24 @@ mutable struct AdjointStats{T} <: KrylovStats{T}
...
+- status
+"""
+mutable struct LNLQStats{T} <: KrylovStats{T}
+  solved :: Bool
+  inconsistent :: Bool
+  residuals :: Vector{T}
+  error_bnd_x :: Vector{T}
+  error_bnd_y :: Vector{T}
+  status :: String
+end
```
amontoison, line 92: You also need to add a `show` method.

src/lnlq.jl
```diff
@@ -70,8 +74,8 @@ function lnlq(A, b :: AbstractVector{T}; kwargs...) where T <: AbstractFloat
 end
 
 function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
-               M=I, N=I, sqd :: Bool=false, λ :: T=zero(T),
-               atol :: T=√eps(T), rtol :: T=√eps(T), itmax :: Int=0,
+               M=I, N=I, sqd :: Bool=false, λ :: T=zero(T), σₐ :: T=zero(T),
+               atol :: T=√eps(T), rtol :: T=√eps(T), etolx :: T=√eps(T), etoly :: T=√eps(T), itmax :: Int=0,
```
amontoison, line 78: Can you also update the docstring ?

src/lnlq.jl
```diff
@@ -191,6 +199,22 @@ function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
   tired = false
   status = "unknown"
 
+  if σₑₛₜ > 0
+    τtildeₖ = βₖ / σₑₛₜ
+    ζtildeₖ = τtildeₖ / σₑₛₜ
+    err_x = τtildeₖ
+    err_y = ζtildeₖ
+
+    solved_lq = err_x <= etolx || err_y <= etoly
```
amontoison, line 208:
```suggestion
    solved_lq = err_x ≤ etolx || err_y ≤ etoly
```

src/lnlq.jl
```diff
@@ -191,6 +199,22 @@ function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
...
+    history && push!(xNorms, err_x)
+    history && push!(yNorms, err_y)
+
+    complex_error_bnd = false
+
+    ρbar = -σₑₛₜ
+    csig = -1
```
amontoison, line 215:
```suggestion
    csig = -one(T)
```

src/lnlq.jl
```diff
@@ -327,6 +408,13 @@ function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
     tired = iter ≥ itmax
     solved_lq = rNorm_lq ≤ ε
     solved_cg = transfer_to_craig && rNorm_cg ≤ ε
+    if σₑₛₜ > 0
+      if transfer_to_craig
+        solved_cg = solved_cg || err_x <= etolx || err_y <= etoly
```
amontoison, line 413:
```suggestion
        solved_cg = solved_cg || err_x ≤ etolx || err_y ≤ etoly
```

src/lnlq.jl
```diff
@@ -327,6 +408,13 @@ function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
...
+      else
+        solved_lq = solved_lq || err_x <= etolx || err_y <= etoly
```
amontoison, line 415:
```suggestion
        solved_lq = solved_lq || err_x ≤ etolx || err_y ≤ etoly
```

src/lnlq.jl
```diff
@@ -292,6 +338,41 @@ function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
...
     @kaxpby!(m, -cₖ₊₁, u, sₖ₊₁, w̄)
 
+    if σₑₛₜ > 0 && !complex_error_bnd
+      if transfer_to_craig
+        disc_x = τtildeₖ^2 - τₖ₊₁^2
+        if disc_x < 0
+          complex_error_bnd = true
+        else
+          err_x = √disc_x
+        end
```
amontoison, line 348:
```suggestion
disc_x < 0 ? complex_error_bnd = true : err_x = √disc_x
```

You can do the same modification for `disc_y`.

amontoison (commented):
Thanks Tangi !
Is it possible to have a graphic with the error norms and the error bounds ?

tmigot:
> Thanks Tangi !
> Is it possible to have a graphic with the error norms and the error bounds ?

Thanks for the comments @amontoison 
I compiled figures on some tests problems here [in a test repo](https://github.com/tmigot/DCI-tests/tree/master/test-lnlq-Krylov) showing both CRAIG and LNLQ bounds.

src/krylov_stats.jl
tmigot: I added it. Is there a trick to increase coverage of this?

amontoison: With #287, I removed all `show` calls. We should add a testset.

src/lnlq.jl
```diff
@@ -26,8 +26,8 @@ export lnlq, lnlq!
 
 """
     (x, y, stats) = lnlq(A, b::AbstractVector{T};
-                         M=I, N=I, sqd::Bool=false, λ::T=zero(T),
-                         atol::T=√eps(T), rtol::T=√eps(T), itmax::Int=0,
+                         M=I, N=I, sqd::Bool=false, λ::T=zero(T), σₐ :: T=zero(T),
+                         atol::T=√eps(T), rtol::T=√eps(T), etolx :: T=√eps(T), etoly :: T=√eps(T), itmax::Int=0,
```
amontoison, line 30:
```suggestion
                         M=I, N=I, sqd::Bool=false, λ::T=zero(T), σₐ::T=zero(T),
                         atol::T=√eps(T), rtol::T=√eps(T), etolx::T=√eps(T), etoly::T=√eps(T), itmax::Int=0,
```

amontoison: We remove spaces for the docstrings.

amontoison:
I will merge this PR after #354 if it's fine for you Tangi.

src/lnlq.jl
```diff
@@ -26,8 +26,8 @@ export lnlq, lnlq!
 
 """
     (x, y, stats) = lnlq(A, b::AbstractVector{T};
-                         M=I, N=I, sqd::Bool=false, λ::T=zero(T),
-                         atol::T=√eps(T), rtol::T=√eps(T), itmax::Int=0,
+                         M=I, N=I, sqd::Bool=false, λ::T=zero(T), σₐ :: T=zero(T),
```
dpo, line 29: This parameter name is kind of hard to type. Why not just σ ?

src/lnlq.jl
```diff
@@ -267,6 +291,28 @@ function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
...
+      μbar = -csig * αhatₖ
+      ρ = √(ρbar^2 + αhatₖ^2)
+      csig = ρbar / ρ
+      ssig = αhatₖ / ρ
+      ρbar = ssig * μbar + csig * σₑₛₜ
+      μbar = -csig * βhatₖ₊₁
+      θ = βhatₖ₊₁ * csig / ρbar
+      ωdisc = σₑₛₜ^2 - σₑₛₜ * βhatₖ₊₁ * θ
+      if ωdisc < 0
+        complex_error_bnd = true
```
dpo, line 304: What is supposed to happen here? This should just be an error because σ is too large or cancellation.

src/lnlq.jl
```diff
@@ -363,6 +439,6 @@ function lnlq!(solver :: LnlqSolver{T,S}, A, b :: AbstractVector{T};
   tired     && (status = "maximum number of iterations exceeded")
   solved_lq && (status = "solutions (xᴸ, yᴸ) good enough for the tolerances given")
   solved_cg && (status = "solutions (xᶜ, yᶜ) good enough for the tolerances given")
-  stats = SimpleStats(solved_lq || solved_cg, false, rNorms, T[], status)
+  stats = LNLQStats(solved_lq || solved_cg, false, rNorms, xNorms, yNorms, status)
```
dpo, line 442: complex_error_bnd isn’t used... ?!

tmigot: I added it in the entries of `LNLQStats`.

src/lnlq.jl
tmigot: Thank you for the comments.

I use `complex_error_bnd` to stop computing the error bound once there is an error. Now it makes more sense because it is returned in the `LNLQStats`.

I thought it is better than throwing an error, as it doesn't perturb the computation of the iterates.

dpo: Ok that makes sense, thank you. As long as you report it in the final stats, it's a good strategy.

dpo (approved):

src/krylov_stats.jl
```diff
@@ -73,6 +73,24 @@ mutable struct AdjointStats{T} <: KrylovStats{T}
   status :: String
 end
 
+"""
+Type for statistics returned by the LNLQ method, the attributes are:
+- solved
+- inconsistent
+- residuals
```
amontoison, line 80:
```suggestion
- residuals
- error_with_bnd
```

amontoison:
Thanks Tangi!

PR #352, opened 2021-06-21
Use 5-argument mul! in USYMQR

amontoison:
@dpo, @geoffroyleconte 
I updated USYMQR to use 5-argument mul!. What do you think of `mul5` for the argument name ?
If it's fine for you, I will update the other Krylov methods.

### DIFF
--- a/src/usymqr.jl
+++ b/src/usymqr.jl
@@ -21,7 +21,7 @@ export usymqr, usymqr!
 
 """
     (x, stats) = usymqr(A, b::AbstractVector{T}, c::AbstractVector{T};
-                        atol::T=√eps(T), rtol::T=√eps(T),
+                        atol::T=√eps(T), rtol::T=√eps(T), mul5::Bool=true,
                         itmax::Int=0, verbose::Int=0, history::Bool=false) where T <: AbstractFloat
 
 Solve the linear system Ax = b using the USYMQR method.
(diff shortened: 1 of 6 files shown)


src/usymqr.jl
```diff
@@ -105,19 +107,32 @@ function usymqr!(solver :: UsymqrSolver{T,S}, A, b :: AbstractVector{T}, c :: Ab
...
     # AᵀVₖ = Uₖ(Tₖ)ᵀ + γₖ₊₁uₖ₊₁(eₖ)ᵀ = Uₖ₊₁(Tₖ.ₖ₊₁)ᵀ
 
-    mul!(q, A , uₖ)  # Forms vₖ₊₁ : q ← Auₖ
-    mul!(p, Aᵀ, vₖ)  # Forms uₖ₊₁ : p ← Aᵀvₖ
+    if mul5
+      mul!(vₖ₋₁, A , uₖ, one(T), -γₖ)  # Forms vₖ₊₁ : vₖ₋₁ ← Auₖ  - γₖvₖ₋₁
+      mul!(uₖ₋₁, Aᵀ, vₖ, one(T), -βₖ)  # Forms uₖ₊₁ : uₖ₋₁ ← Aᵀvₖ - βₖuₖ₋₁
+    else
+      mul!(q, A , uₖ)  # q ← Auₖ
+      mul!(p, Aᵀ, vₖ)  # p ← Aᵀvₖ
```
dpo, line 115: Yuck. I hope we don't really need to do this. It's becoming unreadable. Can't we assume that it's always a 5-arg mul?

amontoison: We can assume that it's always a 5-arg mul! but if the user use automatic differentiation for example and a operator to model `A * v` and `A' * u` he cannot use it without some extra modifications.

amontoison: @dpo, I defined a new macro `@kaAxpby!`. Is it more readable ?

PR #354, opened 2021-06-24
Generic show functions for stats

tmigot:
This way we no longer have to redefine everything for the `show` functions of the Stats.


amontoison:
Thanks Tangi !

Can you add some tests to improve coverage and verify that the new `show` function is working ?
We could do something like this: https://github.com/JuliaSmoothOptimizers/NLPModels.jl/blob/master/test/nlp/show.jl

dpo:
Nice!

PR #359, opened 2021-06-29
Add a boolean for error in bound computation for LSLQ

tmigot:
Add a boolean when there is a `DomainError` in the computation of the upper bound instead of having an error.

### DIFF
--- a/test/test_lslq.jl
+++ b/test/test_lslq.jl
@@ -28,6 +28,10 @@
   @test x == zeros(size(A,1))
   @test stats.status == "x = 0 is a zero-residual solution"
 
+  # Test error in upper bound computation
+  (x, stats) = lslq(A, b, σ=1.)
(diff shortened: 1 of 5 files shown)


amontoison (approved):

test/test_lslq.jl
```diff
@@ -28,6 +28,12 @@
   @test x == zeros(size(A,1))
   @test stats.status == "x = 0 is a zero-residual solution"
 
+  # Test error in upper bound computation
+  A = rand(5, 5)
+  b = A * ones(5)
```
amontoison, line 33: Is it possible to have a test without a random matrix?

tmigot: I just reused the previous `A` and `b`, thanks.

amontoison:
Thanks !

PR #365, opened 2021-07-26
Add generic solve! function

amontoison:
@abelsiqueira

### DIFF
--- a/test/test_inplace.jl
+++ b/test/test_inplace.jl
@@ -47,152 +47,92 @@
     b  = 5 * b
     c  = 3 * c
 
-    x, stats = cg!(cg_solver, A, b)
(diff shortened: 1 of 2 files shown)


test/test_inplace.jl
```diff
@@ -49,91 +49,151 @@
 
     x, stats = cg!(cg_solver, A, b)
     @test stats.solved
+    x, stats = solve!(cg_solver, A, b)
+    @test stats.solved
```
dpo, line 53: Not necessary to call both solve! and cg!

dpo:
thanks

abelsiqueira:
Can we have a new release? 0.7.4 or 0.8.0?

Issue #361, opened 2021-06-30
Add KrylovStats in KrylovSolver


tmigot:
To continue pre-allocating things, it could be interesting to also have the `KrylovStats` in the `KrylovSolver`.

### Krylov methods

- [x] CG
- [x] CR
- [x] SYMMLQ
- [x] CG-LANCZOS
- [x] CG-LANCZOS with shifts
- [x] MINRES
- [x] MINRES-QLP
- [x] DQGMRES
- [x] DIOM
- [x] USYMLQ
- [x] USYMQR
- [x] TRICG
- [x] TRIMR
- [x] TRILQR
- [x] CGS
- [x] BICGSTAB
- [x] BILQ
- [x] QMR
- [x] BILQR
- [x] CGLS
- [x] CRLS
- [x] CGNE
- [x] CRMR
- [x] LSLQ
- [x] LSQR
- [x] LSMR
- [x] LNLQ
- [x] CRAIG
- [x] CRAIGMR

amontoison:
Thank you for the help Tangi!

tmigot:
Great job, thanks!

PR #377, opened 2021-08-06
Add in-place Stats in `cg`

tmigot:
@amontoison toward 0-allocation cg :).
Let me know if you have any comments and I will keep going with the others, #361 .

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -64,7 +64,9 @@ function cg!(solver :: CgSolver{T,S}, A, b :: AbstractVector{T};
 
   # Set up workspace.
   allocate_if(!MisI, solver, :z, S, n)
-  x, r, p, Ap = solver.x, solver.r, solver.p, solver.Ap
+  x, r, p, Ap, stats = solver.x, solver.r, solver.p, solver.Ap, solver.stats
+  rNorms = solver.stats.residuals
+  !history && !isempty(rNorms) && (rNorms = T[])
   z = MisI ? r : solver.z
 
   x .= zero(T)
@@ -72,7 +74,11 @@ function cg!(solver :: CgSolver{T,S}, A, b :: AbstractVector{T};
   MisI || mul!(z, M, r)
   p .= z
   γ = @kdot(n, r, z)
-  γ == 0 && return x, SimpleStats(true, false, [zero(T)], T[], "x = 0 is a zero-residual solution")
+  if γ == 0 
+    stats.solved, stats.inconsistent = true, false
+    stats.status = "x = 0 is a zero-residual solution"
+    return x, stats
+  end
 
   iter = 0
   itmax == 0 && (itmax = 2 * n)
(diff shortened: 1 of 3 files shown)


src/cg.jl
```diff
@@ -64,23 +64,28 @@ function cg!(solver :: CgSolver{T,S}, A, b :: AbstractVector{T};
...
   x .= zero(T)
   r .= b
   MisI || mul!(z, M, r)
   p .= z
   γ = @kdot(n, r, z)
-  γ == 0 && return x, SimpleStats(true, false, [zero(T)], T[], "x = 0 is a zero-residual solution")
+  if γ == 0 
+    stats.solved, stats.inconsistent = true, false
+    stats.status = "x = 0 is a zero-residual solution"
+    return x, stats
```
dpo, line 79: It's enough to return the solver in all cases, isn't it?

src/krylov_solvers.jl
```diff
@@ -76,7 +77,8 @@ mutable struct CgSolver{T,S} <: KrylovSolver{T,S}
     p  = S(undef, n)
     Ap = S(undef, n)
     z  = S(undef, 0)
-    solver = new{T,S}(x, r, p, Ap, z)
+    stats = SimpleStats(false, true, T[], T[], "unknown")
+    solver = new{T,S}(x, r, p, Ap, z, stats)
```
dpo, line 81: Maybe solvers could use a few getters: `solution(solver) = solver.x`, `stats(solver) = solver.stats`, ...

src/cg.jl
tmigot: Sounds like a good idea. Maybe `cg!` returns the solver, but `cg` returns `x, stats` as it is now?

dpo: They should be consistent and always return the solver in my opinion.

amontoison: We can only return the solver after #361 but we need to modify all solvers in the same time.

src/krylov_solvers.jl
amontoison: Good idea ! I propose to do it after #361 :smiley:

src/cg.jl
```diff
@@ -64,23 +64,28 @@ function cg!(solver :: CgSolver{T,S}, A, b :: AbstractVector{T};
 
   # Set up workspace.
   allocate_if(!MisI, solver, :z, S, n)
-  x, r, p, Ap = solver.x, solver.r, solver.p, solver.Ap
+  x, r, p, Ap, stats = solver.x, solver.r, solver.p, solver.Ap, solver.stats
+  rNorms = solver.stats.residuals
```
amontoison, line 68:
```suggestion
  rNorms = solver.stats.residuals
  !history && !isempty(rNorms) && (rNorms = T[])
```

src/cg.jl
```diff
@@ -64,23 +64,29 @@ function cg!(solver :: CgSolver{T,S}, A, b :: AbstractVector{T};
...
+  !history && !isempty(rNorms) && (rNorms .= T[])
```
amontoison, line 69:
```suggestion
  !history && !isempty(rNorms) && (rNorms = T[])
```

tmigot: Thanks

amontoison:
Thanks Tangi !

PR #378, opened 2021-08-06
Add in-place stats for `cr`

tmigot:


src/krylov_solvers.jl
```diff
@@ -115,7 +116,8 @@ mutable struct CrSolver{T,S} <: KrylovSolver{T,S}
     q  = S(undef, n)
     Ar = S(undef, n)
     Mq = S(undef, 0)
-    solver = new{T,S}(x, r, p, q, Ar, Mq)
+    stats = SimpleStats(false, true, T[], T[], "unknown")
```
amontoison, line 119:
```suggestion
    stats = SimpleStats(false, false, T[], T[], "unknown")
```

amontoison: We should do the same modification for MINRES.

src/cr.jl
```diff
@@ -56,7 +56,10 @@ function cr!(solver :: CrSolver{T,S}, A, b :: AbstractVector{T};
 
   # Set up workspace
   allocate_if(!MisI, solver, :Mq, S, n)
-  x, r, p, q, Ar = solver.x, solver.r, solver.p, solver.q, solver.Ar
+  x, r, p, q, Ar, stats = solver.x, solver.r, solver.p, solver.q, solver.Ar, solver.stats
+  rNorms, ArNorms = solver.stats.residuals, solver.stats.Aresiduals
```
amontoison, line 60:
```suggestion
  rNorms, ArNorms = stats.residuals, stats.Aresiduals
```

amontoison:
Thanks !

PR #379, opened 2021-08-07
Add reset! function for KrylovStats

amontoison:
@tmigot

### DIFF
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -17,11 +17,6 @@ mutable struct SimpleStats{T} <: KrylovStats{T}
   status       :: String
 end
 
-function reset!(stats :: SimpleStats)
-  empty!(stats.residuals)
-  empty!(stats.Aresiduals)
-end
-
 """
 Type for statistics returned by CG-LANCZOS, the attributes are:
 - solved
@@ -38,11 +33,6 @@ mutable struct LanczosStats{T} <: KrylovStats{T}
   Anorm     :: T
   Acond     :: T
   status    :: String
-end
-
-function reset!(stats :: LanczosStats)
-  empty!(stats.residuals)
-  isa(stats.flagged, AbstractVector) && empty!(stats.flagged)
 end
 
 """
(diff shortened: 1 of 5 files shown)


src/krylov_stats.jl
```diff
@@ -10,11 +10,16 @@ Type for statistics returned by the majority of Krylov solvers, the attributes a
...
   inconsistent :: Bool
-  residuals :: Vector{T}
-  Aresiduals :: Vector{T}
-  status :: String
+  residuals    :: Vector{T}
+  Aresiduals   :: Vector{T}
+  status       :: String
+end
+
+function reset!(stats :: SimpleStats)
```
tmigot, line 20: I lke the function `reset!` overall. What about we try to make something generic like the `show` function?
For instance [not tested]:
```
for f in ["Simple", "Lanczos", "Symmlq", "Adjoint", "LNLQ", "LSLQ"]
  T = Meta.parse("Krylov." * f * "Stats{S}")
  @eval function reset!(stats :: $T) where S
    for field in fieldnames($T)
      statfield = getfield(stats, field)
      if isa(statfield, AbstractVector)
        empty!(statfield)
      end
    end
  end
end
```

test/test_stats.jl
```diff
@@ -1,22 +1,23 @@
...
-  storage_type = typeof(solver)
+  storage_type = typeof(stats)
   expected = """Simple stats
   solved: true
   inconsistent: true
   residuals: []
   Aresiduals: []
   status: t"""
   @test strip.(split(chomp(showed), "\n")) == strip.(split(chomp(expected), "\n"))
+  Krylov.reset!(stats)
```
tmigot, line 14: This seems ok for the coverage. We should probably test it after a solver call, and check whether it worked?

Same for the others `Krylov.reset!(stats)` below.

src/krylov_stats.jl
amontoison: The issue is the call to `fieldnames($T)` that allocates a vector of symbols.
It's for that reason that I did a `reset!` function for each KrylovStats.

tmigot: Oh... I checked and we might use `for i=1:5 fieldname(SimpleStats{Float64}, i) end` maybe.

amontoison:
You can rebase and merge @tmigot if you want to work on #361 tomorrow.

src/krylov_stats.jl
tmigot: Last comment :). Wouldn't it be better to have the functions `reset!` and `show` in the same loop `for f in ["Simple", "Lanczos", "Symmlq", "Adjoint", "LNLQ", "LSLQ"]` ?

tmigot (approved):

PR #393, opened 2021-08-10
Add in-place stats for `lsmr`

tmigot:
#361 

This one is not completely with 0 allocations as we will need a LsmrSolver I think for the `err_vec`. I suggest we fix this in a separate PR.


amontoison:
> #361
> 
> This one is not completely with 0 allocations as we will need a LsmrSolver I think for the `err_vec`. I suggest we fix this in a separate PR.

Ok, let's stay focus on #361.

amontoison (approved):

PR #395, opened 2021-08-10
Add in-place stats for `lsqr`

tmigot:
#361 

Same as #393 , it will need a LsqrStats to have 0-allocation.

### DIFF
--- a/src/krylov_solvers.jl
+++ b/src/krylov_solvers.jl
@@ -1068,14 +1068,15 @@ The outer constructors
 may be used in order to create these vectors.
 """
 mutable struct LsqrSolver{T,S} <: KrylovSolver{T,S}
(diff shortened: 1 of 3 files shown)


amontoison (approved):

amontoison:
> #361
> 
> Same as #393 , it will need a LsqrStats to have 0-allocation.

Can't we just add `err_vec` in `LsqrSolver` like MINRES ?

tmigot:
> > #361
> > Same as #393 , it will need a LsqrStats to have 0-allocation.
> 
> Can't we just add `err_vec` in `LsqrSolver` like MINRES ?

Sure, we can. Same for `lsmr` then.
I suggest we do this in a separate PR with `lsmr`.

PR #403, opened 2021-08-11
Add a C interface for Krylov.jl with shared libraries

amontoison:
@dpo 
Since the JuliaCon  2021, `PackageCompiler.jl` is able to generate shared libraries for Julia packages.
I tested it with three Krylov methods (`CG`, `LSMR` and `CRAIG`).
It's a basic interface but it works fine and could be a nice feature if we do an article about Krylov.jl before the end of the thesis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ benchmark/tune.json
 benchmark/commit.md
 benchmark/main.md
 benchmark/judgement.md
+**/main
+**/*.o
+app/CKrylov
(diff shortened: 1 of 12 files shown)


dpo:
That's really interesting. It looks like a significant amount of new code...

The inevitable question is going to be: how does the performance compare to that of a pure C implementation?

amontoison:
> That's really interesting. It looks like a significant amount of new code...
> The inevitable question is going to be: how does the performance compare to that of a pure C implementation?

If it's well precompiled, It should be competitive with a pure C implementation.
We need an intern :)

amontoison:
It will be a lot of work to support a C interface...
I close this pull request.

PR #406, opened 2021-08-12
Add in-place stats for SYMMLQ

amontoison:
#361

### DIFF
--- a/test/test_alloc.jl
+++ b/test/test_alloc.jl
@@ -24,7 +24,7 @@ function test_alloc()
   solver = SymmlqSolver(A, b)
   symmlq!(solver, A, b)  # warmup
   inplace_symmlq_bytes = @allocated symmlq!(solver, A, b)
-  @test (VERSION < v"1.5") || (inplace_symmlq_bytes ≤ 672)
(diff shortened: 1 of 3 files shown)


test/test_alloc.jl
```diff
@@ -24,7 +24,7 @@ function test_alloc()
   solver = SymmlqSolver(A, b)
   symmlq!(solver, A, b)  # warmup
   inplace_symmlq_bytes = @allocated symmlq!(solver, A, b)
-  @test (VERSION < v"1.5") || (inplace_symmlq_bytes ≤ 672)
+  @test (VERSION < v"1.5") || (inplace_symmlq_bytes ≤ 624)
```
tmigot, line 27: I am slightly surprised that we saved only 48 bytes ?
I had another version here [tmigot/Krylov.jl#inplace-symmlq](https://github.com/tmigot/Krylov.jl/blob/inplace-symmlq) with 240.

amontoison: You forgot `reset!(stats)`.
`reset!` allocates for SymmlqStats.

PR #410, opened 2021-08-17
Add callback lsmr pr

AntoninKns:
Adding a callback function to lsmr, I could later do it for other solvers if needed.
An example on how to use it is added in the docstring.
Tests to cover it have been added in test_lsmr.jl

### DIFF
--- a/src/lsmr.jl
+++ b/src/lsmr.jl
@@ -92,7 +92,12 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
                atol :: T=zero(T), rtol :: T=zero(T),
                etol :: T=√eps(T), window :: Int=5,
                itmax :: Int=0, conlim :: T=1/√eps(T),
-               radius :: T=zero(T), verbose :: Int=0, history :: Bool=false) where {T <: AbstractFloat, S <: DenseVector{T}}
+               radius :: T=zero(T), verbose :: Int=0, history :: Bool=false,
+               extra_sc :: Union{Nothing, Function}=nothing) where {T <: AbstractFloat, S <: DenseVector{T}}
+
+  if extra_sc !== nothing
+    history = true
+  end
 
   m, n = size(A)
   length(b) == m || error("Inconsistent problem size")
@@ -204,6 +209,7 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
   ill_cond = ill_cond_mach = ill_cond_lim = false
   zero_resid = zero_resid_mach = zero_resid_lim = false
   fwd_err = false
+  extra_sc_bool = false
 
   while ! (solved || tired || ill_cond)
     iter = iter + 1
@@ -322,6 +328,11 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
     zero_resid_lim = (test1 ≤ rNormtol)
     iter ≥ window && (fwd_err = err_lbnd ≤ etol * sqrt(xENorm²))
 
+    # Stopping condition callback
+    if extra_sc !== nothing
+      extra_sc_bool = extra_sc(solver, A, b, iter, rNorms, ArNorms)
+    end
+
     ill_cond = ill_cond_mach | ill_cond_lim
     solved = solved_mach | solved_lim | solved_opt | zero_resid_mach | zero_resid_lim | fwd_err | on_boundary
   end
@@ -334,6 +345,7 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
   zero_resid    && (status = "found approximate zero-residual solution")
   fwd_err       && (status = "truncated forward error small enough")
   on_boundary   && (status = "on trust-region boundary")
+  extra_sc_bool && (status = "extra stop condition triggered")
 
   # Update stats
   stats.solved = solved
--- a/src/lsmr.jl
+++ b/src/lsmr.jl
@@ -77,6 +77,9 @@ The system above represents the optimality conditions of
 In this case, `N` can still be specified and indicates the weighted norm in which `x` and `Aᵀr` should be measured.
 `r` can be recovered by computing `E⁻¹(b - Ax)`.
 
+In order to use callback follow this example :
+callback = example_callback(solver, iter, rNorms, ArNorms)
+
 #### Reference
 
 * D. C.-L. Fong and M. A. Saunders, *LSMR: An Iterative Algorithm for Sparse*, Least Squares Problems, SIAM Journal on Scientific Computing, 33(5), pp. 2950--2971, 2011.
@@ -93,11 +96,7 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
                etol :: T=√eps(T), window :: Int=5,
                itmax :: Int=0, conlim :: T=1/√eps(T),
                radius :: T=zero(T), verbose :: Int=0, history :: Bool=false,
-               extra_sc :: Union{Nothing, Function}=nothing) where {T <: AbstractFloat, S <: DenseVector{T}}
-
-  if extra_sc !== nothing
-    history = true
-  end
+               callback :: F = (args...) -> false ) where {T <: AbstractFloat, S <: DenseVector{T}, F <: Function}
 
   m, n = size(A)
   length(b) == m || error("Inconsistent problem size")
@@ -209,9 +208,9 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
   ill_cond = ill_cond_mach = ill_cond_lim = false
   zero_resid = zero_resid_mach = zero_resid_lim = false
   fwd_err = false
-  extra_sc_bool = false
+  user_requested_exit = false
 
-  while ! (solved || tired || ill_cond)
+  while ! (solved || tired || ill_cond || user_requested_exit)
     iter = iter + 1
 
     # Generate next Golub-Kahan vectors.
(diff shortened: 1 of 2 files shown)


src/lsmr.jl
```diff
@@ -77,6 +77,9 @@ The system above represents the optimality conditions of
 In this case, `N` can still be specified and indicates the weighted norm in which `x` and `Aᵀr` should be measured.
 `r` can be recovered by computing `E⁻¹(b - Ax)`.
 
+In order to use callback follow this example :
+callback = example_callback(solver, iter, rNorms, ArNorms)
```
dpo, line 81: `rNorms` and `ArNorms` are in `stats`, which is in `solver`.

dpo: Please explain what the callback should return.

test/test_lsmr.jl
```diff
@@ -83,4 +83,18 @@
     N⁻¹ = inv(N)
     (x, stats) = lsmr(A, b, M=M⁻¹, N=N⁻¹, sqd=true)
   end
+
+  # Test callback function
+  function test_callback(solver, iter, rNorms, ArNorms)
```
dpo, line 88: `return iter >= 1`

test/test_lsmr.jl
```diff
@@ -83,4 +83,18 @@
...
+    if iter >= 1
+      return true
+    else
+      return false
+    end
+  end
+
+  (x, stats) = lsmr(A,b,callback = test_callback, history=true)
```
dpo, line 96:
```suggestion
  (x, stats) = lsmr(A, b, callback = test_callback, history = true)
```

src/lsmr.jl
```diff
@@ -322,18 +327,23 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
...
-  fwd_err       && (status = "truncated forward error small enough")
-  on_boundary   && (status = "on trust-region boundary")
+  tired               && (status = "maximum number of iterations exceeded")
+  ill_cond_mach       && (status = "condition number seems too large for this machine")
+  ill_cond_lim        && (status = "condition number exceeds tolerance")
+  solved              && (status = "found approximate minimum least-squares solution")
+  zero_resid          && (status = "found approximate zero-residual solution")
+  fwd_err             && (status = "truncated forward error small enough")
+  on_boundary         && (status = "on trust-region boundary")
+  user_requested_exit && (status = "extra stop condition triggered")
```
dpo, line 346:
```suggestion
  user_requested_exit && (status = "user-requested exit")
```

src/lsmr.jl
```diff
@@ -322,18 +327,23 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
     zero_resid_lim = (test1 ≤ rNormtol)
     iter ≥ window && (fwd_err = err_lbnd ≤ etol * sqrt(xENorm²))
 
+    # Stopping condition callback
```
dpo, line 330:
```suggestion
```

src/lsmr.jl
```diff
@@ -322,18 +327,23 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
...
+    # History must be set to true to have access to rNorms and ArNorms
```
dpo, line 331: This should be in the docstring instead

dpo (changes requested):

src/lsmr.jl
AntoninKns: I removed rNorms and ArNorms and gave a more in depth example in the docstring

test/test_lsmr.jl
AntoninKns: I changed it as you mentioned

test/test_lsmr.jl
AntoninKns: I changed it as you mentioned

src/lsmr.jl
AntoninKns: I changed it as you mentioned

src/lsmr.jl
AntoninKns: I removed it

src/lsmr.jl
AntoninKns: I moved it in the docstring as you mentioned

src/lsmr.jl
```diff
@@ -77,6 +77,19 @@ The system above represents the optimality conditions of
 In this case, `N` can still be specified and indicates the weighted norm in which `x` and `Aᵀr` should be measured.
 `r` can be recovered by computing `E⁻¹(b - Ax)`.
 
+In order to use callback follow this example :
+callback = example_callback(solver, iter)
```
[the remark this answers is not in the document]
AntoninKns: I changed it this way, it is more straightforward indeed

src/lsmr.jl
```diff
@@ -91,7 +96,8 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
                λ :: T=zero(T), axtol :: T=√eps(T), btol :: T=√eps(T),
                atol :: T=zero(T), rtol :: T=zero(T),
                etol :: T=√eps(T), itmax :: Int=0, conlim :: T=1/√eps(T),
-               radius :: T=zero(T), verbose :: Int=0, history :: Bool=false) where {T <: AbstractFloat, S <: DenseVector{T}}
+               radius :: T=zero(T), verbose :: Int=0, history :: Bool=false,
+               callback :: F = (args...) -> false ) where {T <: AbstractFloat, S <: DenseVector{T}, F <: Function}
```
amontoison, line 100:
```suggestion
               callback :: F = (args...) -> false) where {T <: AbstractFloat, S <: DenseVector{T}, F <: Function}
```

amontoison (changes requested):
Can you also update the signature of the function in the docstring ?
```
    (x, stats) = lsmr(A, b::AbstractVector{T};
                      M=I, N=I, sqd::Bool=false,
                      λ::T=zero(T), axtol::T=√eps(T), btol::T=√eps(T),
                      atol::T=zero(T), rtol::T=zero(T),
                      etol::T=√eps(T), window::Int=5,
                      itmax::Int=0, conlim::T=1/√eps(T),
                      radius::T=zero(T), verbose::Int=0,
                      history::Bool=false, callback::Function) where T <: AbstractFloat
```

src/lsmr.jl
AntoninKns: I changed it as you mentioned

AntoninKns (commented):
Changements asked have been made

src/lsmr.jl
```diff
@@ -26,13 +26,14 @@ export lsmr, lsmr!
...
 
 """
-    (x, stats) = lsmr(A, b::AbstractVector{T};
-                      M=I, N=I, sqd::Bool=false,
-                      λ::T=zero(T), axtol::T=√eps(T), btol::T=√eps(T),
-                      atol::T=zero(T), rtol::T=zero(T),
-                      etol::T=√eps(T), window::Int=5,
-                      itmax::Int=0, conlim::T=1/√eps(T),
-                      radius::T=zero(T), verbose::Int=0, history::Bool=false) where T <: AbstractFloat
+(x, stats) = lsmr(A, b::AbstractVector{T};
```
dpo, line 29: Please restore the indentation here. This calling sequence needs to be displayed as code by the markdown renderer.

src/lsmr.jl
```diff
@@ -97,7 +98,7 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
                atol :: T=zero(T), rtol :: T=zero(T),
                etol :: T=√eps(T), itmax :: Int=0, conlim :: T=1/√eps(T),
                radius :: T=zero(T), verbose :: Int=0, history :: Bool=false,
-               callback :: F = (args...) -> false ) where {T <: AbstractFloat, S <: DenseVector{T}, F <: Function}
+               callback :: Function = (args...) -> false) where {T <: AbstractFloat, S <: DenseVector{T}}
```
dpo, line 101: Why did this change? I don't think this will work in all cases (functions, anonymous functions, closures, ...)

src/lsmr.jl
```diff
@@ -329,7 +330,7 @@ function lsmr!(solver :: LsmrSolver{T,S}, A, b :: AbstractVector{T};
     zero_resid_lim = (test1 ≤ rNormtol)
     iter ≥ window && (fwd_err = err_lbnd ≤ etol * sqrt(xENorm²))
 
-    user_requested_exit = callback(solver, iter)
+    user_requested_exit = callback(solver, iter) :: Bool
```
dpo, line 333: Why was this necessary?

dpo (changes requested):
Please explain why you change parts of the code.

src/lsmr.jl
AntoninKns: I talked with Alexis, and this would avoid the following problem : if you want to use lsmr with "callback1" and then with "callback2" julia would compile the lsmr function 2 times, but if you think it would be better with the latest version I can change it back

src/lsmr.jl
AntoninKns: Sorry, I will change it back

src/lsmr.jl
AntoninKns: The idea was to force the boolean answer from the function, but I can change it back if you thin it would be better

src/lsmr.jl
dpo: Can you see if you can pass callbacks defined in several ways. For example:
```julia
lsmr(..., callback = (args...) -> false)
lmsr(..., callback = (args...) -> begin return false; end)
f1(solver, iter) = iter ≥ 1
lsmr(..., callback = f1)
```
etc.

AntoninKns: I tested with the old and new version and tests with different types of functions seem to work well. I also added a test to test what happens if the function returns a non boolean

src/lsmr.jl
AntoninKns: This has been changed

src/lsmr.jl
AntoninKns: I added a test to check what happens if the function returns a non boolean, and with this the error handling seems better

PR #416, opened 2021-08-24
Add restart in CG

amontoison:
@dpo


dpo:
this is a very interesting way to implement the initial guess feature. A restart is exactly what that is.

dpo (approved):

PR #424, opened 2021-08-26
Add functions nsolution, solution, statistics and issolved

amontoison:
Related to #420.
It will be easier to add tests and update the documentation after #420.

### DIFF
--- a/src/krylov_solvers.jl
+++ b/src/krylov_solvers.jl
@@ -5,7 +5,7 @@ BilqSolver, QmrSolver, BilqrSolver, CglsSolver, CrlsSolver, CgneSolver, CrmrSolv
 LslqSolver, LsqrSolver, LsmrSolver, LnlqSolver, CraigSolver, CraigmrSolver,
 GmresSolver, FomSolver
 
-export solve!, solution, num_solution, statistics, solved, solved_primal, solved_dual
+export solve!, solution, nsolution, statistics, solved, solved_primal, solved_dual
 
 "Abstract type for using Krylov solvers in-place"
 abstract type KrylovSolver{T,S} end
@@ -1439,15 +1439,17 @@ function solve! end
     solution(solver)
 
 Return the solution(s) stored in the `solver`.
+Optionally you can specify which solution you want to recover,
+`solution(solver, 1)` returns `x` and `solution(solver, 2)` returns `y`.
 """
 function solution end
 
 """
-    num_solution(solver)
+    nsolution(solver)
 
-Return the number of solutions stored in the `solver`.
+Return the number of outputs of `solution(solver)`.
 """
-function num_solution end
+function nsolution end
 
 """
     statistics(solver)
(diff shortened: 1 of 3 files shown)


src/krylov_solvers.jl
```diff
@@ -1435,40 +1435,79 @@ Use the in-place Krylov method associated to `solver`.
...
-  (MinresQlpSolver     , :minres_qlp!)
-  (QmrSolver           , :qmr!       )
-  (GmresSolver         , :gmres!     )
-  (FomSolver           , :fom!       )
+"""
+    solution(solver)
+
+Return the solution(s) stored in the `solver`.
+"""
+function solution end
```
tmigot, line 1443: For type-stability, wouldn't it be better if `solution` always returns the same type (a vector)?
Maybe, we can use another name for the `solver.y`, like 'multiplier'?

amontoison: `solution` has a different method for each `KrylovSolver`, so Julia can easily inferred the return type.
I don't like `multiplier` because `solver.y` is the solution of an adjoint systems or the residual of a least-squares problem.

src/krylov_solvers.jl
```diff
@@ -1435,40 +1435,79 @@ Use the in-place Krylov method associated to `solver`.
...
+
+"""
+    num_solution(solver)
+
+Return the number of solutions stored in the `solver`.
```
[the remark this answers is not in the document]
amontoison: Yes, good point. I will update the docstring.

src/krylov_solvers.jl
amontoison: Maybe we could do something similar to `size`, `solution(solver)` returns all outputs, `solution(solver, 1)` returns `solver.x` and `solution(solver, 2)`returns `solver.y`.

amontoison:
Should we change `solved` function for `issolved` ?
If we keep `solved`, we won't be able to name a variable `øolved` too :thinking:

tmigot:
> Should we change `solved` function for `issolved` ?
> If we keep `solved`, we won't be able to name a variable `øolved` too 🤔

I like the `issolved`

tmigot (approved):

src/krylov_solvers.jl
```diff
@@ -1435,41 +1435,84 @@ Use the in-place Krylov method associated to `solver`.
...
+
+"""
+    statistics(solver)
+
+Return the statistics stored in the `solver`.
+"""
+function statistics end
+
+"""
+    solved(solver)
```
amontoison, line 1462:
```suggestion
    issolved(solver)
```

src/krylov_solvers.jl
```diff
@@ -1435,41 +1435,84 @@ Use the in-place Krylov method associated to `solver`.
...
+
+Return a boolean that determines whether the Krylov method associated to `solver` succeeded.
+"""
+function solved end
```
amontoison, line 1466:
```suggestion
function issolved end
```

PR #430, opened 2021-09-03
Generic `reset!` function

tmigot:
Solving the issue raised in #417 .
However, I don't see how to include the `LanczosShift` as adding
```
@eval function empty_field!(stats :: $T, i, ::Type{Vector{Vec}}) where {S, Vec<:AbstractVector}
  statfield = getfield(stats, i)
  for vec in statfield
    empty!(vec)
  end
end
```
seems to allocate (the type of `statfield` being problematic apparently...).

### DIFF
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -17,11 +17,6 @@ mutable struct SimpleStats{T} <: KrylovStats{T}
   status       :: String
 end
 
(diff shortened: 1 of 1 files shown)


amontoison (approved):

amontoison:
Thanks Tangi ! It's not a big issue if `LanczosShiftStats` has its own `reset!` function.

PR #438, opened 2021-09-22
Acond in MINRES stats

geoffroyleconte:


amontoison:
@geoffroyleconte Can you add the unit for `MinresStats` in `test/test_stats.jl` ?

dpo:
Other solvers estimate Acond. Shouldn't we include it in the generic stats?

amontoison:
Good idea, let's add it in the `SimpleStats` directly.

geoffroyleconte:
I think not all solvers using `SimpleStats` estimate `Acond` right?

dpo:
Probably not, but most *could*.

amontoison:
You can just add it in the `SimpleStats` and update `Acond` for `MINRES`.
It will be an empty vector for the other methods (like `Aresiduals` sometimes).
I will check if `Acond` is estimate for the other methods later :+1:

amontoison (approved):

dpo (approved):
Looks great! Thank you!

PR #445, opened 2021-11-12
An implementation of GPMR

amontoison:


amontoison:
@dpo can you just verify the docstring of GPMR before I merge ?
You don't need to review the Julia code (except if you want to do it).

src/gpmr.jl
```diff
@@ -0,0 +1,402 @@
...
+                         restart::Bool=false, λ::T=one(T), μ::T=one(T),
+                         verbose::Int=0, history::Bool=false) where T <: AbstractFloat
+
+GPMR solves the unsymmetric partitioned linear system
+
+    [ λI   A ] [ x ] = [ b ]
+    [  B  μI ] [ y ]   [ c ],
+
+where λ and μ are real numbers.
+`A` can have any shape and `B` has the shape of `Aᵀ`.
```
dpo, line 27: A must be nonzero?

src/gpmr.jl
```diff
@@ -0,0 +1,402 @@
...
+`b` and `c` must be both nonzero.
```
dpo, line 28:
```suggestion
`b` and `c` must both be nonzero.
```

src/gpmr.jl
```diff
@@ -0,0 +1,402 @@
...
+
+GPMR stops when `itmax` iterations are reached or when `‖rₖ‖ ≤ atol + ‖r₀‖ * rtol`.
+`atol` is an absolute tolerance and `rtol` is a relative tolerance.
+
+Additional details can be displayed if verbose mode is enabled (verbose > 0).
+Information will be displayed every `verbose` iterations.
+
+#### Reference
+
+* A. Montoison and D. Orban, *GPMR: An Iterative Method for Unsymmetric Partitioned Linear Systems*, Cahier du GERAD G-2021-62, GERAD, Montréal, 2021.
```
dpo, line 57: Maybe give the DOI?

dpo (commented):
Great! Thank you. Just a few minor comments.

src/gpmr.jl
amontoison: I would like to give the DOI with a link (`[doi:xyz](https://dx.doi.org/xyz)`) but it's only correctly displayed by Documenter.jl.
The output is not great in the REPL.

amontoison: If we use the title of the article for the link, the result is neat.
REPL
documenter

PR #458, opened 2021-11-30
Loosen preconditioner eltype check

### LINKED ISSUE #457: loosen preconditioner eltype check
A preconditioner doesn't necessarily need to have the same `eltype` as `A` or `b`. It should be fine if its `eltype` could be promoted, i.e. a preconditioner in Bool or Float32 can be promoted to Float64.

So the check can be modified to `is_concrete(promote_type(eltype(P),eltype(A))`.

https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/ac2e1869bc0fa16088a29607f05c703b0762d69d/src/gmres.jl#L53-L57

ref - https://github.com/SciML/LinearSolve.jl/pull/33
cc - @ChrisRackauckas

amontoison:
Fix #457 
@vpuri3 do you a need a minor release after this PR ?

@dpo @geoffroyleconte 
This modification will allow the use of a preconditioner with a lower precision than `A` and `b`.
For instance, you will be able to solve a linear system in double precision with a preconditioner stored / computed in single precision.


dpo (changes requested):
Please add a couple of unit tests.

amontoison:
> Please add a couple of unit tests.

I added some tests to verify that linear systems in double precision can be solved with a Jacobi preconditioner in single precision.

vpuri3:
yes please -  i can add the compat for the release. thank you very much!

dpo (approved):

dpo:
Looks great! Thank you!

PR #465, opened 2021-12-07
Add a docstring for all in-place Krylov methods

amontoison:


tmigot (approved):
That's great, thanks!

docs/src/solvers/as.md
```diff
@@ -1,4 +1,6 @@
 ```@docs
```
dpo, line 1: Could you add a comment at the top of these doc files to say what they are documenting? I guess "as" stands for "adjoint system"?!

dpo (commented):
(my suggestion could happen in another pr)

docs/src/solvers/as.md
amontoison: I used the trick of https://github.com/JuliaDocs/Documenter.jl/issues/674 to add a comment at the top of these doc files.
I reused the information of `make.jl` : https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/docs/make.jl#L12-L19.

dpo: Nice!

dpo (approved):

PR #469, opened 2021-12-11
Update `zero_resid` in `lsqr`

tmigot:
The returned status on the zero residual wasn't updated during the iteration process.


dpo (approved):
thanks!

PR #473, opened 2021-12-18
harmonize verbose mode

amontoison:
To harmonize the verbose mode of the different Krylov methods, I replaced in @dpo implementations the number of matrix-vector products by the number of iterations.

Even if the number of matrix-vector products is a more relevant metric, our stopping conditions are based on the maximum number of iterations (`itmax`) and it's easier for the user to understand what is the current iteration now.

We could also display the number of matrix-vector products but I don't see in which context it can be useful. If we want to count them for a paper, we should distinguish the matvec products between `A`, `Aᵀ` and potentially `B`. In this case we should use `LinearOperators` to do that correctly.


dpo (approved):
sounds good

PR #475, opened 2021-12-20
Add tests for verbose mode

amontoison:
open("test.txt", "w") do fileio
    redirect_stdout(fileio) do
      symmlq(A, b, verbose=1)
      Base.Libc.flush_cstdio()
    end
  end
  showed = read("test.txt", String)
  @test strip(showed, '\n') == strip(expected, '\n')


amontoison:
It should be easier to finalize this PR because all Krylov methods have a keyword argument `iostream`.

PR #477, opened 2022-01-09
Remove unnecessary eltype checks of preconditioners

ChrisRackauckas:
These aren't even used and should not be expected to be defined. Matrix-free preconditioners, like geometric multigrid, do not necessarily have a sense of what eltype is.


amontoison:
Thanks @ChrisRackauckas 
I will update the other Krylov methods and do a new release.

ChrisRackauckas:
Thanks!

PR #481, opened 2022-01-14
[QMR / GMRES / BICGSTAB] Support of complex numbers

platawiec:
This PR adds complex support for the following solvers:
* `qmr`
* `gmres`
* `bicgstab`

The intent is to show what it would take to start to address #103 , namely, the following for the remaining solvers:

* Separating the type of the history `T` (as in `SimpleStats{T}`) from the type parameter of the eltype the solver expects. This involves adding an additional solver type parameter
* Minor adjustments to the solver to respect solving over complex matrices.

As part of this PR, the type restrictions in `krylov_utils.jl` were relaxed, and meaningful tests for complex matrices were added.

PR #483, opened 2022-01-20
[GPMR] Support of complex numbers

amontoison:
I need #481 to add tests with complex numbers.

### DIFF
--- a/src/gpmr.jl
+++ b/src/gpmr.jl
@@ -15,7 +15,7 @@ export gpmr, gpmr!
     (x, y, stats) = gpmr(A, B, b::AbstractVector{FC}, c::AbstractVector{FC};
                          C=I, D=I, E=I, F=I, atol::T=√eps(T), rtol::T=√eps(T),
                          gsp::Bool=false, reorthogonalization::Bool=false, itmax::Int=0,
-                         restart::Bool=false, λ::T=one(T), μ::T=one(T),
+                         restart::Bool=false, λ::FC=one(FC), μ::FC=one(FC),
                          verbose::Int=0, history::Bool=false)
 
 `T` is an `AbstractFloat` such as `Float32`, `Float64` or `BigFloat`.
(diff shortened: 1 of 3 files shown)


src/gpmr.jl
```diff
@@ -169,8 +169,8 @@ function gpmr!(solver :: GpmrSolver{T,FC,S}, A, B, b :: AbstractVector{FC}, c ::
   γ ≠ 0 || error("c must be nonzero")
   @. U[1] = c₀ / γ
 
-  # Compute ‖r₀‖² = γ² + β²
-  rNorm = sqrt(γ^2 + β^2)
+  # Compute ‖r₀‖² = |γ|² + |β|²
+  rNorm = sqrt(abs(γ)^2 + abs(β)^2)
```
amontoison, line 173:
```suggestion
  # Compute ‖r₀‖² = γ² + β²
  rNorm = sqrt(γ^2 + β^2)
```

amontoison: I forgot that `γ` and `β` are real numbers...

src/gpmr.jl
```diff
@@ -397,8 +397,8 @@ function gpmr!(solver :: GpmrSolver{T,FC,S}, A, B, b :: AbstractVector{FC}, c ::
       pos = pos - j + 1               # position of rᵢ.ⱼ₋₁
     end
     # Rₖ can be singular is the system is inconsistent
```
dpo, line 399:
```suggestion
    # Rₖ can be singular if the system is inconsistent
```

dpo (changes requested):

src/gpmr.jl
amontoison: Thanks !
It's already fixed on the `main` branch (https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/gpmr.jl#L404).
The mistake will be removed after the rebase.

PR #484, opened 2022-01-20
Update tests with krylov solvers

amontoison:
I added a kwarg `show_stats` to display (or not) the KrylovStats of a KrylovSolver.

### DIFF
--- a/src/krylov_solvers.jl
+++ b/src/krylov_solvers.jl
@@ -1581,7 +1581,7 @@ for (KS, fun, nsol) in [
 end
 
 """
-    show(io, solver; show_stats)
+    show(io, solver; show_stats=true)
 
 Statistics of `solver` are displayed if `show_stats` is set to true.
 """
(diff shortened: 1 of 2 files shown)


src/krylov_solvers.jl
```diff
@@ -1580,7 +1580,7 @@ for (KS, fun, nsol) in [
   end
 end
 
-function show(io :: IO, solver :: KrylovSolver)
+function show(io :: IO, solver :: KrylovSolver; show_stats :: Bool=true)
```
tmigot, line 1583: I think that's a nice feature. However, I would add some docstring here as it is not a standard use of `show`.

tmigot (changes requested):
This would also be a nice feature to be mentioned in a tutorial on `KrylovSolvers` :).

amontoison:
We can add in our future tutorial with GridAp ;)

tmigot (approved):

Issue #13, opened 2016-09-25
Add iter on stats


abelsiqueira:
Add iter on stats (`cg.jl`, at least).

dpo:
I care less and less about the number of iterations because it's not representative of the computational effort. The most expensive operations in Krylov methods are usually the operator-vector products, so that's what I count. The rest is a O(1) number of vector and scalar operations per iteration.

dpo:
Does this seem reasonable? I think when comparing methods, the most intuitive basis for comparison is the number of operator-vector products (and transposed operator-vector products).

abelsiqueira:
Yes, sorry, I forgot to reply. I was considering it for total number of inner iterations on tron, but op-v products work as well, and I also agree that is a better means of comparing.

dpo:
Ok, thanks.

PR #485, opened 2022-01-20
Add niterations, Aprod, Atprod and Bprod functions

### LINKED ISSUE #476: Add iter count to output stats
This is useful for downstream to know how many times that `A*x` was called.

amontoison:
fix #476 

@dpo, @geoffroyleconte, @tmigot, @abelsiqueira, @AntoninKns 
I updated the `KrylovStats` to store the number of iterations.
We don't need anymore to set `history=true` and compute `niter = length(stats.residuals) - 1)`.
We can directly recover `niter` with `stats.niter`.

The number of iterations is not a relevant metric in practice (https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/13, https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/232) but thanks to the `KrylovSolvers`, we can easily determine the number of operator-vector products with `A` and `A'`. :smiley: 
You just need to call `Aprod(solver)` and `Atprod(solver)`.

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -99,6 +99,7 @@ function bicgstab!(solver :: BicgstabSolver{T,S}, A, b :: AbstractVector{T}; c :
   rNorm = @knrm2(n, r)
   history && push!(rNorms, rNorm)
   if rNorm == 0
+    stats.niter = 0
     stats.solved, stats.inconsistent = true, false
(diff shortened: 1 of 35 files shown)


dpo (approved):
Number of iterations is irrelevant, but users keep wanting it.

PR #505, opened 2022-01-24
[MINRES-QLP] Support of complex numbers

amontoison:


src/minres_qlp.jl
```diff
@@ -116,15 +116,15 @@ function minres_qlp!(solver :: MinresQlpSolver{T,FC,S}, A, b :: AbstractVector{F
...
   ξₖ₋₁    = zero(T)
   τₖ₋₂    = τₖ₋₁ = τₖ = zero(T)
   ψbarₖ₋₂ = zero(T)
   μbisₖ₋₂ = μbarₖ₋₁ = zero(T)
-  wₖ₋₁ .= zero(T)
-  wₖ   .= zero(T)
-  cₖ₋₂  = cₖ₋₁ = cₖ = zero(T)  # Givens cosines used for the QR factorization of Tₖ₊₁.ₖ
+  wₖ₋₁ .= zero(FC)
+  wₖ   .= zero(FC)
+  cₖ₋₂  = cₖ₋₁ = cₖ = one(T)   # Givens cosines used for the QR factorization of Tₖ₊₁.ₖ
```
dpo, line 127: same weirdness as in USYMQR here. Maybe it just needs to be rebased?!

dpo (changes requested):

src/minres_qlp.jl
amontoison: Yes, you're right. It just needs to be rebased.

PR #527, opened 2022-02-07
Fix sqrt CR

geoffroyleconte:
Hi, I have numerical issues with CR used with RipQP:
https://github.com/JuliaSmoothOptimizers/RipQP.jl/pull/139
I think this PR should fix my issues, but I can't be sure because the tests only fail with MacOS.

### DIFF
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -264,7 +264,7 @@ function cr!(solver :: CrSolver{T,S}, A, b :: AbstractVector{T};
     @kaxpby!(n, one(T), Ar, β, q)
 
     pNorm² = rNorm² + 2 * β * pr - 2 * β * α * pAp + β^2 * pNorm²
-    pNorm = (pNorm² < zero(T)) ? zero(T) : sqrt(pNorm²)
+    pNorm = (pNorm² < eps(T)) ? zero(T) : sqrt(pNorm²)
     pr = rNorm² + β * pr - β * α * pAp # pᵀr
     abspr = abs(pr)
     pAp = ρ + β^2 * pAp # pᵀq
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -269,7 +269,11 @@ function cr!(solver :: CrSolver{T,S}, A, b :: AbstractVector{T};
     elseif abs(pNorm²) ≤ sqrt(eps(T))
       pNorm = zero(T)
     else
-      error("cr encountered numerical issues")
+      stats.niter = iter
+      stats.solved = solved
+      stats.inconsistent = false
+      stats.status = "solver incountered numerical issues"
+      return solver
     end
     pr = rNorm² + β * pr - β * α * pAp # pᵀr
     abspr = abs(pr)
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -272,7 +272,7 @@ function cr!(solver :: CrSolver{T,S}, A, b :: AbstractVector{T};
       stats.niter = iter
       stats.solved = solved
       stats.inconsistent = false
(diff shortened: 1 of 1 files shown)


src/cr.jl
```diff
@@ -264,7 +264,7 @@ function cr!(solver :: CrSolver{T,S}, A, b :: AbstractVector{T};
     @kaxpby!(n, one(T), Ar, β, q)
 
     pNorm² = rNorm² + 2 * β * pr - 2 * β * α * pAp + β^2 * pNorm²
-    pNorm = sqrt(pNorm²)
+    pNorm = (pNorm² < zero(T)) ? zero(T) : sqrt(pNorm²)
```
dpo, line 267: That's interesting. It must be an instability in the recursion formula for ‖p‖². For years, I've been wondering if that could be an issue but we never hit an example. So thank you!

For now, please only set `pNorm` to zero if `abs(pNorm^)` is very, very small. Otherwise, it would be a math bug.

amontoison, line 267:
```suggestion
    pNorm = (pNorm² < eps(T)) ? zero(T) : sqrt(pNorm²)
```

dpo: I guess. We have to keep in mind that it's a patch. It would be great to add a comment in the code with a `TODO` or `FIXME`.

geoffroyleconte: Here is the error:
 ```
DomainError with -6.841306819148919e-16:
  sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
```
The only issue is when `pNorm < 0`. What do you want me to change?

Comment (geoffroyleconte) [reply src/cr.jl]: @dpo I don't understand. The value I get is negative so it's smaller than `0` and `eps(T)`

Comment (dpo) [reply src/cr.jl]: It should be smaller in absolute value.

File: src/cr.jl [hunk 2]
```diff
@@ -264,7 +264,13 @@ function cr!(solver :: CrSolver{T,S}, A, b :: AbstractVector{T};
     @kaxpby!(n, one(T), Ar, β, q)
 
     pNorm² = rNorm² + 2 * β * pr - 2 * β * α * pAp + β^2 * pNorm²
-    pNorm = (pNorm² < eps(T)) ? zero(T) : sqrt(pNorm²)
+    if pNorm² > sqrt(eps(T))
+      pNorm = sqrt(pNorm²)
+    elseif abs(pNorm²) ≤ sqrt(eps(T))
+      pNorm = zero(T)
+    else
+      error("cr encountered numerical issues")
```
dpo, line 272: Instead of just crashing, you could exit the solver with an error status.

geoffroyleconte:
Are the JSOSolvers.jl test failures due to this PR? The error for the "latest" version mentions CG

src/cr.jl
```diff
@@ -264,7 +264,17 @@ function cr!(solver :: CrSolver{T,S}, A, b :: AbstractVector{T};
...
-    pNorm = sqrt(pNorm²)
+    if pNorm² > sqrt(eps(T))
+      pNorm = sqrt(pNorm²)
+    elseif abs(pNorm²) ≤ sqrt(eps(T))
+      pNorm = zero(T)
+    else
+      stats.niter = iter
+      stats.solved = solved
+      stats.inconsistent = false
+      stats.status = "solver incountered numerical issues"
```
dpo, line 275:
```suggestion
      stats.status = "solver encountered numerical issues"
```

dpo (changes requested):
Almost there.

dpo (approved):

PR #537, opened 2022-02-14
Update docstrings of LSLQ, LSQR, LSMR

amontoison:
@geoffroyleconte 
The `sqd` parameter was only used to set `λ=1`.
I propose to use `λ` parameter directly.

### DIFF
--- a/docs/src/examples.md
+++ b/docs/src/examples.md
@@ -504,11 +504,10 @@ b = MatrixMarket.mmread(joinpath(path[1], "$(matrix.name[1])_b.mtx"))[:]
 (m, n) = size(A)
 @printf("System size: %d rows and %d columns\n", m, n)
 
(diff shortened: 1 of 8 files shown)


dpo:
I find that this makes it much less obvious to the non-expert how to solve a SQD system.

geoffroyleconte:
At least the effect of `sqd` on `λ` should be documented.

Also @dpo is it possible to have some information about `axtol` and `btol`? How are they different from `atol` and `rtol` and in which case should we use them?

dpo:
@geoffroyleconte They implement other stopping conditions, e.g., https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/lsqr.jl#L202

The docstring should ideally explain stopping conditions.

geoffroyleconte (approved):

PR #538, opened 2022-02-14
Update docstrings of LNLQ, CRAIG, CRAIGMR

amontoison:
In the same spirit of #537 
I also fixed some errors in the docstrings.

### DIFF
--- a/src/craigmr.jl
+++ b/src/craigmr.jl
@@ -46,32 +46,30 @@ but is more stable. When λ = 0, this method solves the minimum-norm problem
 
     min ‖x‖  s.t.  x ∈ argmin ‖Ax - b‖.
 
-When λ > 0, this method solves the problem
+If `λ > 0`, CRAIGMR solves the symmetric and quasi-definite system
 
-    min ‖(x,y)‖₂  s.t.  Ax + λ²y = b.
-
-If `sqd = true`, CRAIGMR solves the symmetric and quasi-definite system
-
-    [ -F   Aᵀ ] [ x ]   [ 0 ]
-    [  A   E  ] [ y ] = [ b ],
+    [ -F    Aᵀ ] [ x ]   [ 0 ]
+    [  A  λ²E  ] [ y ] = [ b ],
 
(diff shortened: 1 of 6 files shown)


dpo:
Same comment as in #537. What's the motivation for this change?

src/craigmr.jl
```diff
@@ -46,32 +46,28 @@ but is more stable. When λ = 0, this method solves the minimum-norm problem
 
     min ‖x‖  s.t.  x ∈ argmin ‖Ax - b‖.
 
-When λ > 0, this method solves the problem
+If `λ > 0`, CRAIGMR solves the symmetric and quasi-definite system
 
-    min ‖(x,y)‖₂  s.t.  Ax + λ²y = b.
```
amontoison, line 51: I removed this relation because it was wrong.
```
min ‖(x,λy)‖₂ s.t. Ax + λ²y = b.
```
or
```
min ‖x‖² + λ²‖y‖²   s.t.  Ax + λ²y = b.
```

dpo:
What I mean is that it's no longer obvious that the user should set λ = 1.

amontoison:
> Same comment as in #537. What's the motivation for this change?

`sqd=true` silently changes the value of `λ` to 1 even if the user provides an other value.
We could only set `λ=1` directly.

amontoison:
> What I mean is that it's no longer obvious that the user should set λ = 1.

Ok, let's keep `sqd` kwarg and add in the docstring that `sqd=true` overwrite the value of `λ`.

amontoison:
The initial issue is that Geoffroy wants to solve SQD systems with preconditioners of the form `λI` without forming `F=λI` explicitly.
It's possible to direcly change the `λ` parameter in LS / LN methods and keep `F=I` but it was not documented.
And because he solves SQD systems, he sets `sqd` to `true`.

amontoison:
@JSOBot runbenchmarks cg_bmark.jl

geoffroyleconte (approved):

PR #540, opened 2022-02-14
Use `Krylov.jl` for `llsmodels`

tmigot:
It uses `Require` so `LLSModels` would be a dependency.
We ignore potential constraints in `LLSModels`.

I believe it would be a neat feature for JSO users. What do you think?

This could be documented in a separate tutorial.


amontoison:
Tangi, don't you think that the user is able to do `cgls(lls.A, lls.b)` ?

dpo:
I find this pretty neat and high level. It's intuitive to say "solve this least-squares problem with LSQR".

tmigot:
> 

I think that a user doesn't have to know what's inside the llsmodels, and it's pretty convenient to just say "solve this".

amontoison:
Ok, but I propose to directly add them in `LLSModels.jl` in that case.
All Krylov methods are not relevant for linear least squares problems and some Krylov methods requires particular 
inputs, such as two initial vectors or two initial matrices, etc....

I suggest to do what Geoffroy did for RipQP: https://github.com/JuliaSmoothOptimizers/RipQP.jl/blob/main/src/iterations/solvers/Krylov_utils.jl 
He only uses Krylov methods relevant for him and the arguments of the Krylov methods are adapted for its IPM formulations (K1, K2, K3.5, etc...).
The `lsqr`, `cgls`, `lsmr`, ... methods specialized for LLSModels should be also modified to check that you don't have constraints.

dpo:
Sounds good to me! (ps: that KSolver thing should be a Dict!)

tmigot:
Moved to `LLSModels.jl`

PR #544, opened 2022-02-21
Add a constant KRYLOV_SOLVERS

amontoison:
@tmigot 
Do you want this constant exported ?


tmigot (approved):
LGTM
No need to export I think.
Is there a way this is useful in the unit tests? That would serve as a test for this.

dpo (approved):

tmigot (approved):

PR #558, opened 2022-04-07
Add allocation tests for complex linear systems

amontoison:
I have some issues with TriCG... :thinking: 
We have many allocations with the in-place method but only when we solve complex linear systems.

The good news is that the other 30 methods don't allocate with their in-place method when we solve complex linear systems !


amontoison:
The tests passed with Julia nightly.
Hum, it will not be easy to find the origin of the problem.

amontoison:
I trackad memory allocations with `julia --track-allocation=user` and I found the culprit in TriCG:
tricg

I replaced these lines with `@axpy!` commands and it solved the problem.
The issue is probably at the compiler level and they fixed it because we don't have allocation with Julia nighly.

PR #562, opened 2022-04-21
[MINRES] Move restart option to MinresSolver

geoffroyleconte:
I found the use of the restart option not convenient to write generic code. Also the fact that the `restart` parameter was in the arguments of the function was misleading because it was not possible to restart `minres` (the not in-place version) (correct me if I'm wrong).

I added a function `warm_start!(solver, x0)` that is used to tell MINRES to solve the problem from x0. 
`MinresSolver` has now an attribute `restart` set to false by default. When using `warm_start!`, it is set to true. After `minres!` is called, it is set to false again.

If we agree to merge with, we could do the same for all other Krylov solvers by adding a method to the `warm_start!` function.

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -28,8 +28,8 @@ export minres, minres!
                         rtol::T=√eps(T)/100, ratol :: T=zero(T), 
                         rrtol :: T=zero(T), etol::T=√eps(T),
                         window::Int=5, itmax::Int=0,
-                        conlim::T=1/√eps(T), restart::Bool=false,
-                        verbose::Int=0, history::Bool=false)
+                        conlim::T=1/√eps(T), verbose::Int=0,
+                        history::Bool=false)
 
 `T` is an `AbstractFloat` such as `Float32`, `Float64` or `BigFloat`.
 `FC` is `T` or `Complex{T}`.
@@ -74,10 +74,11 @@ See [`MinresSolver`](@ref) for more details about the `solver`.
 function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
                  M=I, λ :: T=zero(T), atol :: T=√eps(T)/100, rtol :: T=√eps(T)/100, 
                  ratol :: T=zero(T), rrtol :: T=zero(T), etol :: T=√eps(T),
-                 itmax :: Int=0, conlim :: T=1/√eps(T), restart :: Bool=false,
-                 verbose :: Int=0, history :: Bool=false) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: DenseVector{FC}}
+                 itmax :: Int=0, conlim :: T=1/√eps(T), verbose :: Int=0,
+                 history :: Bool=false) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: DenseVector{FC}}
 
   n, m = size(A)
+  restart = solver.restart
   m == n || error("System must be square")
   length(b) == n || error("Inconsistent problem size")
   (verbose > 0) && @printf("MINRES: system of size %d\n", n)
@@ -102,7 +103,6 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
   ctol = conlim > 0 ? 1 / conlim : zero(T)
 
   # Initial solution x₀
-  restart && (Δx .= x)
   x .= zero(FC)
 
   if restart
@@ -126,6 +126,7 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
     history && push!(rNorms, β₁)
     history && push!(ArNorms, zero(T))
     history && push!(Aconds, zero(T))
+    solver.restart = false
     return solver
   end
   β₁ = sqrt(β₁)
@@ -271,6 +272,7 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
       stats.niter = 0
       stats.solved, stats.inconsistent = true, true
       stats.status = "x is a minimum least-squares solution"
+      solver.restart = false
       return solver
     end
 
@@ -312,5 +314,6 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
   stats.solved = solved
   stats.inconsistent = !zero_resid
   stats.status = status
+  solver.restart = false
   return solver
 end
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -92,7 +92,6 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
 
   # Set up workspace.
   allocate_if(!MisI  , solver, :v , S, n)
-  allocate_if(restart, solver, :Δx, S, n)
   Δx, x, r1, r2, w1, w2, y = solver.Δx, solver.x, solver.r1, solver.r2, solver.w1, solver.w2, solver.y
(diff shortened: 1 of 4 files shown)


dpo:
The use of the word "restart" is very confusing here and we should remove it. There is no restarting MINRES because it's based on a 3-term recurrence. The use of "restart" is reserved for methods based on the Arnoldi process (FOM, GMRES, and now, GPMR). We should simply be able to pass an initial guess to `solve!()` and what you're doing in `warm_start!()` should happen.

geoffroyleconte:
@dpo I don't think that `x0` should be provided when calling `minres!`. I find it easier to specify `x0` in the `MinresSolver` to write generic code with different Krylov Methods. 
Also after discussing with @amontoison some methods such as GPMR would require in the warm start process the inverse of the diagonal block (`M` and `N` when we require `inv(M)` and `inv(N)`).
Should i change the `restart` boolean attribute of `MinresSolver` to `warm_start`?

dpo:
In my opinion, `x0` is not related to the solver, but to the solve. The role of the solver object is to preallocate memory. However, we should be able to perform several solves with the same solver object, for example with different right-hand sides or different initial guesses.

geoffroyleconte:
But in this case, what should be the default value of `x0`? `nothing`? Because if we set it to a vector of zeros we allocate a vector at each `minres!` call.
An other advantage of `warm_start!` is that it removes the `Δx` allocation from `minres!` (I just changed this in my last commit because I forgot to remove it).

test/test_minres.jl
```diff
@@ -74,10 +74,16 @@
       solver = MinresSolver(A, b)
       minres!(solver, A, b, itmax=50)
       @test !solver.stats.solved
-      minres!(solver, A, b, restart=true)
-      r = b - A * solver.x
-      resid = norm(r) / norm(b)
-      @test(resid ≤ minres_tol)
+      @test solver.warm_start == false
+      (x, stats) = minres(A, b, solver.x)
```
dpo, line 78: Are you restarting it from the solution? Why not run the first solve with relaxed tolerances? Otherwise I supposed the second solve doesn't perform any iterations at all?!

geoffroyleconte: I think the tests are the same, `minres(A, b, solver.x)` creates internally a new `MinresSolver` that is different from `solver`

dpo: but isn't it starting from the solution?

geoffroyleconte: Oh yes my bad I forgot to add `itmax = 50`

dpo (approved):
ok great, thank you!

src/minres.jl
```diff
@@ -64,20 +70,28 @@ function minres(A, b :: AbstractVector{FC}; window :: Int=5, kwargs...) where FC
...
 function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
                  M=I, λ :: T=zero(T), atol :: T=√eps(T)/100, rtol :: T=√eps(T)/100, 
                  ratol :: T=zero(T), rrtol :: T=zero(T), etol :: T=√eps(T),
-                 itmax :: Int=0, conlim :: T=1/√eps(T), restart :: Bool=false,
-                 verbose :: Int=0, history :: Bool=false) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: DenseVector{FC}}
+                 itmax :: Int=0, conlim :: T=1/√eps(T), verbose :: Int=0,
+                 history :: Bool=false) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: DenseVector{FC}}
 
   n, m = size(A)
+  warm_start = solver.warm_start
```
amontoison, line 94: Can you move this line in the section `# Set up workspace.` ?

src/minres.jl
```diff
@@ -64,20 +70,28 @@ function minres(A, b :: AbstractVector{FC}; window :: Int=5, kwargs...) where FC
   return (solver.x, solver.stats)
 end
 
+function minres(A, b :: AbstractVector{FC}, x0 :: AbstractVector{FC0}; window :: Int=5, kwargs...) where {FC <: FloatOrComplex, FC0 <: FloatOrComplex}
```
amontoison, line 73:
```suggestion
function minres(A, b :: AbstractVector{FC}, x0 :: AbstractVector{FC}; window :: Int=5, kwargs...) where {FC <: FloatOrComplex}
```

src/minres.jl
```diff
@@ -305,12 +319,19 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
...
   # Update stats
   stats.niter = iter
   stats.solved = solved
   stats.inconsistent = !zero_resid
   stats.status = status
+  solver.warm_start = false
   return solver
 end
+
+function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC}, x0::AbstractVector{FC0};
```
amontoison, line 333:
```suggestion
function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC}, x0::AbstractVector{FC};
```

src/minres.jl
```diff
@@ -305,12 +319,19 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
...
+                 kwargs...) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: DenseVector{FC}, FC0 <: FloatOrComplex}
```
amontoison, line 334:
```suggestion
                 kwargs...) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: DenseVector{FC}}
```

amontoison: Can you write lines 333-334 on the same one?

test/test_minres.jl
amontoison: Yes, It only gives an approximate solution after `itmax = 50` iterations.

amontoison (changes requested):
Geoffroy, is it ok to keep the same type for `b` and `x0` ?
I would like to avoid overparameterization.
If you want to `warm_start` from a solution with a different precision,
you will still be able to use the `warm_start!` function.

amontoison:
I like the current solution because we now have the warm start for the non-inplace method and `x0` can be given for the solve (`minres(A, b, x0)` and `minres!(solver, A, b, x0)`.
The previous implementation was not user-friendly (solver.x = x0) and it's also great that we can remove the misleading `restart` kwarg.

@dpo You are right that `x0` should not be related to the solver but if we want to have the same code of MINRES working for both methods (`minres(A,b)` and `minres(A, b, x0)`), you should add `warm_start` attribute in the solver and store the initial guess in the solver.

dpo:
It's ok to store x0 in the solver. I just didn't agree that it should be passed to the solver constructor.

src/minres.jl
geoffroyleconte: I'm not sure it is a good idea. If the user has an approximate solution in `Float32` and wants to refine the system in `Float64` he would have to allocate one more vector. I can change it to `x0::AbstractVector` if this is really an issue, but what's the problem with adding one more type parameter?

amontoison: If you add one more type parameter you will compile a new version of `minres` / `minres!` specialized for `FC0`.
We only need one method for all type of `x0`.

amontoison:
We have another issue Geoffroy, if the user provide `x0` in `minres` or `minres!`, `\Delta x` is not needed anymore to store the previous solution except if `x0 == solver.x`. It should be possible to do `solver.x .+= x0` at the end of the method.

amontoison (approved):

PR #568, opened 2022-04-27
Fix the documentation

amontoison:
@dpo, Documenter.jl is not able to display docstrings of function defined as:
```
"""
My function foo
"""
function foo end
```
if we have duplicates.
I didn't find how to solve this problem with removing the `@autodocs` block.

The docstring of `minres`, `gmres`, ... are not display anymore.

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -9,9 +9,4 @@ Pages = ["reference.md"]
 ## Index
 
 ```@index
-Pages = ["reference.md"]
-```
-
-```@autodocs
-Modules = [Krylov]
 ```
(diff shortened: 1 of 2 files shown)


geoffroyleconte:
Why don't you just remove ```function foo end```  ?
That's not necessary as long as you put the docstring above one of the methods of `foo`.

geoffroyleconte:
Why don't you just remove ```function foo end```  ?
That's not necessary as long as you put the docstring above one of the methods of `foo`.

geoffroyleconte (approved):

dpo:
Did you ask on Discourse and Slack?

amontoison:
Thanks for the fix @geoffroyleconte!

amontoison:
> Did you ask on Discourse and Slack?

No but I found similar issues in Documenter.jl and the solution is to remove the duplicate docstrings.
@geoffroyleconte just found a way to keep a list of all functions.

dpo:
ok

PR #579, opened 2022-05-06
Add cg_lanczos_shift function for cg_lanczos with shifts

amontoison:
I can't add warm-start feature in CG-LANCZOS without this modification.
The multiple dispatch is unable to decide which function is appropriate when a `x0` is provided.

```
cg_lanczos(A, b, x0)
cg_lanczos(A, b, shifts)
```
Note that each Krylov method will have its own KrylovSolver with this modification.


src/krylov_solvers.jl
```diff
@@ -1738,14 +1739,8 @@ for (KS, fun, nsol, nA, nAt, warm_start) in [
...
-          # Special case for KS == CgLanczosShiftSolver
-          if eltype(solver.x) <: AbstractVector
-            n = length(solver.x[1])
-            S = typeof(solver.x[1])
-          else
-            n = length(solver.x)
-            S = typeof(solver.x)
-          end
+          n = length(solver.x)
+          S = typeof(solver.x)
```
amontoison, line 1743:
```suggestion
          S = typeof(solver.x)
```

dpo:
Yes, that makes sense. But where is the warm-started version? I don't see it in here.

amontoison:
I will add it with another pull request.
12 files are already modified to just rename `cg_lanczos` whift shifts.

dpo (approved):
Ok! Thank you!

PR #586, opened 2022-05-30
[MINRES] callback

geoffroyleconte:


src/minres.jl
```diff
@@ -29,7 +29,8 @@ export minres, minres!
                         rrtol :: T=zero(T), etol::T=√eps(T),
                         window::Int=5, itmax::Int=0,
                         conlim::T=1/√eps(T), verbose::Int=0,
-                        history::Bool=false)
+                        history::Bool=false,
+                        callback::Function=(args...)->false)
```
dpo, line 33: Maybe use `<:Function` to be more general, so you can use anonymous functions, callable structs, etc.

src/minres.jl
```diff
@@ -313,6 +318,7 @@ function minres!(solver :: MinresSolver{T,FC,S}, A, b :: AbstractVector{FC};
     zero_resid_lim = (test1 ≤ tol)
     resid_decrease_lim = (rNorm ≤ rNormtol)
     iter ≥ window && (fwd_err = err_lbnd ≤ etol * sqrt(xENorm²))
+    user_requested_exit = callback(solver, A, b) :: Bool
```
dpo, line 321:
```suggestion

    user_requested_exit = callback(solver, A, b) :: Bool
```

dpo (commented):
Looks great! Thank you!

Does this work for you in RipQP?

src/minres.jl
geoffroyleconte: I'm not sure I understood your comment, it is not possible to replace `::Function` with `<:Function`.
This works:
```julia
mutable struct ResidData{T}
  resid::Vector{T}
end
rd = ResidData(rand(10).*1.0e-1)
(rd::ResidData)(solver, A, b) = all(abs.(A*solver.x - b) .≤ rd.resid)
x, stats = minres(A, b, callback = (solver, A, b) -> rd(solver, A, b)) # n = 10
```

geoffroyleconte:
> Looks great! Thank you!
> 
> Does this work for you in RipQP?

I did not try it with RipQP yet but as I can have stopping criteria based on a vector (as show in the above example with `ResidData` this should work fine without preconditioner.
But if I use a preconditioner I think I must add it to the callback arguments. Should I do it in this PR?

dpo (approved):

geoffroyleconte:
> > Looks great! Thank you!
> > Does this work for you in RipQP?
> 
> I did not try it with RipQP yet but as I can have stopping criteria based on a vector (as show in the above example with `ResidData`) this should work fine without preconditioner. But if I use a preconditioner I think I must add it to the callback arguments. Should I do it in this PR?

Actually In my above example with `ResidData` I could add the preconditioner to the fields of the struct to be able to use it in the callback so this should work anyway.

amontoison (approved):

amontoison:
Sorry for the delay

PR #588, opened 2022-06-03
Add more callbacks

geoffroyleconte:
- symmlq
- minres_qlp
- gmres
- dqgmres
- cg
- cg_lanczos
- cr
- bicgstab
- fom
- diom
- qmr

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -18,7 +18,8 @@ export bicgstab, bicgstab!
 """
     (x, stats) = bicgstab(A, b::AbstractVector{FC}; c::AbstractVector{FC}=b,
                           M=I, N=I, atol::T=√eps(T), rtol::T=√eps(T),
-                          itmax::Int=0, verbose::Int=0, history::Bool=false)
+                          itmax::Int=0, verbose::Int=0, history::Bool=false,
+                          callback::Function=(args...)->false)
 
 `T` is an `AbstractFloat` such as `Float32`, `Float64` or `BigFloat`.
 `FC` is `T` or `Complex{T}`.
@@ -46,6 +47,9 @@ BICGSTAB can be warm-started from an initial guess `x0` with the method
     (x, stats) = bicgstab(A, b, x0; kwargs...)
 
 where `kwargs` are the same keyword arguments as above.
+
+The callback is called as `callback(solver, A, b)` and should return `true` if the main loop should terminate,
+and `false` otherwise.
 
 #### References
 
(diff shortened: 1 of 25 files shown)


dpo (approved):
Looks great! Thank you!

@tmigot Is this the same kind of callback that you used in `cg_lanczos` with shifts?

tmigot:
> Looks great! Thank you!
> 
> @tmigot Is this the same kind of callback that you used in `cg_lanczos` with shifts?

It is! I just need a fourth argument for the shifts: `callback(solver, A, b, shifts)`.

dpo:
Would it make sense to store the shifts inside the solver object?

tmigot:
> Would it make sense to store the shifts inside the solver object?

I am not sure as it is part of the problem as much is `A` and `b`, what do you think?

dpo:
Do they change, or do only A and b change? Even if they change, we could have a method to update them inside the solver object.

tmigot:
> Do they change, or do only A and b change? Even if they change, we could have a method to update them inside the solver object.

In theory, we could decide to resolve the system with more shifts if we were having an adaptive way of selecting them.

geoffroyleconte:
@tmigot @dpo yes I will add different callback arguments for block methods and `cg_lanczos_shift` so I think `callback(solver, A, b, shifts)` makes sense.
I particular, can you check `bicgstab` in this PR? I left `callback(solver, A, b)` even though there is `c` because it is a keyword argument, but I can change it to `callback(solver, A, b, c)` if that is an issue.

dpo:
In fact, why do you need to send A and b back to the user? The user passed them to the Krylov method, so (s)he has them...

geoffroyleconte:
Yes, (s)he could directly add them to the default argument of the callback or to the structure if the callback is a `struct`. I thought that maybe it would be more user-friendly and easier to implement if only `A`, `b` and `solver` are needed for the callback (the arguments of the callback are the same as the arguments of the Krylov method), but I can remove `A` and `b` if you want.

dpo:
Yes I would think that `callback(solver)` is enough.

geoffroyleconte:
@dpo I've made the changes, can you review it once more please?
Also should I do it for all Krylov methods in this PR or can I separate it in several PRs?

dpo (approved):
Looks great! Thank you!

amontoison (approved):

amontoison:
Thanks @geoffroyleconte !

PR #590, opened 2022-06-07
Callbacks 2

geoffroyleconte:
- bilq
- bilqr
- trilqr
- cg_lanczos_shifts
- cgls
- crls
- lslq
- lsqr
- lsmr
- cgne
- crmr
- lnlq
- craig
- craigmr
- usymlq
- usymqr
- cgs
- tricg
- trimr
- gpmr

PR #591, opened 2022-06-08
Add documentation for callbacks

geoffroyleconte:
Once https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/590 is merged, I can rebase this PR to tackle https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/587.

### DIFF
--- a/docs/src/callbacks.md
+++ b/docs/src/callbacks.md
@@ -1,0 +1,43 @@
+## Callbacks
+
+Each Krylov method is able to use a callback function that is called as `callback(solver)`, and that should return `true` if the main loop should terminate, and `false` otherwise.
+If the method terminated because of the callback, the output status will be `"user-requested exit"`.
+For example, if the user has a function `my_callback(solver::MinresSolver)`, it can be used with:
+
+```julia
+(x, stats) = minres(A, b, callback = my_callback)
+```
+
+If you need to write a callback that uses variables that are not in the `MinresSolver`, it is possible to use an anonymous function:
+
+```julia
+function my_callback2(solver::MinresSolver, A, b, storage_vec, tol::Float64)
+  mul!(storage_vec, A, solver.x)
+  storage_vec .-= b
+  return norm(storage_vec) ≤ tol # tolerance based on the 2-norm of the residual
+end
+
+storage_vec = similar(b)
+(x, stats) = minres(A, b, callback = solver -> my_callback2(solver, A, b, storage_vec, 0.1))
+```
+
+You can also use a structure and call it as an anonymous function:
+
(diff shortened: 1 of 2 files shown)


docs/src/callbacks.md
```diff
@@ -0,0 +1,43 @@
+## Callbacks
+
+Each Krylov method is able to use a callback function that is called as `callback(solver)`, and that should return `true` if the main loop should terminate, and `false` otherwise.
+If the method terminated because of the callback, the output status will be `"user-requested exit"`.
+For example, if the user has a function `my_callback(solver::MinresSolver)`, it can be used with:
```
dpo, line 5:
```suggestion
For example, if the user has a function `my_callback(solver::MinresSolver)`, it can be used as
```

docs/src/callbacks.md
```diff
@@ -0,0 +1,43 @@
...
+MyCallback3(A, b; tol = 0.1) = MyCallback3(A, b, similar(b), tol)
+
+function (my_cb::MyCallback3)(solver)
+  mul!(my_cb.storage_vec, my_cb.A, solver.x)
+  my_cb.storage_vec .-= my_cb.b
+  return norm(my_cb.storage_vec) ≤ my_cb.tol # tolerance based on the 2-norm of the residual
+end
+
+my_cb = MyCallback3(A, b; tol = 0.1)
+(x, stats) = minres(A, b, callback = solver -> my_cb(solver))
```
dpo, line 42: Couldn't this be just `callback = my_cb`?

geoffroyleconte: No because `my_cb` is a struct of type `MyCallback3`, not a `Function`

dpo: But that's what I meant when I said it would be better to parametrize the type of `callback`. If the user has to define a function based on the struct, it defeats the purpose. We should be able to pass anything that's callable.

docs/src/callbacks.md
```diff
@@ -0,0 +1,43 @@
...
+  mul!(storage_vec, A, solver.x)
+  storage_vec .-= b
+  return norm(storage_vec) ≤ tol # tolerance based on the 2-norm of the residual
+end
+
+storage_vec = similar(b)
+(x, stats) = minres(A, b, callback = solver -> my_callback2(solver, A, b, storage_vec, 0.1))
+```
+
+You can also use a structure and call it as an anonymous function:
```
dpo, line 24:
```suggestion
Alternatively, use a structure and make it callable:
```

dpo (approved):

PR #593, opened 2022-06-09
[documentation] Add a page about the foundations of Krylov methods

amontoison:
I could give a link to this page to justify my message in this discussion: https://discourse.julialang.org/t/how-to-solve-this-ax-b-faster/81693/22


dpo:
Are the docs the right place to add such background? The background only talks about square systems. I think that if you really want some background in here, it should be less specific and more general: Krylov methods build approximations by constructing subspaces of increasing dimension and choosing the approximations using an optimization problem.

dpo (commented):
Are the docs the right place to add such background? The background only talks about square systems. I think that if you really want some background in here, it should be less specific and more general: Krylov methods build approximations by constructing subspaces of increasing dimension and choosing the approximations using an optimization problem.

PR #595, opened 2022-06-17
Machine Tolerances MINRES_QLP and SYMMLQ

geoffroyleconte:
- For MINRES_QLP, test1 is correctly computed at iter 3 (because x is not updated until iter 3, so `xNorm == 0`)
- I commented the `# zero_resid_lim = (test1 ≤ ε)` test, but I uncomment it if you want. I this case I will have to change the MINRES_QLP residual tests to `@test(Aresid ≤ minres_qlp_tol * norm(A) * norm(x))`.

### DIFF
--- a/src/minres_qlp.jl
+++ b/src/minres_qlp.jl
@@ -377,7 +377,7 @@ function minres_qlp!(solver :: MinresQlpSolver{T,FC,S}, A, b :: AbstractVector{F
     # Stopping conditions that do not depend on user input.
     # This is to guard against tolerances that are unreasonably small.
     ill_cond_mach = (one(T) + one(T) / Acond ≤ one(T))
-    resid_decrease_mach = (rNorm + one(T) ≤ one(T))
+    resid_decrease_mach = (one(T) + rNorm ≤ one(T))
     zero_resid_mach = (one(T) + backward ≤ one(T))
 
     # Stopping conditions based on user-provided tolerances.
--- a/src/symmlq.jl
+++ b/src/symmlq.jl
@@ -354,7 +354,7 @@ function symmlq!(solver :: SymmlqSolver{T,FC,S}, A, b :: AbstractVector{FC};
 
     # Stopping conditions that do not depend on user input.
     # This is to guard against tolerances that are unreasonably small.
-    resid_decrease_mach = (rNorm + one(T) ≤ one(T))
+    resid_decrease_mach = (one(T) + rNorm ≤ one(T))
     ill_cond_mach = (one(T) + one(T) / Acond ≤ one(T))
     zero_resid_mach = (one(T) + test1 ≤ one(T))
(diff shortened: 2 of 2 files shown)


src/minres_qlp.jl
```diff
@@ -345,12 +349,27 @@ function minres_qlp!(solver :: MinresQlpSolver{T,FC,S}, A, b :: AbstractVector{F
...
+    tired = iter ≥ itmax
+    resid_decrease_lim = (rNorm ≤ ε)
+    # zero_resid_lim = (test1 ≤ ε)
     breakdown = βₖ₊₁ ≤ btol
-    solved = rNorm ≤ ε
+
+    user_requested_exit = callback(solver) :: Bool
+    zero_resid = zero_resid_mach #| zero_resid_lim
+    resid_decrease = resid_decrease_mach | resid_decrease_lim
+    solved = resid_decrease || zero_resid
```
dpo, line 371: should this be just `|` for consistency?

amontoison (approved):

src/minres_qlp.jl
```diff
@@ -345,12 +355,42 @@ function minres_qlp!(solver :: MinresQlpSolver{T,FC,S}, A, b :: AbstractVector{F
...
+    history && push!(Aconds, Acond)
+    xNorm = @knrm2(n, x)
+    backward = rNorm / (ANorm * xNorm)
+
     # Update stopping criterion.
-    user_requested_exit = callback(solver) :: Bool
+    # Stopping conditions that do not depend on user input.
+    # This is to guard against tolerances that are unreasonably small.
+    ill_cond_mach = (one(T) + one(T) / Acond ≤ one(T))
+    resid_decrease_mach = (rNorm + one(T) ≤ one(T))
```
dpo, line 380: The other two are `one(T) + ...` and this one is `... + one(T)`. Let's use the same format everywhere if we can.

src/minres_qlp.jl
```diff
@@ -345,12 +355,42 @@ function minres_qlp!(solver :: MinresQlpSolver{T,FC,S}, A, b :: AbstractVector{F
...
+    elseif iter == 2
+      μmax = max(μmax, μbisₖ₋₁, abs_μbarₖ)
+      μmin = min(μmin, μbisₖ₋₁, abs_μbarₖ)
+    else
+      μmax = max(μmax, μₖ₋₂, μbisₖ₋₁, abs_μbarₖ)
+      μmin = min(μmin, μₖ₋₂, μbisₖ₋₁, abs_μbarₖ)
+    end
+    Acond = μmax / μmin
+    history && push!(Aconds, Acond)
+    xNorm = @knrm2(n, x)
```
amontoison, line 373: @dpo Do you have a good reference about backward errors?
We always compute `xNorm` in Euclidean norm here too.

PR #602, opened 2022-07-26
Bumping version to 0.8.3

AntoninKns:
Bumping krylov version to v0.8.3 in order to adapt to latest changes.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Krylov"
 uuid = "ba0b0d4f-ebba-5204-a429-3ac8c609bfb7"
-version = "0.8.2"
+version = "0.8.3"
 
 [deps]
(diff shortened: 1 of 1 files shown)


dpo:
I wonder why 0.8.2 doesn't appear in the releases.

AntoninKns:
I was wondering that too, because I can still download it.

amontoison:
> I wonder why 0.8.2 doesn't appear in the releases.

Let's see if it was just a bug with GitHub or an issue a workflow.

dpo:
Now you have 0.8.3 without 0.8.2.

amontoison:
> Now you have 0.8.3 without 0.8.2.

It's fixed, I added a tag v0.8.2 and the associated release.

PR #604, opened 2022-08-03
use strict=true in docs

dpo:


amontoison:
The problem with SSMC.jl can't be solved by us: https://github.com/JuliaSmoothOptimizers/SuiteSparseMatrixCollection.jl/issues/61

amontoison (approved):

amontoison:
Thanks @dpo!

Issue #605, opened 2022-08-15
Compatibility with BlockArrays


lijas:
I have a linear system in the form: A = [K C'; C, 0], (C is relatively small) which requires a large amount of iterations with minres/gmres in order to converge. There is no "out of the box" preconditioner that helps with lowering the amount of iterations. 

I dont know if this is the best way, but a few [tutorials](https://www.dealii.org/current/doxygen/deal.II/step_43.html#Thelinearsystemanditspreconditioning) suggests creating a custom preconditioner, P^-1 = [ilu(K) 0; 0 C'*inv(diag(K))*C] or something similar (perhaps there is a better preconditioner for this case), for which you preferable want to use BlockArrays and overload `ldiv!` in for your costum preconditier:
```

function LinearAlgebra.ldiv!(y::BlockVector, A::CostumBlockPrecon, b::BlockVector)
    y[Block(1)] = A.K_ilu\b[Block(1)] #precomputed ilu preconditioner
    y[Block(2)] = A.B * (diag(A.K)\(A.B*b[Block(2)]))
end
```

This is the error you get if you try to use BlockArrays+Krylov:

```
using Krylov
using BlockArrays

K = rand(10,10); 
B = rand(2, 10)
b = rand(10 + 2);

b = BlockVector(rand(Float64,12), [10,2])
A = BlockArray(rand(Float64,12,12), [10,2], [10,2])
A[Block(1,1)] = K
A[Block(1,2)] = B'
A[Block(2,1)] = B

Krylov.gmres(A, b)
```

```
ERROR: MethodError: no method matching BlockVector{Float64, Vector{Vector{Float64}}, Tuple{BlockedUnitRange{Vector{Int64}}}}(::UndefInitializer, ::Int64)
Closest candidates are:
  BlockArray{T, N, R, BS}(::UndefInitializer, ::Tuple{Vararg{Int64, N}}) where {T, N, R<:(AbstractArray{<:AbstractArray{T, N}, N}), BS<:Tuple{Vararg{AbstractUnitRange{Int64}, N}}} at ~/.julia/packages/BlockArrays/UKehW/src/blockarray.jl:162
[...]

PR #606, opened 2022-08-16
Update ktypeof function to support BlockArrays and FillArrays

### LINKED ISSUE #573: Compatibility with FillArrays
Hi! First of all, love your package :)
I'm trying to use it within automatic differentiation, and running into issues because some array types are not accepted. Typically, Flux may return FillArrays when all partial derivatives are equal (eg. for sums), but these are not supported by Krylov solvers (whereas dense and sparse arrays are): is there a workaround?
Here's an MWE:
```julia
julia> using FillArrays, Krylov, SparseArrays

julia> A = rand(10, 10);

julia> b_dense = zeros(10);

julia> b_sparse = spzeros(10);

julia> b_fill = Zeros(10);

julia> gmres(A, b_dense);

julia> gmres(A, b_sparse);

julia> gmres(A, b_fill);
ERROR: MethodError: no method matching Zeros{Float64, 1, Tuple{Base.OneTo{Int64}}}(::UndefInitializer, ::Int64)
Stacktrace:
 [1] GmresSolver(n::Int64, m::Int64, memory::Int64, S::Type{Zeros{Float64, 1, Tuple{Base.OneTo{Int64}}}})
   @ Krylov ~/.julia/packages/Krylov/gJboP/src/krylov_solvers.jl:1439
 [2] GmresSolver(A::Matrix{Float64}, b::Zeros{Float64, 1, Tuple{Base.OneTo{Int64}}}, memory::Int64)
   @ Krylov ~/.julia/packages/Krylov/gJboP/src/krylov_solvers.jl:1457
 [3] gmres(A::Matrix{Float64}, b::Zeros{Float64, 1, Tuple{Base.OneTo{Int64}}}; memory::Int64, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
   @ Krylov ~/.julia/packages/Krylov/gJboP/src/gmres.jl:38
 [4] gmres(A::Matrix{Float64}, b::Zeros{Float64, 1, Tuple{Base.OneTo{Int64}}})
   @ Krylov ~/.julia/packages/Krylov/gJboP/src/gmres.jl:38
 [5] top-level scope
   @ REPL[58]:1
```

amontoison:
close #573 
It also concerns #605

### DIFF
--- a/src/krylov_utils.jl
+++ b/src/krylov_utils.jl
@@ -180,13 +180,17 @@ end
 
 Return a dense storage type `S` based on the type of `v`.
 """
-function ktypeof(v :: AbstractVector)
+function ktypeof(v :: AbstractVector{T}) where T
   S = typeof(v)
   if S <: SubArray
-    S = S.types[1]  # SubArray
+    S = S.types[1]
   end
-  if S <: AbstractSparseVector
-    S = S <: SparseVector ? S.types[3] : S.types[2]  # SparseVector / CuSparseVector
+  if !(S <: DenseVector)
+    if S <: AbstractSparseVector
+      S = S <: SparseVector ? S.types[3] : S.types[2]  # SparseVector / CuSparseVector
+    else
+      S = Vector{T}
+    end
   end
   return S
 end


gdalle:
Hi there, and thanks for fixing #573! I just took a look at the code, and I was wondering why the function `ktypeof` doesn't exploit multiple dispatch? My first thought would have been something like this:
```julia
function ktypeof(v::S) where {S<:SubArray}
    return S.types[1]
end

function ktypeof(v::S) where {T,S<:DenseVector{T}}
    return Vector{T}
end

function ktypeof(v::S) where {S<:SparseVector}
    return S.types[3]
end

function ktypeof(v::S) where {S<:AbstractSparseVector}
    return S.types[2]
end

function ktypeof(v::S) where {S<:AbstractVector}
    return S
end
```
Would that work too?

PR #610, opened 2022-08-19
Replace mul! by mulorldiv! for preconditioners

amontoison:
We should be able to switch between `mul!̀` and `ldiv!` if we add a boolean argument to all Krylov methods that support preconditioners. This parameter will be the last argument of `kmuldiv!` function.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -20,7 +20,7 @@ makedocs(
            "In-place methods" => "inplace.md",
            "GPU support" => "gpu.md",
            "Warm start" => "warm_start.md",
-           "Factorization-free operators" => "factorization-free.md",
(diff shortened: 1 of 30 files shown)


docs/make.jl
```diff
@@ -20,7 +20,7 @@ makedocs(
            "In-place methods" => "inplace.md",
            "GPU support" => "gpu.md",
            "Warm start" => "warm_start.md",
-           "Factorization-free operators" => "factorization-free.md",
+           "Matrix-free operators" => "matrix-free.md",
```
dpo, line 23: I think factorization-free makes more sense here. You can argue that there's always a matrix somewhere.

dpo:
Great! Thank you! It would be great to document the `ldiv` argument.

PR #620, opened 2022-09-06
Update CommentPR.yml

JarLob:
The expression was treated as a string and always evaluated as true.

### DIFF
--- a/.github/workflows/CommentPR.yml
+++ b/.github/workflows/CommentPR.yml
@@ -13,8 +13,8 @@ jobs:
   upload:
     runs-on: ubuntu-latest
     if: >
-      ${{ github.event.workflow_run.event == 'pull_request' &&
-      github.event.workflow_run.conclusion == 'success' }}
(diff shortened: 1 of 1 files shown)


amontoison:
Thanks @JarLob!

amontoison:
@JarLob 
Do you know if it's easy to update `CommentPR.yml` such that the same message is edited if new commits are added?
I would like this kind of situation: https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/625

JarLob:
I don't know which way is the easiest, but I would probably look for an existing comment with `octokit.rest.issues.listComments` (pull request is an issue with code).

amontoison:
Thanks for your help @JarLob, with the new `CommentPR.yml`, the comment of GHA[bot] is updated.
Examples: https://github.com/amontoison/Krylov.jl/pull/2#issuecomment-1242637400

JarLob:
Glad I could help!

PR #624, opened 2022-09-07
[KrylovSolvers] inner constructors -> outer constructors

amontoison:
@paraynaud 
It should solve our issue with `CgSolver`.

### DIFF
--- a/src/krylov_solvers.jl
+++ b/src/krylov_solvers.jl
@@ -68,29 +68,29 @@ mutable struct MinresSolver{T,FC,S} <: KrylovSolver{T,FC,S}
   err_vec    :: Vector{T}
   warm_start :: Bool
   stats      :: SimpleStats{T}
+end
 
-  function MinresSolver(n, m, S; window :: Int=5)
-    FC = eltype(S)
-    T  = real(FC)
-    Δx = S(undef, 0)
-    x  = S(undef, n)
-    r1 = S(undef, n)
-    r2 = S(undef, n)
-    w1 = S(undef, n)
-    w2 = S(undef, n)
-    y  = S(undef, n)
-    v  = S(undef, 0)
-    err_vec = zeros(T, window)
-    stats = SimpleStats(0, false, false, T[], T[], T[], "unknown")
-    solver = new{T,FC,S}(Δx, x, r1, r2, w1, w2, y, v, err_vec, false, stats)
-    return solver
-  end
+function MinresSolver(n, m, S; window :: Int=5)
+  FC = eltype(S)
+  T  = real(FC)
+  Δx = S(undef, 0)
+  x  = S(undef, n)
+  r1 = S(undef, n)
+  r2 = S(undef, n)
+  w1 = S(undef, n)
+  w2 = S(undef, n)
+  y  = S(undef, n)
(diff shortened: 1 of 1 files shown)


paraynaud:
Yesterday, I worked on a forked version of Krylov.jl and I defined my own `CgSolver` from `PartitionedVector` (after defining the constructors out the `mutable struct CgSolver`).
Now I am trying to make `cg` works on these `PartitionedVector`s.

ChrisRackauckas:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blame/main/src/krylov_solvers.jl#L1548

Most array types do not have an undef initializer. Is there a reason this was used/assumed? Is the plan to no longer support generic array types in this package?

amontoison:
> https://github.com/JuliaSmoothOptimizers/Krylov.jl/blame/main/src/krylov_solvers.jl#L1548
> 
> Most array types do not have an undef initializer. Is there a reason this was used/assumed?

I use `undef` because I don't want to initialize vectors when I create the workspace.
The vectors of a `KrylovSolver` are initialized in the associated Krylov method.
It's the same behaviour if I just create a `KrylovSolver` is if I want to reuse the `KrylovSolver`.

I use the type of `b` to determine the best type of `DenseVector` that we will use to store the vectors of the workspace.
It's the function [ktypeof](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/krylov_utils.jl#L196-L223).
I don't want to use `typeof(b)` because the methods are not efficient if:
- `b` is stored in a subtype of an `AbstractSparseVector`;
- `b` is wrapped in a `Transpose` or `Adjoint`;
- `b` is a view of a vector or a matrix.
It's also for these reasons that I can't use `similar(b, n)`.

It's working well for `Vector`, `CuVector`, `ROCVector` and `oneVector` but I don't see any issue to use another default constructor if it's more generic.
 
The only kind of `AbstractVector` that I'm aware it's not working is `GBVector`.
But `SuiteSparseGraphBLAS.jl` is efficient for sparse matrix - (sparse matrix / sparse vector) products)
and I don't see improvements for sparse matrix - dense vector products. (@Wimmerer can you confirm?)

> Is the plan to no longer support generic array types in this package?

I don't understand your comment... `S(undef, n)` was always used in this package.
The goal of this PR was to transform inner constructors into outer constructors, which allow users to define their own constructors for array types that don't support `undef initializer` or want to use their own initializer for instance.
It's more generic than before with this modification.

P.S. [Bug reports and discussions](https://juliasmoothoptimizers.github.io/Krylov.jl/dev/#Bug-reports-and-discussions)

Wimmerer:
I'm not sure I have the full context here, but for 5000x5000 0.01 sparsity matrix I get:

- SparseMatrixCSC * dense vector: 127 microseconds
- GBMatrix (held by column): 108 microseconds
- GBMatrix (held by row): 70 microseconds

I suspect as the matrix gets much larger we would see a much more significant difference as well (since SuiteSparseGraphBLAS is fully threaded). I will be updating the docs with some performance hints with the next update (the next update will also do quite a few things better, but performance will be about the same if you're hitting the right code paths).

PR #625, opened 2022-09-07
Use buildkite to test the GPU support

amontoison:
:crossed_fingers:

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,65 @@
+steps:
+  - label: "Nvidia GPUs -- CUDA.jl"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: 1.8
+    agents:
+      queue: "juliagpu"
+      cuda: "*"
+    command: |
+      julia --color=yes --project -e '
+      using Pkg
+      Pkg.add("CUDA")
+      Pkg.instantiate()
+      include("test/gpu/nvidia.jl")'
+    timeout_in_minutes: 30
+
+  - label: "AMD GPUs -- AMDGPU.jl"
+    plugins:
+      - JuliaCI/julia#v1:
+          version: 1.8
+    agents:
+      queue: "juliagpu"
+      rocm: "*"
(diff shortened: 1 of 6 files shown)


amontoison:
Krylov.jl works with all GPUs supported by Julia :tada: 

We just need to change some BLAS1 functions.
My main issue is to find a generic implementation for the dot product...
The generic implementation `reduce(+, map((xᵢ,yᵢ) -> conj(xᵢ) * yᵢ, x, y))` allocates a vector to store `conj(x) .* y` and we want to avoid it.
For the norm we can do `sqrt(mapreduce(abs2, +, x))`, which is exactly what we need.

dpo:
> My main issue is to find a generic implementation for the dot product...
The generic implementation reduce(+, map((xᵢ,yᵢ) -> conj(xᵢ) * yᵢ, x, y)) allocates a vector to store conj(x) .* y and we want to avoid it.

Why is that necessary? Do `x'y` and `dot(x, y)` not work on GPU?

> For the norm we can do sqrt(mapreduce(abs2, +, x)), which is exactly what we need.

It's unfortunately a simplistic implementation of the norm though, that's more prone to overflow. A typical implementation first scaled x by `maximum(abs.(x))`, then sums the squares (ideally from smallest to largest in absolute value), then scales back.

amontoison:
`x'y` calls `dot(x, y)` and it's not always implemented. `CUDA.jl` and `AMDGPU.jl` have their own BLAS but the BLAS of `oneAPI.jl` (oneMKL) is not interfaced.

amontoison:
@dpo `norm`, `axpy!` and `axpby!` have already generic implementations for GPUs:
https://github.com/JuliaGPU/GPUArrays.jl/blob/master/src/host/linalg.jl#L376-L408
https://github.com/JuliaGPU/GPUArrays.jl/blob/master/src/host/linalg.jl#L447-L450

The macro `@code_warntype` is very useful to find which routine is called!
We just need to add `dot` in [GPUArrays.jl](https://github.com/JuliaGPU/GPUArrays.jl).

PR #628, opened 2022-09-09
Update workflows

amontoison:


amontoison:
Let's see if the comment for breakages is updated :crossed_fingers:

tmigot:
Why did you remove the CompatHelper?

amontoison:
It was a mistake @tmigot.

This PR needs to be merged for the changes take effect since GHA only considers workflow file on the default branch for `workflow_run` event.

PR #645, opened 2022-09-29
Support PartitionedVectors in `to_boundary`

paraynaud:
@amontoison, I replaced `Vector{T}` by `AbstractVector{T}` to support `PartitionedVectors`.
What do you think?


paraynaud:
Nevermind, I just saw the conflict in `krylov_utils.jl`, the current `to_boundary` should work!

Issue #564, opened 2022-04-24
Warm start


amontoison:
### Krylov methods

- [ ] CG
- [ ] CR
- [ ] SYMMLQ
- [ ] CG-LANCZOS without shifts
- [ ] CG-LANCZOS with shifts
- [ ] MINRES (#562)
- [ ] MINRES-QLP
- [ ] DQGMRES
- [ ] DIOM
- [ ] GMRES
- [ ] FOM
- [ ] USYMLQ
- [ ] USYMQR
- [ ] TRICG
- [ ] TRIMR
- [ ] TRILQR
- [ ] CGS
- [ ] BICGSTAB
- [ ] BILQ
- [ ] QMR
- [ ] BILQR
- [ ] CGLS
- [ ] CRLS
- [ ] CGNE
- [ ] CRMR
- [ ] LSLQ
- [ ] LSQR
- [ ] LSMR
- [ ] LNLQ
- [ ] CRAIG
- [ ] CRAIGMR
- [ ] GPMR

amontoison:
CG-LANCZOS-SHIFT can't be warm-started because if we want to solve `(A + λᵢI)xᵢ = b` with `xᵢ = x₀ + Δxᵢ` then we will need a different right-hand side for each `λᵢ`:
```julia
(A + λᵢI)Δxᵢ = b - (A + λᵢI)x₀
```

PR #647, opened 2022-10-03
add imput and output arguments in all docstrings

amontoison:
I will add a `#### Optional argument(s)` section when #564 will be closed.
The final goal is to also add `#### Keyword arguments` and `#### Stopping conditions` in all docstrings such that we can do a release 1.0.

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -24,7 +24,7 @@ export bicgstab, bicgstab!
 `T` is an `AbstractFloat` such as `Float32`, `Float64` or `BigFloat`.
 `FC` is `T` or `Complex{T}`.
 
-Solve the square linear system Ax = b using BICGSTAB.
+Solve the square linear system Ax = b of size n using BICGSTAB.
 BICGSTAB requires two initial vectors `b` and `c`.
 The relation `bᴴc ≠ 0` must be satisfied and by default `c = b`.
 
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -53,12 +53,12 @@ and `false` otherwise.
 
 #### Input arguments
 
-* `A`: a linear operator that models a matrix of dimension n.
+* `A`: a linear operator that models a matrix of dimension n;
 * `b`: a vector of length n.
 
 #### Output arguments
 
-* `x`: a dense vector of length n.
+* `x`: a dense vector of length n;
 * `stats`: statistics collected on the run in a [`SimpleStats`](@ref) structure.
 
 #### References
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -51,6 +51,16 @@ where `kwargs` are the same keyword arguments as above.
 The callback is called as `callback(solver)` and should return `true` if the main loop should terminate,
 and `false` otherwise.
 
+#### Input arguments
+
+* `A`: a linear operator that models a matrix of dimension n.
+* `b`: a vector of length n.
+
+#### Output arguments
+
+* `x`: a dense vector of length n.
+* `stats`: statistics collected on the run in a [`SimpleStats`](@ref) structure.
+
 #### References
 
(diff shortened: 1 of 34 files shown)


src/bicgstab.jl
```diff
@@ -51,6 +51,16 @@ where `kwargs` are the same keyword arguments as above.
 The callback is called as `callback(solver)` and should return `true` if the main loop should terminate,
 and `false` otherwise.
 
+#### Input arguments
+
+* `A`: a linear operator that models a matrix of dimension n.
```
dpo, line 56:
```suggestion
* `A`: a linear operator that models a matrix of dimension n;
```

dpo: I don't see where n is defined.

src/bicgstab.jl
```diff
@@ -51,6 +51,16 @@ where `kwargs` are the same keyword arguments as above.
...
+* `b`: a vector of length n.
+
+#### Output arguments
+
+* `x`: a dense vector of length n.
```
dpo, line 61:
```suggestion
* `x`: a dense vector of length n;
```

src/bicgstab.jl
amontoison: `n` is not a parameter but I use it to say that the solution is of length `n` in the docstring.
When we have multiple solution, which is the case for `TriCG`, `TriMR` or `GPMR`, it helps to give the length of output vectors.

dpo: I know but n doesn't seem to be defined in the docstring. It should say somewhere that Ax=b is a system of size n.

dpo:
> Is there a reason why you use `####` instead of just `#`?

The font size would be obnoxiously large in the docs.

src/cgls.jl
```diff
@@ -42,7 +42,7 @@ Solve the regularized linear least-squares problem
 
     minimize ‖b - Ax‖₂² + λ‖x‖₂²
 
-using the Conjugate Gradient (CG) method, where λ ≥ 0 is a regularization
+of size n × m using the Conjugate Gradient (CG) method, where λ ≥ 0 is a regularization
```
dpo, line 45: m x n

src/gpmr.jl
```diff
@@ -27,7 +27,7 @@ GPMR solves the unsymmetric partitioned linear system
     [ λI   A ] [ x ] = [ b ]
     [  B  μI ] [ y ]   [ c ],
 
-where λ and μ are real or complex numbers.
+of size (n+m) × (n+m) where λ and μ are real or complex numbers.
```
dpo, line 30: give the size of A and B instead

dpo (changes requested):

src/cgls.jl
amontoison: Do you want to use `m x n` when `A` has more rows than columns or when A is rectangular?

dpo: You grab `m, n` as the size of A : https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/647/files/a2e5796af0bdee8d225347cbd9f984fd554a72c2#diff-695bcdad4090fc22200b17203cb5318c0aa0f12e11a6ecdc048ea445ce5a22f1R96

But yes, a matrix is always m x n :-)

src/gpmr.jl
amontoison: What do you think of:
```
Given matrices `A` of dimension m × n and `B` of dimension n × m,
GPMR solves the unsymmetric partitioned linear system

    [ λIₘ   A  ] [ x ] = [ b ]
    [  B   μIₙ ] [ y ]   [ c ],

of size (n+m) × (n+m) where λ and μ are real or complex numbers.
```
and
```
Given a matrix `A` of dimension m × n, TriCG / TriMR solves the symmetric linear system

    [ τE    A ] [ x ] = [ b ]
    [  Aᴴ  νF ] [ y ]   [ c ],

of size (n+m) × (n+m) where τ and ν are real numbers, E = M⁻¹ ≻ 0, F = N⁻¹ ≻ 0.
```
?

Issue #646, opened 2022-10-02
Use `Real` as FC instead of `AbstractFloat` ?


j-fu:
First, thank you for this package !

I am currently trying out its possibilities. 
This includes differentiating through the calculations. In the type specification you always seem to require that FC is an `AbstractFloat`. However, this excludes `ForwardDiff.Dual` which is passed when performing forward mode autodiff. Their common supertype is `Real`. 

So my question: Would it be possible replace `AbstractFloat` by `Real` ?

amontoison:
Hi @j-fu!
Can you give a small example of a linear system where you use `ForwardDiff.Dual`?
I use `AbstractFloat` instead of `Real` to exclude `Integer` numbers but I don't know that  `ForwardDiff.Dual` is a subtype of `Real`.

j-fu:
Hi, thanks for consideration, here is a simple example:
```
using LinearAlgebra
using ForwardDiff
using DiffResults
using Krylov

function makematrix(n,X)
    m=zeros(eltype(X),n,n)
    for i=1:n
        m[i,i]=10+X[1]^2
    end
    
    for i=1:n-2
        m[i,i+2]=-X[2]^2
        m[i+2,i]=-X[2]^2
    end
    m
end

function evalderiv(f,X)
    println("    direct evaluation:  f(X)=$(f(X))")
    
    dresult=DiffResults.JacobianResult(ones(1),X)
    ForwardDiff.jacobian!(dresult,f,X)
    println("ForwarDiff evaluation:  f(X)=$(DiffResults.value(dresult))")
    println(" ForwardDiff Jacobian: df(X)=$(DiffResults.jacobian(dresult))")
end

function testfwd(X=[1.0,0.1];n=20)

    fdirect(Y)=[sum(makematrix(n,Y)\ones(eltype(Y),n))]
    println("direct solver:")
    evalderiv(fdirect,X)

    println("cg iteration:")
    fcg(Y)=[sum( cg(makematrix(n,Y),ones(eltype(Y),n))[1])]
    evalderiv(fcg,X)
    
end
```
May be a bit contrived but it makes the point of matrix entries depending on parameters in a nonlinear way. 

Generally I aim at differentiating through nonlinear parameter dependent PDE solves to do some interesting stuff (parameter ID, Bayes, Bifurcation). 
[...]

PR #648, opened 2022-10-04
Use Real instead of AbstractFloat

amontoison:
#646 
@j-fu


amontoison:
@dpo @tmigot 
I didn't replace `FC` by `RC` because of the "diff" size but I can open another PR with just this modifcation.
I also think that we should add a inner constructor in each of our Krylov solvers to exclude integer types and throw an explicit error message.

dpo:
Is it really necessary to distinguish between "real" and "complex"? Why not `T <: Number`? Krylov methods should apply to any arithmetic type that supports +, -, *, / and sqrt.

amontoison:
> Is it really necessary to distinguish between "real" and "complex"?
Am I not sure to understand your question, are you asking if we could use one type instead of `T` and `FC`?
Try to use `eps` with a complex number.

> Why not `T <: Number`? Krylov methods should apply to any arithmetic type that supports +, -, *, / and sqrt.
Do you mean `FC` ?
We need more than +, -, *, / and sqrt operations in Krylov methods.
I don't have a generic `sym_givens`routine for `T <: Number` for instance.
Do you think that quaternions and complex can use the same routines?

j-fu:
Great that you consider this! I am getting some errors when trying this PR with my example 

https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/646#issuecomment-1265935010

Also, if I add `rtol=1.0e-10` I get the complaint that a value of Type T is expected instead of Float64. 

These errors are the same as in the code where I would like to use it: https://github.com/j-fu/VoronoiFVM.jl/blob/genericfloat/src/vfvm_solver.jl#L217

I am not in a hurry, though it may seem like that, so take your time...

amontoison:
> Great that you consider this! I am getting some errors when trying this PR with my example
> 
> [#646 (comment)](https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/646#issuecomment-1265935010)
> 
> Also, if I add `rtol=1.0e-10` I get the complaint that a value of Type T is expected instead of Float64.
> 
> These errors are the same as in the code where I would like to use it: https://github.com/j-fu/VoronoiFVM.jl/blob/genericfloat/src/vfvm_solver.jl#L217
> 
> I am not in a hurry, though it may seem like that, so take your time...

The issue is the `eps` function. All `AbstractFloat` returns a value of type `T` when we call `eps(T)` but with ForwardDiff.Dual it's not the case.
It should work if you provide `atol=T(eps(T))` and `rtol = T(eps(T))`  where T = `eltype(b)`.

j-fu:
Hi, when I update my MWE accordingly: 
```
function testfwd(X=[1.0,0.1];n=20)

    fdirect(Y)=[sum(makematrix(n,Y)\ones(eltype(Y),n))]
    println("direct solver:")
    evalderiv(fdirect,X)

    println("cg iteration:")
    function fcg(Y)
        T=eltype(Y)
        rtol=sqrt(T(eps(T)))
        A=makematrix(n,Y)
        b=ones(T,n)
        @assert eltype(A)==eltype(b)==typeof(rtol)
        sol=cg(A,b; rtol)
        [sum(sol[1])]
    end
    evalderiv(fcg,X)
end
```
I get another MethodError:
```
direct solver:
    direct evaluation:  f(X)=[9.107136641581038]
ForwarDiff evaluation:  f(X)=[9.107136641581036]
 ForwardDiff Jacobian: df(X)=[-1.658798808811399 0.3251361376331415]
cg iteration:
    direct evaluation:  f(X)=[9.107136641581036]
ERROR: MethodError: no method matching var"#cg!#77"(::UniformScaling{Bool}, ::Float64, ::ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}, ::Int64, ::ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}, ::Bool, ::Int64, ::Bool, ::Bool, ::Krylov.var"#79#81", ::typeof(cg!), ::CgSolver{ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}, ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}, Vector{ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}}}, ::Matrix{ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}}, ::Vector{ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}})
Closest candidates are:
  var"#cg!#77"(::Any, ::T, ::T, ::Int64, ::T, ::Bool, ::Int64, ::Bool, ::Bool, ::Any, ::typeof(cg!), ::CgSolver{T, FC, S}, ::Any, ::AbstractVector{FC}) where {T<:Real, FC<:Union{Complex{T}, T}, S<:DenseVector{FC}} at ~/.julia/packages/Krylov/HvO4O/src/cg.jl:83
Stacktrace:
 [1] cg(A::Matrix{ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}}, b::Vector{ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}}; kwargs::Base.Pairs{Symbol, ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}, Tuple{Symbol}, NamedTuple{(:rtol,), Tuple{ForwardDiff.Dual{ForwardDiff.Tag{var"#fcg#12"{Int64}, Float64}, Float64, 2}}}})
   @ Krylov ~/.julia/packages/Krylov/HvO4O/src/cg.jl:63
...
```

j-fu:
To have rtol (and possibly other parameters) as T in this context feels counterintuitive. I understand that you want to be able e.g. BigFloat. So may be a Union could do here ?

amontoison:
> To have rtol (and possibly other parameters) as T in this context feels counterintuitive. I understand that you want to be able e.g. BigFloat. So may be a Union could do here ?

Yes, I agree with you, but for type-stability it's better to keep it.
From my point of view, we should overload the `eps` of ForwardDiff to also return a `Dual`.

PR #654, opened 2022-10-07
Better display of Krylov solvers

amontoison:
@dpo @tmigot 
I modified the `show` function for the Krylov solvers.
It displays `nrows`, `ncols` AND the `storage` used by the solver!!!
Maybe we should say `unallocated` for the size of empty vectors instead of `0` byte.
What do you think? 

```julia
using SuiteSparseMatrixCollection, MatrixMarket, Krylov

ssmc = ssmc_db()

matrix_name = "pre2"
matrix = ssmc_matrices(ssmc, "", matrix_name)
paths = fetch_ssmc(matrix, format="MM")
M = MatrixMarket.mmread(joinpath(paths[1], matrix_name * ".mtx"))

bilq_solver = BilqSolver(size(M)..., Vector{Float64})
bicgstab_solver = BicgstabSolver(size(M)..., Vector{Float64})

matrix_name = "nlpkkt200"
matrix = ssmc_matrices(ssmc, "", matrix_name)
paths = fetch_ssmc(matrix, format="MM")
M = MatrixMarket.mmread(joinpath(paths[1], matrix_name * ".mtx"))

cg_solver = CgSolver(size(M)..., CuVector{Float64})
minresqlp_solver = MinresQlpSolver(size(M)..., CuVector{Float64})
```
Capture d’écran du 2022-10-07 02-07-58
Capture d’écran du 2022-10-07 02-08-14
Capture d’écran du 2022-10-07 02-12-08
Capture d’écran du 2022-10-07 02-12-24

### DIFF
--- a/src/krylov_solvers.jl
+++ b/src/krylov_solvers.jl
@@ -8,7 +8,7 @@ GmresSolver, FomSolver, GpmrSolver, FgmresSolver
 export solve!, solution, nsolution, statistics, issolved, issolved_primal, issolved_dual,
 niterations, Aprod, Atprod, Bprod, warm_start!
 
-import Base.size
+import Base.size, Base.sizeof
 
 const KRYLOV_SOLVERS = Dict(
   :cg               => :CgSolver            ,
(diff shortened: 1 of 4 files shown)


dpo:
Beautiful! How about something like this for the size: https://github.com/IainNZ/Humanize.jl/blob/master/src/Humanize.jl#L30 or https://discourse.julialang.org/t/how-to-print-human-readable-byte-counts/31877/2 ?

amontoison:
`Base.format_bytes` is exactly what I needed!
Capture d’écran du 2022-10-07 11-30-47
Capture d’écran du 2022-10-07 11-30-56

dpo:
Nice! Some tests are failing though.

amontoison:
The tests are failing because I changed the `show` output and I didn't updated the tests.
I wanted some feedbacks before.

test/test_solvers.jl
```diff
@@ -11,1175 +11,139 @@ function test_solvers(FC)
...
+          @test nsolution(solver) == 1
+        end
+
+        @test niter > 0
+        @test statistics(solver) === solver.stats
+        @test issolved(solver)
+      end
+
+      io = IOBuffer()
+      show(io, solver, show_stats=false)
```
tmigot, line 129:
```suggestion
      show(io, solver, show_stats=true)
```
just to increase coverage :).

tmigot (approved):
Thanks!

test/test_solvers.jl
amontoison: You should look at line 147 ;)

src/krylov_solvers.jl
```diff
@@ -1885,45 +1885,69 @@ for (KS, fun, nsol, nA, nAt, warm_start) in [
   end
 end
 
+function ksizeof(attribute)
+  if isa(attribute, Vector{<:AbstractVector}) && !isempty(attribute)
+    # A vector of vector is a vector of pointers in Julia.
```
amontoison, line 1890:
```suggestion
    # A vector of vectors is a vector of pointers in Julia.
```

amontoison:
Thanks for the feedbacks @tmigot and @dpo!

PR #656, opened 2022-10-10
Add optional argument(s) in the docstrings

amontoison:
Is `⦗...⦘` well displayed on Windows and Mac?
I don't want to use`[]` for optional arguments but I hope that all OS have a font for this unicode character.
https://juliasmoothoptimizers.github.io/Krylov.jl/previews/PR656/solvers/spd/#CG

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -16,13 +16,18 @@
 export bicgstab, bicgstab!
 
 """
-    (x, stats) = bicgstab(A, b::AbstractVector{FC}; c::AbstractVector{FC}=b,
-                          M=I, N=I, atol::T=√eps(T), rtol::T=√eps(T),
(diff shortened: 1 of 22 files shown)


dpo (commented):
It looks weird. Why not use `args...` for positional arguments and describe them in the docs? Or simply leave out "[" and "]".

amontoison:
> It looks weird. Why not use args... for positional arguments and describe them in the docs? Or simply leave out "[" and "]".

I would like to keep positional arguments (`A` and `b`) in the header of the docstring.
What do you think of `optargs...` for positional argument(s) and describe them in the docs?
If I leave out "[" and "]", i'm scared that the user thinks that it's a mandatory argument.

dpo:
Looks good!

PR #662, opened 2022-10-12
[LNLQ] Use utolx and utoly instead of etolx and etoly

amontoison:
I also fixed a bug. `solved_lq` must be always updated if `σₑₛₜ > 0`.


tmigot (approved):
Looks good to me. So `utol` now has the same meaning as in `lslq`. Thanks

PR #666, opened 2022-10-14
[documentation] Add a section storage requirements

amontoison:
https://juliasmoothoptimizers.github.io/Krylov.jl/previews/PR666/storage/

### DIFF
--- a/docs/src/storage.md
+++ b/docs/src/storage.md
@@ -59,7 +59,7 @@ Each table summarizes the storage requirements of Krylov methods recommended to 
 |:-------:|:-----------------------:|:-----------------------:|:-------------------------------:|
 | Storage | $5n$                    | $6n$                    | $6n$                            |
 
-#### Non-Hermitian linear systems
+#### Non-Hermitian square linear systems
 
 | Methods | [`CGS`](@ref cgs) | [`BICGSTAB`](@ref bicgstab) | [`BiLQ`](@ref bilq) | [`QMR`](@ref qmr) |
 |:-------:|:-----------------:|:---------------------------:|:-------------------:|:-----------------:|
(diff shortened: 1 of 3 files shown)


amontoison:
@dpo, It's for the last part of the JOSS paper.

docs/src/storage.md
```diff
@@ -0,0 +1,152 @@
...
+|:-------:|:---------------:|:---------------:|:-------------------------------:|:-------------------------------------------:|
+ Storage  | $4n$            | $5n$            | $5n$                            | $3n + 2np + 5p$                             |
+
+#### Hermitian indefinite linear systems
+
+| Methods | [`SYMMLQ`](@ref symmlq) | [`MINRES`](@ref minres) | [`MINRES-QLP`](@ref minres_qlp) |
+|:-------:|:-----------------------:|:-----------------------:|:-------------------------------:|
+| Storage | $5n$                    | $6n$                    | $6n$                            |
+
+#### Non-Hermitian linear systems
```
dpo, line 62:
```suggestion
#### Non-Hermitian square linear systems
```

dpo (approved):
Looks great! Just a minor comment.

docs/src/storage.md
```diff
@@ -0,0 +1,152 @@
...
+```
+
+Thereafter, we can use `Base.format_bytes(nbytes)` to recover what is displayed in the REPL.
+
+```@example storage
+Base.format_bytes(nbytes)
+```
+
+To verify that we match the theoretical results, we just need to multiply the storage requirement of a method by the number of bytes associated to the precision of the linear problem.
+For instance, we need 4 bytes for the precision `Float32` , 8 bytes for precisions `Float64` and `ComplexF32`, and 16 bytes for the precision `ComplexF64`.
```
amontoison, line 135:
```suggestion
For instance, we need 4 bytes for the precision `Float32`, 8 bytes for precisions `Float64` and `ComplexF32`, and 16 bytes for the precision `ComplexF64`.
```

PR #671, opened 2022-10-19
Improve Krylov processes

amontoison:
- The upper Hessenberg matrices are now stored as dense matrices (`Arnoldi` and `montoison_orban` processes).
It's less expensive in storage for `Float32`, `Float64`, `ComplexF32` and the difference is quite small for `ComplexF64`.

- For the four other processes, I replaced all `M[i,j]` where `M` is the projection of `A` in the Krylov subspace (`T`, `Tᴴ` or `L` variables) by `nzval[k]` where `k` is the index of `M[i,j]` in `M.nzval`.
It will be a little bit faster because Julia won't need to compute the index `k` and check if `M[i,j]` is already allocated. 

**Note:** I don't want to store a sparse matrix `A` with the type `SparseMatrixCSC{T,Int32}` instead of `SparseMatrixCSC{T,Int64}` because `A \ b` is not working for this type.
We must use `Int64` to dispatch to `SuiteSparse` routines.

### DIFF
--- a/src/krylov_processes.jl
+++ b/src/krylov_processes.jl
@@ -44,6 +44,7 @@ function hermitian_lanczos(A, b::AbstractVector{FC}, k::Int) where FC <: FloatOr
   V = M(undef, n, k+1)
   T = SparseMatrixCSC(k+1, k, colptr, rowval, nzval)
 
+  pαᵢ = 1  # Position of αᵢ in the vector `nzval`
   for i = 1:k
     vᵢ = view(V,:,i)
     vᵢ₊₁ = q = view(V,:,i+1)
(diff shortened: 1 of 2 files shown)


dpo:
> The upper Hessenberg matrices are now stored as dense matrices (Arnoldi and montoison_orban processes).
It's less expensive in storage for Float32, Float64, ComplexF32 and the difference is quite small for ComplexF64.

What do you mean? It depends on the matrix size, doesn't it? n² * 32 bits for dense vs (n * (n+3) / 2 - 1) * (32 + 2 * 64) for sparse (assuming 64 bit Ints).

And we can't use the predefined Hessenberg type because it doesn't allow rectangular matrices, is that it?

amontoison:
> What do you mean? It depends on the matrix size, doesn't it? n² * 32 bits for dense vs (n * (n+3) / 2 - 1) * (32 + 2 * 64) for sparse (assuming 64 bit Ints).

- storage for H stored in dense format: k*(k+1) *  nbits_FC
- storage for H stored in sparse format: (k * (k+3) / 2)  * nbits_FC +  (k * (k+3) / 2)  * nbits_I + (k+1) * nbits_I

It's less expensive for`Float32`, `Float64` and `ComplexF32` because `nbits_FC ≤ nbits_I`.

> And we can't use the predefined Hessenberg type because it doesn't allow rectangular matrices, is that it?

I just checked and `UpperHessenberg` works if `A` is not square.
We could use it if you want.

```julia
help?> UpperHessenberg
search: UpperHessenberg

  UpperHessenberg(A::AbstractMatrix)


  Construct an UpperHessenberg view of the matrix A. Entries of A below the first subdiagonal are ignored.

  Efficient algorithms are implemented for H \ b, det(H), and similar.

  See also the hessenberg function to factor any matrix into a similar upper-Hessenberg matrix.

  If F::Hessenberg is the factorization object, the unitary matrix can be accessed with F.Q and the Hessenberg matrix with F.H. When Q
  is extracted, the resulting type is the HessenbergQ object, and may be converted to a regular matrix with convert(Array, _) (or
  Array(_) for short).

  Iterating the decomposition produces the factors F.Q and F.H.
```

`Tridiagonal` is still only working on square matrices.

dpo:
> storage for H stored in dense format: k*(k+1) * nbits_FC

No; it's k² * nbits.

> I just checked and UpperHessenberg works if A is not square.
We could use it if you want.

Is it more efficient than a dense matrix?

amontoison:
> > storage for H stored in dense format: k*(k+1) * nbits_FC
> 
> ~No; it's k² * nbits.~ Ugh, sorry, nevermind. I was reading k * (k+1) / 2.
> 
> > I just checked and UpperHessenberg works if A is not square.
> > We could use it if you want.
> 
> Is it more efficient than a dense matrix?

No, it's just a wrapper like `Symmetric` or `Diagonal`.
It could help to have a better dispatch if the user performs specific operations like `\` or `det`.
I prefer to not wrapped by default, the users can easily wrapped it if they need it.

dpo:
If Hessenberg doesn't use less storage, let's keep your dense format then.

PR #677, opened 2022-11-08
Add a keyword argument iostream for all Krylov methods

amontoison:
https://juliasmoothoptimizers.github.io/Krylov.jl/previews/PR677/

We will be able to store the outputs of verbose mode with this modification.

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -6,6 +6,7 @@
 ```
 
 ```@docs
+Krylov.kstdout
 Krylov.FloatOrComplex
 Krylov.niterations
 Krylov.Aprod
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -20,7 +20,7 @@ export bicgstab, bicgstab!
                           c::AbstractVector{FC}=b, M=I, N=I,
                           atol::T=√eps(T), rtol::T=√eps(T), itmax::Int=0,
                           verbose::Int=0, history::Bool=false,
-                          ldiv::Bool=false, callback=solver->false, iostream::IO=stdout)
+                          ldiv::Bool=false, callback=solver->false, iostream::IO=kstdout)
 
 `T` is an `AbstractFloat` such as `Float32`, `Float64` or `BigFloat`.
 `FC` is `T` or `Complex{T}`.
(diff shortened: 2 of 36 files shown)


amontoison:
I don't understand why I have allocations when I use a sparse matrix and not when I use a dense one. :thinking: 
```julia
using Printf

function krylov(A, b::AbstractVector{T}; verbose::Int=0, iostream::IO=stdout) where {T}
  verbose > 0 && @printf(iostream, "%3d\n", 42)
  return nothing
end

A = rand(10, 10)
b = rand(10)
krylov(A, b)  # warm up
@allocated krylov(A, b)                   # 0 byte
@allocated krylov(A, b, iostream=stdout)  # 32 bytes

A2 = sprand(10, 10, 0.5)
b2 = rand(10)
krylov(A2, b2)  # warm up
@allocated krylov(A2, b2)                   # 48 bytes ???
@allocated krylov(A2, b2, iostream=stdout)  # 32 bytes

const kstdout = stdout

function krylov2(A, b::AbstractVector{T}; verbose::Int=0, iostream::IO=kstdout) where {T}
  verbose > 0 && @printf(iostream, "%3d\n", 42)
  return nothing
end

A = rand(10, 10)
b = rand(10)
krylov2(A, b)  # warm up
@allocated krylov2(A, b)                    # 0 byte
@allocated krylov2(A, b, iostream=kstdout)  # 16 bytes

A2 = sprand(10, 10, 0.5)
b2 = rand(10)
krylov2(A2, b2)  # warm up
@allocated krylov2(A2, b2)                    # 0 byte
@allocated krylov2(A2, b2, iostream=kstdout)  # 16 bytes
```

amontoison:
> What does the `@code_warntype` say? If nothing useful, you may have to look at the low-level translation.

I checked with `@code_warntype` and it says nothing special.
We have some small differences with `@code_llvm`but I don't understand them.

I will open an issue in the repository of Julia,

amontoison:
@dpo 
 I found the problem, `stdout` from `Base` is a global variable and its type is not fixed.
It means that the compiler can only determine the type at runtime.
The type of `stdout` is different if we use `redirect_stdout(mystream)` for instance.

I don't have allocations with `Core.stdout` because its type can't be modified.

dpo:
great that it was not a bug!

PR #699, opened 2023-01-22
Add new tests for GPUs

amontoison:
A new release of `GPUArrays.jl` is available.
My generic fallbacks for `dot` and `reflect!` should work now.
It will solve some issues with Intel / Apple M1 GPUs.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -18,7 +18,7 @@ steps:
   - label: "AMD GPUs -- AMDGPU.jl"
     plugins:
       - JuliaCI/julia#v1:
-          version: 1.8
+          version: 1.9-nightly
     agents:
       queue: "juliagpu"
       rocm: "*"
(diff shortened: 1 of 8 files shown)


amontoison:
The culprit is the stopping condition based on the backward error...
I should have known!
This condition is not reliable at all because it's an upper bound of the backward error.
It's nice for theoretical articles but in practice it doesn't fit well in the implementations.

amontoison:
@jpsamaroo
Do you know how to solve the error with AMDGPU.jl (buildkite checks)?

amontoison:
It seems that Julia 1.9 + master branch of AMDGPU.jl solved the problem with AMD GPUs.
I have the following message but the tests passed:
```
LoadLib(libhsa-amd-aqlprofile64.so) failed: libhsa-amd-aqlprofile64.so: cannot open shared object file: No such file or directory
```

jpsamaroo:
@amontoison good to hear! AMDGPU CI has been shaky recently, the CI machines enter a state where some GPUs just stop working and we need to reboot. They usually show up as `ERROR_OUT_OF_RESOURCES`.

amontoison:
I understand now what is the issue if I encounter a `ERROR_OUT_OF_RESOURCES` with buildkite CI. Thanks!
We will be able to test the RocSPARSE support here as soon as it will be available in AMDGPU.jl.

PR #702, opened 2023-02-02
Add a type for some arguments of KrylovSolvers

amontoison:


amontoison:
Let's not add more restrictions. It's working fine without them and we have errors in this PR.

PR #705, opened 2023-02-09
Allow more generic storage type S

amontoison:
@paraynaud 
You will be able to remove the statement that you vector type is a DenseVector.


amontoison:
Yes, You can still use `DenseVector` if you want but it's not anymore mandatory for Krylov.jl.

PR #707, opened 2023-02-11
Check that the KrylovSolvers are compatible with the dimension of the linear problems

amontoison:
I recently added `m` and `n` as attributes in the KrylovSolvers.
We can check that the KrylovSolvers are compatible with the size of the linear problems now.


dpo (commented):
Great! Maybe add a unit test where the check fails.

PR #716, opened 2023-03-24
Update tests for Intel GPUs

amontoison:
Issue related: https://github.com/JuliaGPU/oneAPI.jl/issues/283


amontoison:
We probably need a new release of `oneAPI_Support_jll`. I added the interfaces in OneAPI.jl but we also need to add the symbols in the shared library `liboneapi_support.so`.

amontoison:
Fixed by https://github.com/JuliaPackaging/Yggdrasil/pull/6447

PR #730, opened 2023-05-03
Add the keyword argument timemax in all Krylov methods

amontoison:
Maybe we should add a `time` attribute in the stats structures :thinking:


src/minres_qlp.jl
```diff
@@ -447,6 +451,7 @@ function minres_qlp!(solver :: MinresQlpSolver{T,FC,S}, A, b :: AbstractVector{F
   zero_resid          && (status = "found approximate zero-residual solution")
   solved              && (status = "solution good enough given atol and rtol")
   user_requested_exit && (status = "user-requested exit")
+  overtimed           && (status = "maximum time limit exceeded")
```
tmigot, line 454:
```suggestion
  overtimed           && (status = "time limit exceeded")
```

tmigot (commented):
I agree with you, there should be a `elapsed_time` in the stats as well.

src/minres_qlp.jl
```diff
@@ -55,6 +55,7 @@ M also indicates the weighted norm in which residuals are measured.
 * `atol`: absolute stopping tolerance based on the residual norm;
 * `rtol`: relative stopping tolerance based on the residual norm;
 * `itmax`: the maximum number of iterations. If `itmax=0`, the default number of iterations is set to `2n`;
+* `timemax`: the maximum time limit in seconds;
```
amontoison, line 58:
```suggestion
* `timemax`: the time limit in seconds;
```

amontoison:
The elapsed time to allocate the workspace is never take into acount with this implementation. We should find a way to mesure that too.

tmigot:
> The elapsed time to allocate the workspace is never take into acount with this implementation. We should find a way to mesure that too.

Something like this ?
```
function minres_qlp(A, b :: AbstractVector{FC}, x0 :: AbstractVector; max_time::Float64 = Inf, kwargs...) where FC <: FloatOrComplex
  start_time = time()
  solver = MinresQlpSolver(A, b)
  el_time = time() - start_time
  minres_qlp!(solver, A, b, x0; max_time = max_time - el_time, kwargs...)
  solver.stats.elapsed_time += el_time
end
```
It is a bit of extra work for each solver, but you are right, it would be more precise.

src/minres_qlp.jl
amontoison: The issue is that I want to be consistent with `itermax`. You use `max_iter` in JSOSolvers but O try to avoid the underscore.

tmigot: Ok, it makes sense.

amontoison:
Tangi, I just added the `timemax` keyword argument for all in-place methods.
I propose to not add more modifications in this PR.

I will add other PRs to:
- Take into account the time to allocate the workspace for non in-place methods;
- Add a `timer` attribute in all `KrylovStats`.

It could also be great to display the elapsed time in the verbose mode but it's not a priority.

src/minres_qlp.jl
```diff
@@ -198,8 +200,10 @@ function minres_qlp!(solver :: MinresQlpSolver{T,FC,S}, A, b :: AbstractVector{F
   tired = iter ≥ itmax
   status = "unknown"
   user_requested_exit = false
+  overtimed = false
+  overtimed = false
```
tmigot, line 204:
```suggestion
  overtimed = false
```

test/test_fom.jl
```diff
@@ -126,13 +126,6 @@
       end
       
       # test callback function
-      solver = FomSolver(A, b)
-      tol = 1.0e-1
-      cb_n2 = TestCallbackN2(A, b, tol = tol)
-      fom!(solver, A, b, restart = true, callback = cb_n2)
-      @test solver.stats.status == "user-requested exit"
-      @test cb_n2(solver)
-
```
tmigot, line 135: Why do you remove this test?

tmigot (commented):
Looks great, thanks @amontoison ! It's gonna be a nice feature. I just have two small comments

tmigot:
I just saw that CIs failed on Windows...

test/test_fom.jl
amontoison: Because `TestCallbackN2` is not working for FOM. We don't update the solution x in FOM, GMRES and GPMR.
I got some random `NaN` so I removed the test and I will add a new callback later.

amontoison:
> I just saw that CIs failed on Windows...

It seems that something is broken on Windows...

jd-foster:
@amontoison I came here to suggest you use `time_ns()` in place of `time()` but it looks like you already had that idea :)

amontoison:
Thanks @jd-foster!
I checked the GitHub repository of Julia, `time_ns()` and `time()` use different `ccall`:
- https://github.com/JuliaLang/julia/blob/master/base/libc.jl#L249
- https://github.com/JuliaLang/julia/blob/master/base/Base.jl#L114

Something is probably not working as intended in the C function `jl_clock_now`.

@tmigot @geoffroyleconte It could be relevant for your packages too.

geoffroyleconte (approved):
I've actually not encountered issues with `time()` on windows with RipQP or RegularizedOptization. My guess is that it may behave like if your two `time()` calls are almost simultaneous, i.e. your function is very fast to execute. Maybe you should report this if you can reproduce the error (I can test it on windows if needed).

amontoison:
`time()` calls a C function with `ccall` so it's a platform dependent function. Could you test this function on Windows?
```julia
function debug_time()
  start_time = time()
  for i = 1:10
    timer = time() - start_time
    (timer == 0) && println("problem -- $i")
  end
end
```

geoffroyleconte:
```julia
julia> function debug_time()
         start_time = time()
         overtimed = false
         i = 0
         while !overtimed && i ≤ 100000
           i = i + 1
           timer = time() - start_time
           overtimed = timer > 0
         end
         return i
       end
debug_time (generic function with 1 method)

julia> debug_time()
19425

julia> debug_time()
3043

julia> debug_time()
35180

julia> debug_time()
6292
```

amontoison:
Thanks Geoffroy!
You were right.
Could run these functions such that we can determine the sensibility of `time()` on Windows?
```julia
function debug_time()
  start_time = time()
  overtimed = false
  i = 0
  timer = 0.0
  while !overtimed && i ≤ 100000
  	i = i + 1
    timer = time() - start_time
    overtimed = timer > 0
  end
  return i, timer * 1e9
end

debug_time()

function debug_time_ns()
  start_time = time_ns()
  overtimed = false
  i = 0
  timer = 0.0
  while !overtimed && i ≤ 100000
  	i = i + 1
    timer = time_ns() - start_time
    overtimed = timer > 0
  end
  return i, timer
end

debug_time_ns()
```

geoffroyleconte:
```julia

julia> function debug_time()
         start_time = time()
         overtimed = false
         i = 0
         timer = 0.0
         while !overtimed && i ≤ 100000
           i = i + 1
           timer = time() - start_time
           overtimed = timer > 0
         end
           return i, timer
       end
debug_time (generic function with 1 method)

julia> debug_time()
(29166, 0.0009999275207519531)

julia> debug_time()
(16320, 0.002000093460083008)

julia> debug_time()
(12206, 0.0009999275207519531)

julia> function debug_time_ns()
         start_time = time_ns()
           overtimed = false
         i = 0
         timer = 0
         while !overtimed && i ≤ 100000
           i = i + 1
           timer = time_ns() - start_time
           overtimed = timer > 0
         end
         return i, Float64(timer)/1.0e9
       end
debug_time_ns (generic function with 1 method)

julia> debug_time_ns()
(1, 3.0e-7)

julia> debug_time_ns()
(2, 1.0e-7)

julia> debug_time_ns()
(1, 3.0e-7)

julia> debug_time_ns()
(1, 1.0e-7)
```

I switched to seconds. Could you share your results as well?

amontoison:
```julia
julia> debug_time()
(1, 9.5367431640625e-7)

julia> debug_time()
(4, 9.5367431640625e-7)

julia> debug_time()
(10, 9.5367431640625e-7)

julia> debug_time()
(14, 1.1920928955078125e-6)

julia> debug_time()
(18, 9.5367431640625e-7)
```
```julia
julia> debug_time_ns()
(1, 7.9e-8)

julia> debug_time_ns()
(1, 8.7e-8)

julia> debug_time_ns()
(1, 7.9e-8)

julia> debug_time_ns()
(1, 5.3e-8)

julia> debug_time_ns()
(1, 5.5e-8)

julia> debug_time_ns()
(1, 5.1e-8)
```

geoffroyleconte:
```julia
julia> function debug_time_acc()
         start_time = time()
         start_time_ns = time_ns()
         overtimed = false
         i = 0
         timer = 0.0
         timer_ns = 0
         while !overtimed && i ≤ 100000
           i = i + 1
           timer = time() - start_time
           timer_ns = time_ns() - start_time_ns
           overtimed = timer > 0
         end
         time_ns_sec = Float64(timer_ns)/1.0e9
           return i, timer, time_ns_sec, abs(time_ns_sec - timer)
       end
debug_time_acc (generic function with 1 method)

julia> function debug_time_acc()
         start_time = time()
julia> function debug_time_acc()
         start_time = time()
julia> debug_time_acc()
(2538, 0.0009999275207519531, 0.000141, 0.0008589275207519531)

julia> debug_time_acc()
(3187, 0.0009999275207519531, 0.0004378, 0.0005621275207519531)

julia> debug_time_acc()
(15325, 0.0009999275207519531, 0.0008505, 0.0001494275207519531)

julia> debug_time_acc()
(5008, 0.0009999275207519531, 0.0006465, 0.0003534275207519531)
```
Looks pretty inaccurate below the millisecond.

amontoison:
I will keep `time_ns`, `time` only has different values every millisecond on Windows.

PR #735, opened 2023-05-09
Use metaprogramming to define the kwargs of all Krylov methods

amontoison:
Thanks @mosullivan93 for your help with metaprogramming!
@dpo @tmigot Julia will not recompile the non in-place methods anymore if we use different keyword arguments.

It will be also easier to maintain.

ps: I found some mistakes in the docstrings of `CRLS`, `CGLS`, `CGNE` and `CRMR`.

### DIFF
--- a/docs/src/reference.md
+++ b/docs/src/reference.md
@@ -6,10 +6,11 @@
 ```
 
 ```@docs
-Krylov.kstdout
 Krylov.FloatOrComplex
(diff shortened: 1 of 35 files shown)


tmigot (approved):
Thanks @amontoison !
It looks good to me. I guess this will increase the pre-compilation of the package though, but that's alright for now.

PR #739, opened 2023-05-10
Display a warning when the operator and the right-hand side don't have the same type

### LINKED ISSUE #736: Unnecessary `eltype` restriciton on A
it should not matter if `eltype(A) == eltype(b)`. What should matter is that the multiplication (with `*` or `mul!`) produce an object of the same `eltype` as `b`.

I understand the reason for this restriction is to avoid unseen allocations due to type conversions.
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/cec1dd814292a012080b63c30665e89f839a04bc/src/fgmres.jl#L126
However, the assertion makes sense as a constraint on subtypes of `AbstractMatrix` where multiplication of say a `ComplexF64` matrix with a `Float64` vector will not be a Float64 vector.
```julia
julia> ones(ComplexF64, 4, 4) * ones(Float64, 4) |> eltype
ComplexF64 (alias for Complex{Float64})
```
However, the constraint isn't symmetric: multiplying a real matrix with a complex vector results in a complex vector without any allocations.
```julia
julia> ones(Float64, 4, 4) * ones(ComplexF64, 4) |> eltype
ComplexF64 (alias for Complex{Float64})

julia> using LinearAlgebra

julia> A = ones(Float64, 4, 4); b=ones(ComplexF64, 4); x=copy(b);

julia> @allocated mul!(x, A, b)
0
```

I encountered this issue today while writing a PDE operator using the FFT. My `A` matrix-free (written using `SciMLOperators`) and accepts/returns `Vector{Float64}` types. However, as `A` uses the FFT internally, its `eltype` must be `ComplexF64`.

I therefore argue that this constraint is unnecessarily restrictive, especially for matrix-free operators, and should be replaced with a `warning """eltype(A) \neq eltype(b). This could lead to unseen allocations in operator application. """` and a `Tricks.static_hasmethod(LinearAlgebra.mul!, typeof((x, A, b)))`. Additionally you could include a conditional to promote types for `AbstractMatrices`

```julia
if isa(A, AbstractMatrix) & (eltype(A) \neq eltype(b))
    Ta = eltype(A)
    Tb = eltype(b)
    T = promote_type(Ta, Tb)
    Ta !== T && A = T.(A)
    Tb !== T && b = T.(b)
end
```

Please let me know what you think. Happy to follow this up with a PR. :D

amontoison:
close #736

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -141,7 +141,7 @@ kwargs_bicgstab = (:c, :M, :N, :ldiv, :atol, :rtol, :itmax, :timemax, :verbose, 
     NisI = (N === I)
 
     # Check type consistency
-    eltype(A) == FC || error("eltype(A) ≠ $FC")
+    eltype(A) == FC || @warn "eltype(A) ≠ $FC. This could lead to errors or additional allocations in operator-vector products."
     ktypeof(b) <: S || error("ktypeof(b) is not a subtype of $S")
     ktypeof(c) <: S || error("ktypeof(c) is not a subtype of $S")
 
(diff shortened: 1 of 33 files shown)


tmigot (approved):
LGTM, thanks

PR #743, opened 2023-05-11
Update timemax when a KrylovSolver is allocated

amontoison:
Take into account the time to allocate the KrylovSolver and perform warm-start when the timemax option is used.

### DIFF
--- a/src/bicgstab.jl
+++ b/src/bicgstab.jl
@@ -107,20 +107,26 @@ kwargs_bicgstab = (:c, :M, :N, :ldiv, :atol, :rtol, :itmax, :timemax, :verbose, 
 
 @eval begin
   function bicgstab(A, b :: AbstractVector{FC}, x0 :: AbstractVector; $(def_kwargs_bicgstab...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}}
+    start_time = time_ns()
     solver = BicgstabSolver(A, b)
     warm_start!(solver, x0)
+    timemax -= (time_ns() - start_time) / 1e9
     bicgstab!(solver, A, b; $(kwargs_bicgstab...))
     return (solver.x, solver.stats)
   end
 
   function bicgstab(A, b :: AbstractVector{FC}; $(def_kwargs_bicgstab...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}}
+    start_time = time_ns()
     solver = BicgstabSolver(A, b)
+    timemax -= (time_ns() - start_time) / 1e9
     bicgstab!(solver, A, b; $(kwargs_bicgstab...))
     return (solver.x, solver.stats)
   end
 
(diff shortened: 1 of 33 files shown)


src/bicgstab.jl
```diff
@@ -107,20 +107,26 @@ kwargs_bicgstab = (:c, :M, :N, :ldiv, :atol, :rtol, :itmax, :timemax, :verbose,
 
 @eval begin
   function bicgstab(A, b :: AbstractVector{FC}, x0 :: AbstractVector; $(def_kwargs_bicgstab...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}}
+    start_time = time_ns()
     solver = BicgstabSolver(A, b)
     warm_start!(solver, x0)
+    timemax -= (time_ns() - start_time) / 1e9
     bicgstab!(solver, A, b; $(kwargs_bicgstab...))
```
tmigot, line 114: That's quite implicit... Did you make sure that the updated `timemax` has been passed to the solver?

src/bicgstab.jl
```diff
@@ -107,20 +107,26 @@ kwargs_bicgstab = (:c, :M, :N, :ldiv, :atol, :rtol, :itmax, :timemax, :verbose,
...
   function bicgstab(A, b :: AbstractVector{FC}; $(def_kwargs_bicgstab...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}}
+    start_time = time_ns()
     solver = BicgstabSolver(A, b)
+    timemax -= (time_ns() - start_time) / 1e9
     bicgstab!(solver, A, b; $(kwargs_bicgstab...))
     return (solver.x, solver.stats)
   end
 
   function bicgstab!(solver :: BicgstabSolver{T,FC,S}, A, b :: AbstractVector{FC}, x0 :: AbstractVector; $(def_kwargs_bicgstab...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: AbstractVector{FC}}
+    start_time = time_ns()
```
tmigot, line 127: Any idea why some lines like this one are not covered by the tests?

amontoison: Because the warm-start with in-place methods are not tested.

src/bicgstab.jl
amontoison: > That's quite implicit...
 
I don't understand your comment.

> Did you make sure that the updated `timemax` has been passed to the solver?

`timemax` is always a keyword argument so I didn't checked but I can quickly verify if you want.

src/bicgstab.jl
tmigot: Ok, something else to be added to the todo :)

amontoison: I probably reduce the code coverage with this [PR](https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/742) but I prefer to limit the number of intermediate calls.

PR #744, opened 2023-05-12
Update the verbose mode to display a timer

amontoison:
The last step is to add a field `timer` in all KrylovStats.


tmigot (approved):
LGTM, thanks

PR #745, opened 2023-05-12
Add a field timer in all KrylovStats

### LINKED ISSUE #661: Add stopping criterion on time limit
We could check at the end of each iteration wether or not we respect a `max_time::Float64`

amontoison:
close #661


tmigot (approved):
Look good, thanks

PR #752, opened 2023-06-28
Optimize the generic function solve!

### LINKED ISSUE #751: Optimize the function solve!
Our implementation of [`solve!`](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/krylov_solvers.jl#L1835) should not by defined by `args...` and `kwargs...`.

amontoison:
close #751 
I also improved the code coverage for in-place methods with keyword arguments.


tmigot (approved):
LGTM, thanks @amontoison !

PR #767, opened 2023-08-08
Support various array types in extensions

### LINKED ISSUE #766: StaticArrays support
At the moment, the solvers fail on StaticArrays

```julia
julia> using Krylov, StaticArrays

julia> A = rand(2, 2)
2×2 Matrix{Float64}:
 0.175316  0.379652
 0.908308  0.655592

julia> b = SVector{2}(rand(2))
2-element SVector{2, Float64} with indices SOneTo(2):
 0.4292733187710226
 0.6223773606304001

julia> A \ b
2-element Vector{Float64}:
 -0.19634658574195155
  1.2213703847547805

julia> gmres(A, b)
ERROR: MethodError: Cannot `convert` an object of type UndefInitializer to an object of type Float64

Closest candidates are:
  convert(::Type{T}, ::T) where T<:Number
   @ Base number.jl:6
  convert(::Type{T}, ::Number) where T<:Number
   @ Base number.jl:7
  convert(::Type{T}, ::Base.TwicePrecision) where T<:Number
   @ Base twiceprecision.jl:273
  ...

Stacktrace:
  [1] macro expansion
    @ ~/.julia/packages/StaticArraysCore/P6PH7/src/StaticArraysCore.jl:88 [inlined]
  [2] convert_ntuple
    @ ~/.julia/packages/StaticArraysCore/P6PH7/src/StaticArraysCore.jl:84 [inlined]
  [3] SVector{2, Float64}(x::Tuple{UndefInitializer, Int64})
    @ StaticArraysCore ~/.julia/packages/StaticArraysCore/P6PH7/src/StaticArraysCore.jl:120
  [4] SVector{2, Float64}(::UndefInitializer, ::Vararg{Any})
    @ StaticArrays ~/.julia/packages/StaticArrays/9KYrc/src/convert.jl:165
  [5] GmresSolver(m::Int64, n::Int64, memory::Int64, S::Type{SVector{2, Float64}})
    @ Krylov ~/.julia/packages/Krylov/Iuavd/src/krylov_solvers.jl:1549
  [6] GmresSolver(A::Matrix{Float64}, b::SVector{2, Float64}, memory::Int64)
    @ Krylov ~/.julia/packages/Krylov/Iuavd/src/krylov_solvers.jl:1567
  [7] gmres(A::Matrix{Float64}, b::SVector{2, Float64}; memory::Int64, M::LinearAlgebra.UniformScaling{Bool}, N::LinearAlgebra.UniformScaling{Bool}, ldiv::Bool, restart::Bool, reorthogonalization::Bool, atol::Float64, rtol::Float64, itmax::Int64, timemax::Float64, verbose::Int64, history::Bool, callback::Krylov.var"#164#171", iostream::Core.CoreSTDOUT)
    @ Krylov ~/.julia/packages/Krylov/Iuavd/src/gmres.jl:121
  [8] gmres(A::Matrix{Float64}, b::SVector{2, Float64})
    @ Krylov ~/.julia/packages/Krylov/Iuavd/src/gmres.jl:119
  [9] top-level scope
    @ REPL[7]:1
```

### LINKED ISSUE #769: ComponentArrays support
Similar to #766 

```julia
julia> using Krylov, ComponentArrays

julia> A = rand(2, 2)
2×2 Matrix{Float64}:
 0.106809  0.817828
 0.506456  0.728449

julia> b = ComponentVector(b1=rand(), b2=rand())
ComponentVector{Float64}(b1 = 0.11865752943169328, b2 = 0.5098226106263868)

julia> A \ b
2-element Vector{Float64}:
 0.9825279457773433
 0.016769695792115385

julia> gmres(A, b)
ERROR: MethodError: Cannot `convert` an object of type UndefInitializer to an object of type Vector{Float64}

Closest candidates are:
  convert(::Type{T}, ::LinearAlgebra.Factorization) where T<:AbstractArray
   @ LinearAlgebra ~/.julia/juliaup/julia-1.9.2+0.x64.linux.gnu/share/julia/stdlib/v1.9/LinearAlgebra/src/factorization.jl:59
  convert(::Type{<:Array}, ::ComponentArray)
   @ ComponentArrays ~/.julia/packages/ComponentArrays/ehUwN/src/similar_convert_copy.jl:58
  convert(::Type{T}, ::AbstractArray) where T<:Array
   @ Base array.jl:613
  ...

Stacktrace:
 [1] ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(b1 = 1, b2 = 2)}}}(data::UndefInitializer, axes::Int64)
   @ ComponentArrays ~/.julia/packages/ComponentArrays/ehUwN/src/componentarray.jl:35
 [2] GmresSolver(m::Int64, n::Int64, memory::Int64, S::Type{ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(b1 = 1, b2 = 2)}}}})
   @ Krylov ~/.julia/packages/Krylov/Iuavd/src/krylov_solvers.jl:1549
 [3] GmresSolver(A::Matrix{Float64}, b::ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(b1 = 1, b2 = 2)}}}, memory::Int64)
   @ Krylov ~/.julia/packages/Krylov/Iuavd/src/krylov_solvers.jl:1567
 [4] gmres(A::Matrix{Float64}, b::ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(b1 = 1, b2 = 2)}}}; memory::Int64, M::LinearAlgebra.UniformScaling{Bool}, N::LinearAlgebra.UniformScaling{Bool}, ldiv::Bool, restart::Bool, reorthogonalization::Bool, atol::Float64, rtol::Float64, itmax::Int64, timemax::Float64, verbose::Int64, history::Bool, callback::Krylov.var"#164#171", iostream::Core.CoreSTDOUT)
   @ Krylov ~/.julia/packages/Krylov/Iuavd/src/gmres.jl:121
 [5] gmres(A::Matrix{Float64}, b::ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(b1 = 1, b2 = 2)}}})
   @ Krylov ~/.julia/packages/Krylov/Iuavd/src/gmres.jl:119
 [6] top-level scope
   @ REPL[8]:1
```

gdalle:
Fix #766 by adding an extension for StaticArrays.jl and a custom dispatch for `ktypeof(::StaticVector)`.

The extension is loaded conditionally with Requires.jl for Julia < 1.9.
This means that Requires comes as a new dependency.

EDIT: also fix #769


gdalle:
@amontoison if you have time to take a quick look, I would really appreciate it. This is an important ingredient for the next release of ImplicitDifferentiation.jl, which as you may know relies heavily on Krylov and LinearOperators 🙏

amontoison:
Hi Guillaume!
Is it not possible to solve your issue by just updating this line: https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/krylov_utils.jl#L213?

gdalle:
Thank you for the lightning quick answer!

I'm not at my computer right now so I can't test what S.name.name does for a StaticVector, but it sounds like a working solution. The only trouble is that we would have to list each subtype of StaticVector manually, and that this particular line of code needs to change again for each new array package.

Philosophically, I think Julia 1.9 and its package extensions are a cleaner and more modular answer to this problem, but it's a bit heavier in terms of machinery.

Another doubt I have is whether the current name-based solution is type-stable. I guess it is but I'm not 100% sure, whereas dispatch definitely is.

gdalle:
And come to think of it, the current solution doesn't cover arbitrary subtypes of StaticVector because we have to list concrete ones with their names. That means anyone extending StaticVector will have the same issue.

amontoison:
> Philosophically, I think Julia 1.9 and its package extensions are a cleaner and more modular answer to this problem, but it's a bit heavier in terms of machinery.

I propose to wait the next LTS version for that (Julia 1.10).

> And come to think of it, the current solution doesn't cover arbitrary subtypes of StaticVector because we have to list concrete ones with their names. That means anyone extending StaticVector will have the same issue.

Is it related to the first type `S` of a `StaticArray`?

abelsiqueira:
I think using package extensions is the solution here. It allows using dispatch instead of relying on `.name.name`.
Is there any reason to wait for the next LTS? Is there (a lot of) performance degradation?

amontoison:
A system similar to Requires.jl is available in Julia 1.9 so we will not need to check if `VERSION < 1.9` with Julia 1.10 because it's the next LTS.

abelsiqueira:
The system similar to Requires.jl is the package extensions. Or is there something else as well?
You don't have to check for VERSION < 1.9, although you need to check for `:get_extensions` and use Requires.jl then - which is what is being done here. Is that your opposition? Using the conditional to determine whether to use Requires.jl? Are you concerned with something in particular?
@tmigot and I were discussing JSOSuite using the same system to support multiple solvers, and it will probably be useful in other places as well, so more info is appreaciated.

gdalle:
@abelsiqueira is right, this is what I'm doing in the PR. Using the built in extension mechanism introduced in 1.9, and Requires for previous versions. See https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)

amontoison:
> I'm not at my computer right now so I can't test what S.name.name does for a StaticVector, but it sounds like a working solution. The only trouble is that we would have to list each subtype of StaticVector manually, and that this particular line of code needs to change again for each new array package.

`S.name.name` returns `:SArray`. I don't understand why do your have to list each subtype of a StaticVector.
If you replace
```julia
if S.name.name == :Zeros || S.name.name == :Ones
```
by
```julia
if S.name.name == :Zeros || S.name.name == :Ones || S.name.name == :SArray
```
it should work.

> You don't have to check for VERSION < 1.9, although you need to check for :get_extensions and use Requires.jl then - which is what is being done here.
Is that your opposition? Using the conditional to determine whether to use Requires.jl? Are you concerned with something in particular?

I just find my solution simple and type-stable.
Why do we need to modify 58 lines, add a dependency and an extension if just one is enough and is doing exactly the same job?

If we add one extension for `StaticArrays.jl`, we should also do the same thing for `CUDA.jl`, `AMDGPU.jl`, `OneAPI.jl`, etc...
I prefer to do that when Julia 1.10 will be released such that I don't need to update again these files.

I don't have any opposition again package modules otherwise. :smiley:

abelsiqueira:
It's good that the current solution is type stable. @gdalle, can you comment if that works for your current use case? It looks like it can be tested using the branch from #768.

@amontoison, your code is simple, but also it is obfuscated. `S.name.name` and the symbols (`:Zeros`, etc.) are not really descriptive. Having an extension and using dispatch makes it clear from which package it comes and which function made that happen.

> If we add one extension for StaticArrays.jl, we should also do the same thing for CUDA.jl, AMDGPU.jl, OneAPI.jl, etc...

That is the idea (if necessary). Extensions are supposed to be free and efficient.

> I prefer to do that when Julia 1.10 will be released such that I don't need to update again these files.

The code won't break when 1.10 is released, and if we don't immediately drop 1.6, it might remain in use. Furthermore, I hope @gdalle will still be around to help update the code if needed.

All that being said, my main concern is the cost increase of adding Requires as a dependency. I wonder if either of you has some benchmarks on 1.6 and 1.9?

gdalle:
Indeed the idea is that extensions are basically free, and packages can support as many as they want. I encourage you to watch Kristoffer's JuliaCon 2023 talk about them. 

It takes a bit more code to set them up, but I have done this before and I can help. Once it's done, they are much more readable and maintainable than what's currently in place.

Requires is only necessary for < 1.9, it's not even loaded (just downloaded) in the latest version. I'll run some benchmarks for you in the evening for download, precompilation and loading time

gdalle:
I'll test the #768 branch later tonight, thanks for the quick fix! That PR will probably solve my immediate issue, so I'd be happy if it gets merged quickly. And in the meantime I could turn the current PR into a more generic move from name.name to extensions for all array backends that are known to cause problems?

gdalle:
I did the same thing for FillArrays, which allowed me to remove some lines in `krylov_utils.jl`. I also moved from Requires to PackageExtensionTools, which further cuts down the additional code.

If you disregard the new tests (new tests are never a waste of LOCs), I think you'll find this PR hits a very reasonable tradeoff in complexity :)

gdalle:
Here are some benchmarks, obtained with the following commands:

```julia
shell> rm -rf ~/.julia/compiled; rm -rf ~/.julia/packages;

julia> ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0;

julia> using Pkg

julia> Pkg.activate(temp=true)

# julia> Pkg.add(url="https://github.com/gdalle/Krylov.jl", rev="staticarrays_extension")
julia> Pkg.add(url="https://github.com/JuliaSmoothOptimizers/Krylov.jl", rev="main")

# julia> Pkg.precompile("Krylov", timing=true)
julia> @time Pkg.precompile("Krylov") # on 1.8

julia> @time_imports using Krylov

shell> rm -rf ~/.julia/compiled; rm -rf ~/.julia/packages;

julia> @show VERSION
```

**Julia 1.9**

This branch:

```julia
julia> Pkg.precompile("Krylov", timing=true)
Precompiling Krylov
    900.1 ms  ✓ PackageExtensionCompat
   7800.2 ms  ✓ Krylov
  2 dependencies successfully precompiled in 9 seconds

julia> @time_imports using Krylov
      0.9 ms  PackageExtensionCompat
    246.2 ms  Krylov
```

Main branch:

```julia
julia> Pkg.precompile("Krylov", timing=true)
Precompiling Krylov
   7111.4 ms  ✓ Krylov
  1 dependency successfully precompiled in 7 seconds

julia> @time_imports using Krylov
    224.4 ms  Krylov
```

**Julia 1.8**

This branch:

```julia
julia> @time Pkg.precompile("Krylov") # on 1.8
Precompiling project...
  3 dependencies successfully precompiled in 9 seconds
  8.552460 seconds (98.20 k allocations: 5.775 MiB, 1.46% compilation time)

julia> @time_imports using Krylov
      0.5 ms  Requires
      0.2 ms  PackageExtensionCompat
     84.1 ms  Krylov 32.45% compilation time (8% recompilation)
```

Main branch:

```julia
julia> @time Pkg.precompile("Krylov") # on 1.8
Precompiling project...
  1 dependency successfully precompiled in 7 seconds
  7.477803 seconds (67.81 k allocations: 3.921 MiB, 1.42% gc time, 2.57% compilation time)

julia> @time_imports using Krylov
     45.2 ms  Krylov
```

gdalle:
@amontoison I think this is ready, if you want to approve the workflows

gdalle:
Found this old PR while cleaning things up, is it still relevant @amontoison ?

amontoison:
Hi @gdalle!
Do you have some array types that are not yet supported?
Otherwise, I propose to wait the next LTS version of Julia.

oscardssmith:
I really think this approach is the right way to go. If we want to avoid the overhead for Julia <1.9, we could continue to use the `.name.name` approach and not load the extension with `Requires` at all for older julia versions. Especially since https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/768 didn't actually fix ComponentArrays (see https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/769#issuecomment-1671899667) I think this PR is the better way forward.

amontoison:
@oscardssmith Is your use case with `ComponentArrays`?
I wanted to wait the next LTS release and add extensions for all relevant arrays types.

oscardssmith:
Yes.
Why do you want to wait? I can understand not wanting to support them on versions <1.9, but there's no rule that says that all features need to be supported on all julia versions. Surely supporting ComponentArrays on Julia 1.9+ now is better than not supporting them on any julia version.

amontoison:
> Yes. Why do you want to wait? I can understand not wanting to support them on versions <1.9, but there's no rule that says that all features need to be supported on all julia versions. Surely supporting ComponentArrays on Julia 1.9+ now is better than not supporting them on any julia version.

I wanted to keep support for the LTS version but not using `Requires`, only package extensions.

oscardssmith:
What's wrong with using `Requires` on LTS?

amontoison:
@oscardssmith I opened #859 to fix the issue with ComponentArrays.

oscardssmith:
Thanks! I do really think that pkg extension/requires is a better fix, but #859 is definitely better than status quo.

gdalle:
You can also have package extensions that are just not loaded on <1.9, no need for Requires in that case

amontoison:
> Thanks! I do really think that pkg extension/requires is a better fix, but #859 is definitely better than status quo.

I agree with both of you Oscar and Guillaume.

> You can also have package extensions that are just not loaded on <1.9, no need for Requires in that case

Oh, good to know!

I plan to do a release 1.0.0 of Krylov.jl this year, let's drop Julia 1.6 and requires at least Julia 1.9 and add package extensions at this moment. Is it fine for you?

gdalle:
Now that 1.10 is the LTS, it might be time to revisit this

amontoison:
Is it still relevant after #947?

gdalle:
Thanks for taking steps towards this fix!
Unfortunately, judging by https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/769#issuecomment-2599503781, the solution requires type piracy from the user?
If I am not the developer of either ComponentArrays or Krylov, then doing this is considered bad practice:
```julia
Krylov.ktypeof(v::ComponentArray) = typeof(v)
```
Of course the impacts are likely to be limited, but that is one of the reasons why I recommended package extensions instead. What do you think?

amontoison:
> Thanks for taking steps towards this fix! Unfortunately, judging by [#769 (comment)](https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/769#issuecomment-2599503781), the solution requires type piracy from the user? If I am not the developer of either ComponentArrays or Krylov, then doing this is considered bad practice:
> 
> ```julia
> Krylov.ktypeof(v::ComponentArray) = typeof(v)
> ```
> 
> Of course the impacts are likely to be limited, but that is one of the reasons why I recommended package extensions instead. What do you think?

@gdalle
We don't need any type piracy if I remove this line in Krylov.jl:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/krylov_utils.jl#L209

```julia
Krylov.ktypeof(v::ComponentArray) = typeof(v)
```
was just used to restore the default behavior.

I am just wondering what is the most relevant for the user. Keep some componentVector in the workspace or just Vector?

gdalle:
My point from earlier still applies: checking `sometype.name.name` as in the line you mention is a bit of a hack and relies on internals. The clean way to do this is with a package extension, which also allows for proper versioning of the dependencies, etc. This is even more true now that 1.10 is the LTS.

> I am just wondering what is the most relevant for the user. Keep some componentVector in the workspace or just Vector?

Does the user interact with the Krylov workspace? As long as the solution is a `ComponentArray` like the inputs, I don't think it is crucial what kind of data you use internally?

gdalle:
On second thought it would probably be best to keep everything the same type as the input, eg for GPUs and StaticArrays and such

amontoison:
@gdalle 
I close the PR but don't hesitate to open new issues if needed.
It seems that the current solution work for you.

PR #768, opened 2023-08-08
StaticArrays support

amontoison:


src/krylov_utils.jl
```diff
@@ -210,9 +210,9 @@ function ktypeof(v::S) where S <: DenseVector
 end
 
 function ktypeof(v::S) where S <: AbstractVector
-  if S.name.name == :Zeros || S.name.name == :Ones
+  if S.name.name == :Zeros || S.name.name == :Ones || S.name.name == :SArray || S.name.name == :MArray || S.name.name == :SizedArray
```
gdalle, line 213: I think `:FieldArray` is also a valid subtype of `StaticVector`
```julia
julia> subtypes(StaticVector)
5-element Vector{Any}:
 FieldArray{Tuple{N}, T, 1} where {N, T}
 MArray{Tuple{N}, T, 1} where {N, T}
 SArray{Tuple{N}, T, 1} where {N, T}
 SizedArray{Tuple{N}, T, 1, M} where {N, T, M}
 StaticArrays.SUnitRange
```

gdalle (changes requested):

gdalle:
This works for my use case, thanks for the rapid fix! I'm gonna try to demo a more principled version over on #767

test/test_extensions.jl
```diff
@@ -0,0 +1,25 @@
+using LinearAlgebra, SparseArrays, Test
+using Krylov, StaticArrays
+
+@testset "StaticArrays" begin
```
gdalle, line 4: Would it make sense to also test other array formats like FillArrays? Test dependencies are free

gdalle:
Given #769 I really think the right fix is #767 instead of editing this single line forever with manual type names

src/krylov_utils.jl
gdalle: And if we could add `:ComponentArray` that would solve all my immediate problems

test/test_extensions.jl
amontoison: Yes, I will do that with another PR.

PR #815, opened 2023-09-26
[JOSS] Update CITATION.cff and DOI badge

amontoison:


amontoison:
@abelsiqueira 
As an expert of cff files, do you know if I should keep the following fields in the cff file for Zenedo?
```
identifiers:
  - description: Zenodo archive
    type: doi
    value: 10.5281/zenodo.822073
keywords:
  - Julia
  - linear algebra
  - Krylov methods
  - Krylov processes
  - sparse linear systems
  - GPU computing
license: MPL-2.0
repository-code: "https://github.com/JuliaSmoothOptimizers/Krylov.jl"
```
I replaced the cff file by the one suggested by JOSS but maybe I should do a mix of them.

abelsiqueira:
You should. In fact, the `doi` field at root level is specific for version 0.9.4, and the one we have here is general, so use that instead. You can also create another identifier, with type "other" that links to the JOSS URL.
The relevant part of what they provided is the "preferred-citation" field, so you can just copy that into our usual CITATION.cff. You can use the cffinit website to validate, if needed.

amontoison:
Thanks @abelsiqueira, I updated the CITATION.cff with your comments :smiley:

PR #823, opened 2023-10-11
Test view of AMD GPUs

amontoison:


amontoison:
@pxl-th Are views supported in `AMDGPU.jl`?

pxl-th:
Yes, we do support views. In fact, I think at this point all GPU backends share the same mechainism for views support.

PR #866, opened 2024-05-22
Fix copy_triangle for Julia 1.11

amontoison:


amontoison:
Related issue: https://github.com/JuliaLang/julia/issues/54546

Issue #876, opened 2024-09-26
Document content of stats


gdalle:
It would be nice to have more details on the meaning of each field in the Private = false
. Typically I struggle to interpret the difference between `solved` and `inconsistent`.

amontoison:
Yes, I wanted to that since a long time.
`inconsistent` means that the right-hand side `b` is not in the range of `A`. `Ax = b` can't be solved.
However you can still solve `min ||b - Ax||` is the solver has stopping conditions of `||r_k||` and `||A'r_k||`.

amontoison:
#917 provides a documentation for the most relevant structure `SimpleStats`.
I still need to document the other ones.

PR #917, opened 2024-10-29
[documentation] Improve the docstring of SimpleStats

amontoison:
#876 
cc @gdalle

https://jso.dev/Krylov.jl/previews/PR917

### DIFF
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -11,7 +11,7 @@ Type for storing statistics returned by the majority of Krylov solvers.
 The fields are as follows:
 - `niter`: The total number of iterations completed by the solver;
 - `solved`: Indicates whether the solver successfully reached convergence (`true` if solved, ``false` otherwise);
-- `inconsistent`: Flags whether the system was detected as inconsistent during the solving process;
+- `inconsistent`: Flags whether the system was detected as inconsistent (i.e., when `b` is not in the range of `A`);
 - `residuals`: A vector containing the residual norms at each iteration;
 - `Aresiduals`: A vector of `Aᴴ`-residual norms at each iteration;
 - `Acond`: An estimate of the condition number of matrix `A`.
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -13,7 +13,7 @@ The fields are as follows:
 - `solved`: Indicates whether the solver successfully reached convergence (`true` if solved, ``false` otherwise);
 - `inconsistent`: Flags whether the system was detected as inconsistent (i.e., when `b` is not in the range of `A`);
 - `residuals`: A vector containing the residual norms at each iteration;
-- `Aresiduals`: A vector of `Aᴴ`-residual norms at each iteration;
+- `Aresiduals`: A vector of `A'`-residual norms at each iteration;
 - `Acond`: An estimate of the condition number of matrix `A`.
 - `timer`: The elapsed time (in seconds) taken by the solver to complete all iterations;
 - `status`: A string indicating the outcome of the solve, providing additional details beyond `solved`.
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -7,15 +7,16 @@ import Base.copyto!
 abstract type KrylovStats{T} end
 
 """
-Type for statistics returned by the majority of Krylov solvers, the attributes are:
-- niter
-- solved
-- inconsistent
-- residuals
-- Aresiduals
-- Acond
-- timer
-- status
+Type for storing statistics returned by the majority of Krylov solvers.
+The fields are as follows:
+- `niter`: The total number of iterations completed by the solver;
+- `solved`: Indicates whether the solver successfully reached convergence (`true` if solved, ``false` otherwise);
+- `inconsistent`: Flags whether the system was detected as inconsistent during the solving process;
+- `residuals`: A vector containing the residual norms at each iteration;
+- `Aresiduals`: A vector of `Aᴴ`-residual norms at each iteration;
+- `Acond`: An estimate of the condition number of matrix `A`.
+- `timer`: The elapsed time (in seconds) taken by the solver to complete all iterations;
+- `status`: A string indicating the outcome of the solve, providing additional details beyond `solved`.
 """
 mutable struct SimpleStats{T} <: KrylovStats{T}
   niter        :: Int
@@ -47,7 +48,8 @@ function copyto!(dest :: SimpleStats, src :: SimpleStats)
 end
 
 """
-Type for statistics returned by LSMR. The attributes are:
+Type for storing statistics returned by LSMR.
+The fields are as follows:
 - niter
 - solved
 - inconsistent
@@ -96,7 +98,8 @@ function copyto!(dest :: LsmrStats, src :: LsmrStats)
 end
 
 """
-Type for statistics returned by CG-LANCZOS, the attributes are:
+Type for storing statistics returned by CG-LANCZOS.
+The fields are as follows:
 - niter
 - solved
 - residuals
@@ -134,7 +137,8 @@ function copyto!(dest :: LanczosStats, src :: LanczosStats)
 end
 
 """
-Type for statistics returned by CG-LANCZOS with shifts, the attributes are:
+Type for storing statistics returned by CG-LANCZOS-SHIFT and CGLS-LANCZOS-SHIFT.
+The fields are as follows:
 - niter
 - solved
 - residuals
@@ -174,7 +178,8 @@ function copyto!(dest :: LanczosShiftStats, src :: LanczosShiftStats)
 end
 
 """
-Type for statistics returned by SYMMLQ, the attributes are:
+Type for storing statistics returned by SYMMLQ.
+The fields are as follows:
 - niter
 - solved
 - residuals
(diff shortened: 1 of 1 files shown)


src/krylov_stats.jl
```diff
@@ -7,15 +7,16 @@ import Base.copyto!
...
-- residuals
-- Aresiduals
-- Acond
-- timer
-- status
+Type for storing statistics returned by the majority of Krylov solvers.
+The fields are as follows:
+- `niter`: The total number of iterations completed by the solver;
+- `solved`: Indicates whether the solver successfully reached convergence (`true` if solved, ``false` otherwise);
+- `inconsistent`: Flags whether the system was detected as inconsistent during the solving process;
```
amontoison, line 14:
```suggestion
- `inconsistent`: Flags whether the system was detected as inconsistent (`b` not in the range of `A`);
```

src/krylov_stats.jl
```diff
@@ -7,15 +7,16 @@ import Base.copyto!
...
-- inconsistent
-- residuals
-- Aresiduals
-- Acond
-- timer
-- status
+Type for storing statistics returned by the majority of Krylov solvers.
+The fields are as follows:
+- `niter`: The total number of iterations completed by the solver;
+- `solved`: Indicates whether the solver successfully reached convergence (`true` if solved, ``false` otherwise);
```
gdalle, line 13:
```suggestion
- `solved`: Indicates whether the solver successfully reached convergence (`true` if solved, `false` otherwise);
```

gdalle: Also, convergence according to which criterion? A tolerance I assume?

src/krylov_stats.jl
```diff
@@ -7,15 +7,16 @@ import Base.copyto!
...
+- `inconsistent`: Flags whether the system was detected as inconsistent (i.e., when `b` is not in the range of `A`);
+- `residuals`: A vector containing the residual norms at each iteration;
+- `Aresiduals`: A vector of `Aᴴ`-residual norms at each iteration;
```
gdalle, line 16: What is `Aᴴ`?

src/krylov_stats.jl
```diff
@@ -7,15 +7,16 @@ import Base.copyto!
...
+- `Acond`: An estimate of the condition number of matrix `A`.
+- `timer`: The elapsed time (in seconds) taken by the solver to complete all iterations;
+- `status`: A string indicating the outcome of the solve, providing additional details beyond `solved`.
```
gdalle, line 19: Can you give an example? What are its possible values?

src/krylov_stats.jl
```diff
@@ -47,7 +48,8 @@ function copyto!(dest :: SimpleStats, src :: SimpleStats)
 end
 
 """
-Type for statistics returned by LSMR. The attributes are:
+Type for storing statistics returned by LSMR.
+The fields are as follows:
```
gdalle, line 52:
```suggestion
The fields are as follows (see [`Krylov.SimpleStats`](@ref)):
```

gdalle: Same for the other more sophisticated stats, so that you don't need to document every field again

gdalle: What is `xNorm` here?

src/krylov_stats.jl
```diff
@@ -96,7 +98,8 @@ function copyto!(dest :: LsmrStats, src :: LsmrStats)
 end
 
 """
-Type for statistics returned by CG-LANCZOS, the attributes are:
+Type for storing statistics returned by CG-LANCZOS.
+The fields are as follows:
```
gdalle, line 102:
```suggestion
The fields are as follows (see [`Krylov.SimpleStats`](@ref)):
```

gdalle: What is `indefinite` here?

src/krylov_stats.jl
```diff
@@ -134,7 +137,8 @@ function copyto!(dest :: LanczosStats, src :: LanczosStats)
 end
 
 """
-Type for statistics returned by CG-LANCZOS with shifts, the attributes are:
+Type for storing statistics returned by CG-LANCZOS-SHIFT and CGLS-LANCZOS-SHIFT.
+The fields are as follows:
```
gdalle, line 141:
```suggestion
The fields are as follows (see [`Krylov.SimpleStats`](@ref)):
```

gdalle: What is `indefinite` here?

src/krylov_stats.jl
```diff
@@ -174,7 +178,8 @@ function copyto!(dest :: LanczosShiftStats, src :: LanczosShiftStats)
 end
 
 """
-Type for statistics returned by SYMMLQ, the attributes are:
+Type for storing statistics returned by SYMMLQ.
+The fields are as follows:
```
gdalle, line 182:
```suggestion
The fields are as follows (see [`Krylov.SimpleStats`](@ref)):
```

gdalle: What are `residualscg`, `errors` and `errorscg` here?

src/krylov_stats.jl
```diff
@@ -221,7 +226,8 @@ function copyto!(dest :: SymmlqStats, src :: SymmlqStats)
 end
 
 """
-Type for statistics returned by adjoint systems solvers BiLQR and TriLQR, the attributes are:
+Type for storing statistics returned by adjoint systems solvers BiLQR and TriLQR.
+The fields are as follows:
```
gdalle, line 230:
```suggestion
The fields are as follows (see [`Krylov.SimpleStats`](@ref)):
```

gdalle: What do the `primal` and `dual` suffixes mean here?

src/krylov_stats.jl
```diff
@@ -258,7 +264,8 @@ end
 
 
 """
-Type for statistics returned by the LNLQ method, the attributes are:
+Type for storing statistics returned by the LNLQ method.
+The fields are as follows:
```
gdalle, line 268:
```suggestion
The fields are as follows (see [`Krylov.SimpleStats`](@ref)):
```

gdalle: What do the `error_` fields mean here?

src/krylov_stats.jl
```diff
@@ -298,7 +305,8 @@ function copyto!(dest :: LNLQStats, src :: LNLQStats)
 end
 
 """
-Type for statistics returned by the LSLQ method, the attributes are:
+Type for storing statistics returned by the LSLQ method.
+The fields are as follows:
```
gdalle, line 309:
```suggestion
The fields are as follows (see [`Krylov.SimpleStats`](@ref)):
```

gdalle: What are the new fields here?

gdalle (commented):
Thanks for documenting this! I made a few suggestions, mostly to refer to `SimpleStats` in the more complicated stats docstrings and to only document the fields that don't appear in the simple versoin

src/krylov_stats.jl
amontoison: The adjoint of `A` Guillaume.
`A^T` is the transpose` and `A^H` the conjugate hermitian`.
In the real case `A^H` are `A^T` are the same matrices.

amontoison: https://jso.dev/Krylov.jl/dev/processes/#Notation

src/krylov_stats.jl
amontoison: The norm of the solution `x`.
You can compute it cheaply for many methods because the Krylov basis is orthonormal.

src/krylov_stats.jl
amontoison: It means that the matrix is indefinite.
A matrix is indefinite is it has positive and negative eigenvalues.

src/krylov_stats.jl
gdalle: Some of those were probably explained elsewhere in the docs, like `A^H`, but yeah the point of all my review comments was getting some clarification in the stats docstring itself.

src/krylov_stats.jl
gdalle: okay, we should probably explain it here then

src/krylov_stats.jl
gdalle: okay, we should probably explain it here or in a high-level preamble

src/krylov_stats.jl
gdalle: maybe this docstring could link to an explanation of the algorithm itself for these details

amontoison: It's just that you checked the docstring in the wrong way.
I don't think any user wants to understand how to use a result before knowing how to obtain it. 😉

Nobody will ever use `AdjointStats` without previously calling `bilqr` and `trilqr`, which explain everything and that gives a pointer to this structure: https://jso.dev/Krylov.jl/dev/solvers/as/

amontoison:
@gdalle I agree with you on many point but for specific statistics like LsmrStats, LnlqStats, AdjointStats, etc...
Many details are explained in the docstring of the Krylov method itself.
I don't think that any user will need to understand LnlqStats if he is not using the method that returned this structure. 

Note also that the title of the PR is `Improve the docstring of SimpleStats`.
By lack of time I didn't do the other stuctures yet, so you review of what you don't understand is meningful but not related to the PR.

src/krylov_stats.jl
amontoison, line 16:
```suggestion
- `Aresiduals`: A vector of `A'`-residual norms at each iteration;
```

PR #924, opened 2024-11-01
Update with JSOBestieTemplate for Krylov

tmigot:
Automated update using JSOBestieTemplate


amontoison:
@tmigot Why do remove `CommentPR.yml`?
It allows to update the previous breakage message. Otherwise after 35 rebase, we have 35 messages and it's just spamming.

amontoison:
> @tmigot Why do you remove `CommentPR.yml`? It allows to update the previous breakage message. Otherwise after 35 rebase, we have 35 messages and it's just spamming.

amontoison (approved):

tmigot:
Hi @amontoison ! Sorry, I didn't reply in time, I was also looking for some time to update the script. I see that you were faster :).
Can you open a PR here https://github.com/JuliaSmoothOptimizers/JSOBestieTemplate.jl to include your changes ?

amontoison:
It was planned :)
I just want to check if it works as intended before in this PR #927

tmigot:
Also, note that with this script we only cover JSO packages in the Breakage.
So, if you would want to test other packages I would suggest to add another script Krylov.jl-specific.

PR #958, opened 2025-02-11
CR: Update iterate on first iteration for negative curvature with line search and when zero curvature is detected

farhadrclass:
This PR fixes an issue in the Conjugate Residual (CR) implementation (in `src/cr.jl`) where, on the first iteration (`iter == 0`), if the computed curvature is non-positive (i.e. either zero or negative), the algorithm may return without updating the iterate—even though we have  
$$b = -\nabla f(x_k) \neq 0.$$

To address this, when line search is enabled and a non-positive curvature (zero or negative) is detected at iteration 0, we now explicitly update the iterate by performing

```julia
x .= b
```

This change guarantees that the method behaves consistently by setting `x` appropriately when curvature information is either zero or negative during the first iteration.

---

**Changes Made:**

1. **First Iteration Check (iter == 0):**  
   - When either zero or negative curvature is detected (i.e. `s <= 0`), and `linesearch` is enabled, update `x` to `b` before returning.

2. **Zero Curvature Check (general case):**  
   - (Existing handling) When a zero curvature is detected later in the method (e.g. at a different check), we continue to update `x` to `b` if `linesearch` is enabled.

### DIFF
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -150,9 +150,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
     end
     MisI && kcopy!(n, r, p)  # r ← p
     MisI || mulorldiv!(r, M, p, ldiv)
-    mul!(Ar, A, r)
-    ρ = kdotr(n, r, Ar)
-
+    
     rNorm = knorm_elliptic(n, r, p)  # ‖r‖
     history && push!(rNorms, rNorm)  # Values of ‖r‖
 
@@ -160,17 +158,20 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
       stats.niter = 0
       stats.solved, stats.inconsistent = true, false
       stats.timer = start_time |> ktimer
-      stats.status = "x = 0 is a zero-residual solution"
+      stats.status = "x is a zero-residual solution"
       history && push!(ArNorms, zero(T))
       solver.warm_start = false
       return solver
     end
+
+    mul!(Ar, A, r)
+    ρ = kdotr(n, r, Ar)
     
     if ρ == 0
       stats.niter = 0
       stats.solved, stats.inconsistent = true, false
       stats.timer = start_time |> ktimer
-      stats.status = "0 is a zero-curvature direction"
+      stats.status = "b is a zero-curvature direction"
       history && push!(ArNorms, zero(T))
       solver.warm_start = false
       linesearch && kcopy!(n, x, b)  # x ← b
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -201,8 +201,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         if (pAp ≤ γ * pNorm²) || (ρ ≤ γ * rNorm²)
           npcurv = true
           (verbose > 0) && @printf(iostream, "nonpositive curvature detected: pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
-          solved = true
-          stats.solved = solved
+          stats.solved = true
           stats.inconsistent = false
           stats.timer = start_time |> ktimer
           stats.status = "nonpositive curvature"
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -163,6 +163,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
       stats.status = "x = 0 is a zero-residual solution"
       history && push!(ArNorms, zero(T))
       solver.warm_start = false
+      linesearch && kcopy!(n, x, b)  # x ← b
       return solver
     end
     kcopy!(n, p, r)   # p ← r
@@ -200,10 +201,12 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         if (pAp ≤ γ * pNorm²) || (ρ ≤ γ * rNorm²)
           npcurv = true
           (verbose > 0) && @printf(iostream, "nonpositive curvature detected: pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
+          solved = true
           stats.solved = solved
           stats.inconsistent = false
           stats.timer = start_time |> ktimer
           stats.status = "nonpositive curvature"
+          iter == 0 && kcopy!(n, x, b)  # x ← b
           return solver
         end
       elseif pAp ≤ 0 && radius == 0
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -156,16 +156,27 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
     rNorm = knorm_elliptic(n, r, p)  # ‖r‖
     history && push!(rNorms, rNorm)  # Values of ‖r‖
 
+    if rNorm == 0
+      stats.niter = 0
+      stats.solved, stats.inconsistent = true, false
+      stats.timer = start_time |> ktimer
+      stats.status = "x = 0 is a zero-residual solution"
+      history && push!(ArNorms, zero(T))
+      solver.warm_start = false
+      return solver
+    end
+    
     if ρ == 0
       stats.niter = 0
       stats.solved, stats.inconsistent = true, false
       stats.timer = start_time |> ktimer
-      stats.status = "x is a zero-residual solution"
+      stats.status = "0 is a zero-curvature direction"
       history && push!(ArNorms, zero(T))
       solver.warm_start = false
       linesearch && kcopy!(n, x, b)  # x ← b
       return solver
     end
+
     kcopy!(n, p, r)   # p ← r
     kcopy!(n, q, Ar)  # q ← Ar
     (verbose > 0) && (m = zero(T))  # quadratic model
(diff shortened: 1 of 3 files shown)


src/cr.jl
```diff
@@ -200,10 +201,11 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         if (pAp ≤ γ * pNorm²) || (ρ ≤ γ * rNorm²)
           npcurv = true
           (verbose > 0) && @printf(iostream, "nonpositive curvature detected: pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
-          stats.solved = solved
+          stats.solved = true
```
amontoison, line 204: Should we consider that the system is solved if we have nonpositive curvature?

dpo: For the purposes of an inexact Newton method, yes. It’s similar to CG with `linesearch = true`.

farhadrclass:
@dpo  added couple tests and added some function to create the needed systems.
I think in future we need to expand the tests to CG solver as well since it is missing them

src/cr.jl
```diff
@@ -160,9 +160,10 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
       stats.niter = 0
       stats.solved, stats.inconsistent = true, false
       stats.timer = start_time |> ktimer
-      stats.status = "x = 0 is a zero-residual solution"
+      stats.status = "x is a zero-residual solution"
```
dpo, line 163:
```suggestion
      stats.status = "0 is a zero-curvature direction"
```

There should be a test after line 157:

```julia
if rNorm == 0
  stats.niter = 0
  stats.solved, stats.inconsistent = true, false
  stats.status = "x = 0 is a zero-residual solution"
  solver.warm_start = false
  return solver
end
```

test/test_cr.jl
```diff
@@ -62,6 +62,31 @@
...
+      x, stats = cr(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+
+      # Test Linesearch which would stop on the first call since A is negative definite
+      A, b = symmetric_indefinite_negative_curv(FC=FC)
+      x, stats = cr(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+      @test stats.niter == 0
+      @test norm(x) == norm(b)
+
```
dpo, line 76:
```suggestion
```

dpo (changes requested):

test/test_cr.jl
```diff
@@ -62,6 +62,31 @@
...
+      A, b = symmetric_indefinite(FC=FC)
+      x, stats = cr(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+
+      # Test Linesearch which would stop on the first call since A is negative definite
+      A, b = indefinite_system(FC=FC)
+      x, stats = cr(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+      @test stats.niter == 0
+      @test norm(x) == norm(b)
```
dpo, line 75: This could be `all(x .== b)`.

src/cr.jl
dpo: Yes but:

1. The computation of `rNorm` and the test `rNorm == 0` should occur before `Ar` is computed;
2. the status when `ρ == 0` should be `b is a zero-curvature direction` (my fault).

src/cr.jl
```diff
@@ -156,7 +156,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
     rNorm = knorm_elliptic(n, r, p)  # ‖r‖
     history && push!(rNorms, rNorm)  # Values of ‖r‖
 
-    if ρ == 0
+    if rNorm == 0
```
dpo, line 159: Move this and the computation of `rNorm` **before** computing `Ar` and `ρ`. That was my previous comment.

src/cr.jl
```diff
@@ -165,6 +165,18 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
       solver.warm_start = false
       return solver
     end
+    
+    if ρ == 0
+      stats.niter = 0
+      stats.solved, stats.inconsistent = true, false
+      stats.timer = start_time |> ktimer
+      stats.status = "0 is a zero-curvature direction"
```
dpo, line 173: I also asked you the change this…

test/test_cr.jl
```diff
@@ -54,6 +54,7 @@
       @test norm(x) == 0
       @test stats.status == "x = 0 is a zero-residual solution"
 
+
```
dpo, line 57:
```suggestion
```

test/test_utils.jl
```diff
@@ -21,6 +21,60 @@ function symmetric_indefinite(n :: Int=10; FC=Float64)
   return A, b
 end
 
+
+# Symmetric and indefinite system with negative curvature.
+# need this to test CR with Linesearch and nonpositive curvature at the first iteration
+function indefinite_system(n::Int=10; FC=Float64, shift=5)
```
dpo, line 27: This looks like a duplicate of a function `indefinite_system` that already exists. Add the `shift` argument there if you need it, but don’t duplicate functions.

dpo (changes requested):

src/cr.jl
farhadrclass: the rnorm is calcuated by p, `    rNorm = knorm_elliptic(n, r, p)  # ‖r‖`  so you want the test before this and also recalculate it after ?

dpo (approved):
LGTM. How about you @amontoison ?

amontoison:
I think we need to replace `iter == 0 && kcopy!(n, x, b)` by `iter == 0 && kcopy!(n, x, p)` because `p` is the initial residual `b - Ax0`.
We could have an error if linesearch is combined with warm-start.

farhadrclass:
@amontoison Thanks for pointing it out. However, I don’t think we should do that.   
If we replace `b` with `p`, we’ll end up setting `x` to zero because, in line 170, we have:  

```julia
if ρ == 0
    stats.niter = 0
    stats.solved, stats.inconsistent = true, false
    stats.status = "p is a zero-curvature direction"
    solver.warm_start = false
    linesearch && kcopy!(n, x, p)  # x ← p
    return solver
end
```

I assume that when we use a warm start, the probability of `p` being zero is lower, so the chance of triggering this condition would be reduced.  

The original code was written this way. If we set `p` to `x` instead of `b` to `x`, my R2N solver would encounter the same issue as before (i.e., `x == 0`).  

What do you think?

amontoison:
> The original code was written this way. If we set `p` to `x` instead of `b` to `x`, my R2N solver would encounter the same issue as before (i.e., `x == 0`).

`p` and `b` have the same content at `iter == 0`, I don't understand how it is possible to have the old issue.

amontoison:
@farhadrclass This is not a question of probability; I just want to ensure that `x` contains a relevant solution for the user.

It’s quite complex to test all the options together (like linesearch, preconditioner, warm-start, etc.).  
I should frame the problem like this:  
If `x0 = 0`, we want to solve `A x = r0` where `r0 = b - A x0 = b`, and we set `x = b = p` if we encounter null curvature.
If `x0 ≠ 0`, we solve a different system in Krylov.jl, where the system is shifted.  
We solve `A Δx = r0` where `r0 = b - A x0`, and at the end, we compute `x = Δx + x0`.  
If we encounter null curvature on the shifted system, we should set `Δx = r0` from what I understand from your previous messages.  
Therefore, Krylov.jl needs to return `x = x0 + r0`. Since at `iter == 0`, `p = r0`, we need something like this:

```julia
linesearch && kcopy!(n, x, p)  # x ← p
```

@dpo I think I made a mistake in all the Krylov methods where I implemented warm start...  
For historical reasons, we checked whether the initial residual was zero or not.  
In the case of a warm start, if the residual is zero, then `Δx = 0`, but we still need to set `x = x0`.
I only do `x += x0` at the end of the code and not in specific corner cases where a solution is returned earlier...

dpo:
I think the `linesearch` and `warmstart` options are incompatible. In an inexact Newton method, we want to solve $Bs = -g$ where $s$ is a *step*, i.e., something that starts at zero, so that we can later update $x_{k+1} = x_k + s$ in the optimization method.

There is something special about the rhs $-g$ here: it's a descent direction. If we start from $s_0 = 0$ and encounter negative curvature at the first CG or CR iteration (along the direction $-g$, as it happens), we want to return $s = -g$.

If we supplied a "better" initial guess $s_0$ and solved instead $B \Delta s = r$, where $r = B s_0 + g$, and encountered negative curvature at the first iteration, we still want to return $-g$, because we don't know whether $r$ is descent direction.

If we encounter negative curvature at iteration $k \geq 1$, the code would currently return $s_0 + \Delta s_{k-1}$, but I'm really not sure what that iterate would represent from the point of view of the optimization method.

At least for the time being, I think we should make sure that `linesearch` and `warmstart` are not `true` at the same time.

Maybe `linesearch` is not a good name for that option because what I'm describing is specific to optimization. If you're solving nonlinear equations, there is no natural concept of curvature.

amontoison:
Thanks for the explanation @dpo!

You can merge the PR and I will take care to return errors if some options like `linesearch` / `radius` and `warmstart` are used together.

dpo:
Thanks @farhadrclass !

farhadrclass:
thanks @amontoison for help and working on the follow-up 
@dpo  thank you, do we need to update the release number (version number)?

dpo:
Sure. This is a bug fix.

farhadrclass:
Can I trigger the V0.9.10?

Issue #959, opened 2025-02-15
All more tests with the option warm-start


amontoison:
It seems that in some corner cases, the correct solution is not returned.
If `x0` is the exact solution of `Ax = b`, it's possible that I return `x = 0` but the residual of the shifted system is zero.

amontoison:
I need to fix almost all methods...
The good news is that TriLQR can handle the exact solution with warm-start 😛 
```julia
Test Summary:           | Pass  Fail  Total  Time
warm-start              |  108    50    158  3.4s
  Data Type: Float64    |   54    25     79  0.2s
    bilqr               |    4     2      6  0.0s
    trilqr              |    6            6  0.0s
    tricg               |    2     2      4  0.0s
    trimr               |    2     1      3  0.0s
    gpmr                |    2     1      3  0.0s
    minres_qlp          |    2     1      3  0.0s
    symmlq              |    2     1      3  0.0s
    cg                  |    2     1      3  0.0s
    cr                  |    2     1      3  0.0s
    car                 |    2     1      3  0.0s
    cg_lanczos          |    2     1      3  0.0s
    minres              |    2     1      3  0.0s
    minares             |    2     1      3  0.0s
    diom                |    2     1      3  0.0s
    dqgmres             |    2     1      3  0.0s
    fom                 |    2     1      3  0.0s
    gmres               |    2     1      3  0.0s
    fgmres              |    2     1      3  0.0s
    bicgstab            |    2     1      3  0.0s
    cgs                 |    2     1      3  0.0s
    bilq                |    2     1      3  0.0s
    qmr                 |    2     1      3  0.0s
    usymlq              |    2     1      3  0.0s
    usymqr              |    2     1      3  0.0s
  Data Type: ComplexF64 |   54    25     79  3.2s
    bilqr               |    4     2      6  0.0s
    trilqr              |    6            6  0.0s
    tricg               |    2     2      4  0.9s
    trimr               |    2     1      3  2.1s
    gpmr                |    2     1      3  0.0s
    minres_qlp          |    2     1      3  0.0s
    symmlq              |    2     1      3  0.0s
    cg                  |    2     1      3  0.0s
    cr                  |    2     1      3  0.0s
    car                 |    2     1      3  0.0s
    cg_lanczos          |    2     1      3  0.0s
    minres              |    2     1      3  0.0s
    minares             |    2     1      3  0.0s
    diom                |    2     1      3  0.0s
[...]

PR #961, opened 2025-02-15
Add more tests for the option warm-start

amontoison:
#959 
Return the correct solution if we do a warm-start with an exact solution.


farhadrclass:
@amontoison thanks for the update :)

Issue #939, opened 2025-01-05
Optimize block-MINRES


amontoison:
A version of block-MINRES was implemented in #884.
A few arrays should be moved to the block-Krylov solver and not allocated in the solve itself:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/block_minres.jl#L118-L126

amontoison:
Partially done in https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/1022

PR #969, opened 2025-02-24
Implement Line Search with Negative Curvature Detection for MINRES Based on Liu et al. (2022)

farhadrclass:
**Description:**  
This pull request introduces a new line search routine to the MINRES algorithm, incorporating negative curvature detection as outlined in Algorithm 1 of:

>Liu, Yang & Roosta, Fred. (2022). *A Newton-MR algorithm with complexity guarantees for nonconvex smooth unconstrained optimization*. [[10.48550/arXiv.2208.07095](https://doi.org/10.48550/arXiv.2208.07095)](https://doi.org/10.48550/arXiv.2208.07095)

**Key Changes:**  
- **Negative Curvature Detection:**  
  - The implementation now detects negative curvature based on Algorithm 1 from the paper.  
  - When negative curvature is detected:
    - If the linesearch flag is enabled **and** it occurs on the first iteration, the algorithm returns `b`.
    - Otherwise, it returns the last computed `r1`.
- **Linesearch Flag:**  
  - Introduced a new flag to enable or disable the line search mechanism, allowing users to control the enhanced step size adjustments.

- **Testing & Documentation:**  
  - Updated tests and documentation to cover the new negative curvature detection and line search behavior.

### DIFF
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -12,6 +12,7 @@ The fields are as follows:
 - `niter`: The total number of iterations completed by the solver;
 - `solved`: Indicates whether the solver successfully reached convergence (`true` if solved, `false` otherwise);
 - `inconsistent`: Flags whether the system was detected as inconsistent (i.e., when `b` is not in the range of `A`);
+- `indefinite`: Flags whether the system was detected as indefinite (i.e., when `A` is not positive definite);
 - `residuals`: A vector containing the residual norms at each iteration;
 - `Aresiduals`: A vector of `A'`-residual norms at each iteration;
 - `Acond`: An estimate of the condition number of matrix `A`.
@@ -22,6 +23,7 @@ mutable struct SimpleStats{T} <: KrylovStats{T}
   niter        :: Int
   solved       :: Bool
   inconsistent :: Bool
+  indefinite   :: Bool
   residuals    :: Vector{T}
   Aresiduals   :: Vector{T}
   Acond        :: Vector{T}
@@ -39,6 +41,7 @@ function copyto!(dest :: SimpleStats, src :: SimpleStats)
   dest.niter        = src.niter
   dest.solved       = src.solved
   dest.inconsistent = src.inconsistent
+  dest.indefinite   = src.indefinite
   dest.residuals    = copy(src.residuals)
   dest.Aresiduals   = copy(src.Aresiduals)
   dest.Acond        = copy(src.Acond)
@@ -46,58 +49,6 @@ function copyto!(dest :: SimpleStats, src :: SimpleStats)
   dest.status       = src.status
   return dest
 end
-
-"""
-Type for storing statistics returned by Conjugate Methods.
-Methods icludes:
-- CG (TODO)
-- CR (TODO)
-- MINRES
-The fields are as follows:
-- niter
-- solved
-- nonposi_curv: when a non-positive curvature is detected
-- linesearch: when a line search is performed
-- inconsistent
-- residuals
-- Aresiduals
-- Acond
-- timer
-- status
-"""
-mutable struct conStats{T} <: KrylovStats{T}
-  niter        :: Int
-  solved       :: Bool
-  nonposi_curv :: Bool
-  linesearch   :: Bool
-  inconsistent :: Bool
-  residuals    :: Vector{T}
-  Aresiduals   :: Vector{T}
-  Acond        :: Vector{T}
-  timer        :: Float64
-  status       :: String
-end
-
-function reset!(stats :: conStats)
-  empty!(stats.residuals)
-  empty!(stats.Aresiduals)
-  empty!(stats.Acond)
-end
-
-function copyto!(dest :: conStats, src :: conStats)
-  dest.niter        = src.niter
-  dest.solved       = src.solved
-  dest.nonposi_curv = src.nonposi_curv
-  dest.linesearch   = src.linesearch
-  dest.inconsistent = src.inconsistent
-  dest.residuals    = copy(src.residuals)
-  dest.Aresiduals   = copy(src.Aresiduals)
-  dest.Acond        = copy(src.Acond)
-  dest.timer        = src.timer
-  dest.status       = src.status
-  return dest
-end
-
 
 """
 Type for storing statistics returned by LSMR.
--- a/src/krylov_stats.jl
+++ b/src/krylov_stats.jl
@@ -48,6 +48,58 @@ function copyto!(dest :: SimpleStats, src :: SimpleStats)
 end
 
 """
+Type for storing statistics returned by Conjugate Methods.
+Methods icludes:
+- CG (TODO)
+- CR (TODO)
+- MINRES
+The fields are as follows:
+- niter
+- solved
+- nonposi_curv: when a non-positive curvature is detected
+- linesearch: when a line search is performed
+- inconsistent
+- residuals
+- Aresiduals
+- Acond
+- timer
+- status
+"""
+mutable struct conStats{T} <: KrylovStats{T}
+  niter        :: Int
+  solved       :: Bool
+  nonposi_curv :: Bool
+  linesearch   :: Bool
+  inconsistent :: Bool
+  residuals    :: Vector{T}
+  Aresiduals   :: Vector{T}
+  Acond        :: Vector{T}
+  timer        :: Float64
+  status       :: String
+end
+
+function reset!(stats :: conStats)
+  empty!(stats.residuals)
+  empty!(stats.Aresiduals)
+  empty!(stats.Acond)
+end
+
+function copyto!(dest :: conStats, src :: conStats)
+  dest.niter        = src.niter
+  dest.solved       = src.solved
+  dest.nonposi_curv = src.nonposi_curv
+  dest.linesearch   = src.linesearch
+  dest.inconsistent = src.inconsistent
+  dest.residuals    = copy(src.residuals)
+  dest.Aresiduals   = copy(src.Aresiduals)
+  dest.Acond        = copy(src.Acond)
+  dest.timer        = src.timer
+  dest.status       = src.status
+  return dest
+end
+
+
+"""
 Type for storing statistics returned by LSMR.
 The fields are as follows:
 - niter
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -16,7 +16,7 @@
 # detection to monotonicity properties," SIAM Journal on Optimization 32, no. 4 (2022): 2636–2661):
 #
 # If linesearch is true and negative curvature is detected at iteration k > 0, the solution from iteration k-1 is returned, and `solver.npc_dir` contains the last computed residual, which is a direction of nonpositive curvature.
-# If negative curvature is detected at first iteration, the method returns the right-hand side (i.e., the negative gradient).
+# If linesearch is true and negative curvature is detected at iteration k = 0, the right-hand side is returned as solution (i.e., the negative gradient), and `solver.npc_dir` contains [WHAT?].
 #
 # C. C. Paige and M. A. Saunders, Solution of Sparse Indefinite Systems of Linear Equations,
 # SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -323,8 +323,8 @@ kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itm
 
       if linesearch
         # compute the residual vector and store it in npc_dir
-        kscal!(n, sn * sn, npc_dir  )  # npc_dir  = sn * sn * npc_dir 
-        kaxpy!(n, -ϕbar * cs, v, npc_dir )   # npc_dir  = npc_dir  - ϕbar * cs * v
+        kscal!(n, sn * sn, npc_dir)  # npc_dir  = sn * sn * npc_dir 
+        kaxpy!(n, -ϕbar * cs, v, npc_dir)  # npc_dir  = npc_dir  - ϕbar * cs * v
       end
       
       # Final update of w.
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -18,13 +18,15 @@
 # Dominique Orban, <dominique.orban@gerad.ca>
 # Brussels, Belgium, June 2015.
 # Montréal, August 2015.
+#
+# Liu, Yang & Roosta, Fred. (2022). A Newton-MR algorithm with complexity guarantees for nonconvex smooth unconstrained optimization. 10.48550/arXiv.2208.07095. 
 
 export minres, minres!
 
 """
     (x, stats) = minres(A, b::AbstractVector{FC};
                         M=I, ldiv::Bool=false, window::Int=5,
-                        λ::T=zero(T), atol::T=√eps(T),
+                        linesearch::Bool=false, λ::T=zero(T), atol::T=√eps(T),
                         rtol::T=√eps(T), etol::T=√eps(T),
                         conlim::T=1/√eps(T), itmax::Int=0,
                         timemax::Float64=Inf, verbose::Int=0, history::Bool=false,
@@ -68,6 +70,7 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 * `M`: linear operator that models a Hermitian positive-definite matrix of size `n` used for centered preconditioning;
 * `ldiv`: define whether the preconditioner uses `ldiv!` or `mul!`;
 * `window`: number of iterations used to accumulate a lower bound on the error;
+* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If negative curvature is detected at iteration k > 0, the solution of iteration k-1 is returned. If negative curvature is detected at iteration 0, the right-hand side is returned (i.e., the negative gradient);
 * `λ`: regularization parameter;
 * `atol`: absolute stopping tolerance based on the residual norm;
 * `rtol`: relative stopping tolerance based on the residual norm;
@@ -88,6 +91,8 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 #### Reference
 
 * C. C. Paige and M. A. Saunders, [*Solution of Sparse Indefinite Systems of Linear Equations*](https://doi.org/10.1137/0712047), SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
+
+* Liu, Yang & Roosta, Fred. (2022). A Newton-MR algorithm with complexity guarantees for nonconvex smooth unconstrained optimization. 10.48550/arXiv.2208.07095. 
 """
 function minres end
 
@@ -108,6 +113,7 @@ def_optargs_minres = (:(x0::AbstractVector),)
 
 def_kwargs_minres = (:(; M = I                     ),
                      :(; ldiv::Bool = false        ),
+                     :(; linesearch::Bool = false  ),
                      :(; λ::T = zero(T)            ),
                      :(; atol::T = √eps(T)         ),
                      :(; rtol::T = √eps(T)         ),
@@ -124,7 +130,7 @@ def_kwargs_minres = extract_parameters.(def_kwargs_minres)
 
 args_minres = (:A, :b)
 optargs_minres = (:x0,)
-kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax, :verbose, :history, :callback, :iostream)
+kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax, :verbose, :history, :callback, :iostream)
 
 @eval begin
   function minres!(solver :: MinresSolver{T,FC,S}, $(def_args_minres...); $(def_kwargs_minres...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}, S <: AbstractVector{FC}}
@@ -173,18 +179,36 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
     # β₁ M v₁ = b.
     kcopy!(n, r2, r1)  # r2 ← r1
     MisI || mulorldiv!(v, M, r1, ldiv)
+
+    rNorm =  knorm_elliptic(n, r2, r1)  # = ‖r‖
+    history && push!(rNorms, rNorm)
+
+    if rNorm == 0
+      stats.niter = 0
+      stats.solved, stats.inconsistent = true, false
+      stats.timer = start_time |> ktimer
+      stats.status = "x is a zero-residual solution"
+      history && push!(rNorms, zero(T))
+      history && push!(ArNorms, zero(T))
+      history && push!(Aconds, zero(T))
+      warm_start && kaxpy!(n, one(FC), Δx, x)
+      solver.warm_start = false
+      return solver
+    end
+
     β₁ = kdotr(m, r1, v)
     β₁ < 0 && error("Preconditioner is not positive definite")
     if β₁ == 0
       stats.niter = 0
       stats.solved, stats.inconsistent = true, false
       stats.timer = start_time |> ktimer
-      stats.status = "x is a zero-residual solution"
+      stats.status = "b is a zero-curvature directions"
       history && push!(rNorms, β₁)
       history && push!(ArNorms, zero(T))
       history && push!(Aconds, zero(T))
       warm_start && kaxpy!(n, one(FC), Δx, x)
       solver.warm_start = false
+      linesearch && kcopy!(n, x, b) # x ← b
       return solver
     end
     β₁ = sqrt(β₁)
@@ -275,6 +299,21 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
       ArNorm = ϕbar * root  # = ‖Aᴴrₖ₋₁‖
       history && push!(ArNorms, ArNorm)
 
+      # Check if -ve curvature encountered.
+      if linesearch
+        if cs * γbar ≥ 0 # or eps(T)
+          (verbose > 0) && @printf(iostream, "nonpositive curvature detected:  cs * γbar = %e\n", cs * γbar)
+          stats.solved = true
+          stats.niter = iter
+          stats.inconsistent = false
+          stats.timer = start_time |> ktimer
+          stats.status = "nonpositive curvature"
+          iter == 1 && kcopy!(n, x, b)  # x ← b # we start  at iteration 1
+          solver.warm_start = false
+          return solver
+        end
+      end
+
       # Compute the next plane rotation.
       γ = sqrt(γbar * γbar + β * β)
       γ = max(γ, ϵM)
@@ -331,7 +370,7 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
         stats.solved, stats.inconsistent = true, true
         stats.timer = start_time |> ktimer
         stats.status = "x is a minimum least-squares solution"
-        warm_start && kaxpy!(n, one(FC), Δx, x)#Linesearch
+        warm_start && kaxpy!(n, one(FC), Δx, x)
         solver.warm_start = false
         return solver
       end
@@ -345,7 +384,6 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
       # solved_mach = (ϵx ≥ β₁)
 
       # Stopping conditions based on user-provided tolerances.
-      #TODO test3 then update x to r2
       tired = iter ≥ itmax
       ill_cond_lim = (one(T) / Acond ≤ ctol)
       solved_lim = (test2 ≤ ε)
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -323,10 +323,8 @@ kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itm
 
       if linesearch
         # calculating the residual npc_dir  = sn*sn * npc_dir  - ϕbar * cs   * v
-        sn2 = sn * sn
-        kscal!(n, sn2, npc_dir  )  # npc_dir  = sn2 * npc_dir 
-        ϕ_c = -ϕbar * cs 
-        kaxpy!(n, ϕ_c, v, npc_dir )   # npc_dir  = npc_dir  + ϕ_c * v
+        kscal!(n, sn * sn, npc_dir  )  # npc_dir  = sn * sn * npc_dir 
+        kaxpy!(n, -ϕbar * cs, v, npc_dir )   # npc_dir  = npc_dir  - ϕbar * cs * v
       end
       
 
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -307,7 +307,6 @@ kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itm
           stats.status = "nonpositive curvature"
           iter == 1 && kcopy!(n, x, b)
           solver.warm_start = false
-          # when we use the linesearch and encounter negative curvature, we return the last residual xk but user has access to npc_dir  from solver.npc_dir 
           stats.indefinite = true
           return solver
         end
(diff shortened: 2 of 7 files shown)


farhadrclass:
@dpo 
I think we need to test this case 

```julia
if β₁ == 0
      stats.niter = 0
      stats.solved, stats.inconsistent = true, false
      stats.status = "b is a zero-curvature directions"
      history && push!(rNorms, β₁)
      history && push!(Aconds, zero(T))
      solver.warm_start = false
      linesearch && kcopy!(n, x, b) # x ← b
      return solver
    end
```
But all the systems I test or create does not have \beta_1 == 0, any suggestions on a system that would force this case?

dpo:
$$\beta_1 = 0$$ means $$b = 0$$.

farhadrclass:
then this line is not that important 
`linesearch && kcopy!(n, x, b) # x ← b` ?

src/minres.jl
```diff
@@ -275,6 +299,21 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
+          stats.solved = true
+          stats.niter = iter
+          stats.inconsistent = false
+          stats.timer = start_time |> ktimer
+          stats.status = "nonpositive curvature"
+          iter == 1 && kcopy!(n, x, b)  # x ← b # we start  at iteration 1
+          solver.warm_start = false
+          return solver
+        end
+      end
```
dpo, line 315:
```suggestion
      # Check for nonpositive curvature
      if linesearch
        cγ = cs * γbar
        if cγ ≥ 0 # or eps(T)
          (verbose > 0) && @printf(iostream, "nonpositive curvature detected:  cs * γbar = %e\n", cγ)
          stats.solved = true
          stats.niter = iter
          stats.inconsistent = false
          stats.timer = start_time |> ktimer
          stats.status = "nonpositive curvature"
          iter == 1 && kcopy!(n, x, b)  || kcopy!(n, x, r)
          solver.warm_start = false
          return solver
        end
      end
```

dpo: It’s a little bit more complicated because the nonpositive curvature direction is the residual of the previous iteration. Currently, MINRES does not form the residual vector; it only updates its norm. So you need to allocate an extra vector to recur the residual vector when `linesearch == true`.

dpo (changes requested):

src/minres.jl
dpo: Also make sure that `linesearch` and `warm_start` are not simultaneously `true`.

test/test_minres.jl
```diff
@@ -69,6 +69,27 @@
...
 
+      # Test linesearch
+      A, b = symmetric_indefinite(FC=FC)
+      x, stats = minres(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+
+      # Test Linesearch which would stop on the first call since A is negative definite
+      A, b = symmetric_indefinite(FC=FC; shift = 5)
+      x, stats = minres(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
```
dpo, line 80: It would be useful to verify that the solution returned is indeed a direction of nonpositive curvature.

farhadrclass:
@dpo 

I've updated the code and status to include a flag for non-positive curvature and line search.  

- If line search is enabled and we detect a non-positive or zero curvature, we compute `rk` and return it.  
- I'm not entirely sure about the scaling—could you review my logic?  
- I also need to rethink some tests and update the test cases. Additionally, I need to add tests for the new statistics.  
- allocate an extra vector to recur the residual vector when `linesearch == true`.
- For now, I suggest introducing a **conjugate stats struct**, as we'll reuse it for both the **CR** , **MINRES**, and  **CG solvers** soon.  There we have 2 new flags 
 - nonpositive curv
 - linesearch 
 User then can check them and if both are true, take advantage of the linesearch 
 

Additionally, in **MINRES**, we currently use `r1` and `r2`. I propose using `rk` as the residual when line search is enabled.  

Let me know your thoughts!

### **To-Do List:**  
1. Add tests for statistics.  
2. Verify that the returned solution is indeed a direction of non-positive curvature.  
3. From what I recall in the last PR, @amontoison  mentioned he would handle ensuring that line search and warm start are not simultaneously enabled.  
4. I suggest (based on what @dpo  suggested) I moved the rk to stats and then return x in all cases and user can decide which one to use if needed, however the allocation would be needed if we move if to stats and any suggestion? (I will send the new updated code shortly)

src/krylov_stats.jl
```diff
@@ -47,6 +47,58 @@ function copyto!(dest :: SimpleStats, src :: SimpleStats)
...
+- nonposi_curv: when a non-positive curvature is detected
+- linesearch: when a line search is performed
+- inconsistent
+- residuals
+- Aresiduals
+- Acond
+- timer
+- status
+"""
+mutable struct conStats{T} <: KrylovStats{T}
```
amontoison, line 68: I am not a big fan of a new structure. Is it mandatory?
`conStats` is a not a good name.

src/minres.jl
```diff
@@ -283,6 +319,16 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
       ϕ = cs * ϕbar
       ϕbar = sn * ϕbar
 
+      if linesearch
+        # calculating the residual rk = sn*sn * rk - ϕbar * cs   * v
+        sn2 = sn * sn
+        kscal!(n, sn2, rk )  # rk = sn2 * rk
+        ϕ_c = -ϕbar * cs 
+        kaxpy!(n, ϕ_c, v, rk)   # rk = rk + ϕ_c * v
+        rk = sn*sn * rk - ϕbar * cs   * v
```
amontoison, line 328: This update is not in-place.

src/krylov_stats.jl
dpo: @amontoison In solvers where `linesearch == true` is allowed, when we detect a direction of negative curvature, we need to return both the solution `x` (which is typically the previous Krylov search direction, and which was not a direction of negative curvature), and a direction of negative curvature (the current search direction in CG or the previous residual in CR and MINRES). Do you prefer that we add a field to the `SimpleStats`, that we use a new `KrylovStats` structure, or that we add a field to the solver object to store the negative curvature direction?

src/minres.jl
```diff
@@ -18,13 +18,15 @@
 # Dominique Orban, <dominique.orban@gerad.ca>
 # Brussels, Belgium, June 2015.
 # Montréal, August 2015.
+#
+# Liu, Yang, and Fred Roosta. "MINRES: from negative curvature detection to monotonicity properties." SIAM Journal on Optimization 32, no. 4 (2022): 2636-2661.
```
dpo, line 22: Move this up and explain what we use from this reference.

src/minres.jl
```diff
@@ -275,6 +293,24 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
+      if linesearch
+        cγ = cs * γbar
+        if cγ ≥ 0
+          (verbose > 0) && @printf(iostream, "nonpositive curvature detected:  cs * γbar = %e\n", cγ)
+          stats.solved = true
+          stats.niter = iter
+          stats.inconsistent = false
+          stats.timer = start_time |> ktimer
+          stats.status = "nonpositive curvature"
+          iter == 1 && kcopy!(n, x, r1)
```
dpo, line 306: `kcopy!(n, x, b)` ?

src/krylov_stats.jl
farhadrclass: **@amontoison** The main thing for me is having a flag to indicate whether a negative curve is detected so that the user can decide what to do next.  

I see four possible approaches :  

1. **New stats structure** – This would involve creating a new struct to store the flag, but I’m not a big fan of this approach because of the extra complexity and additional testing required.  
2. **Adding fields or flags to the existing struct** – This could work, but it would require significant testing to ensure we don’t break anything.  
3. **A more unified status system (long-term solution)** – Eventually, we could implement a system similar to JSOSolvers, where we use symbols instead of strings and have a dedicated method to set the status. This would also require adding fields to the stats, one for status text and another for a status symbol. However, this is too time-consuming for my current paper, though I’d be happy to contribute to it afterward.  
4. **(My suggested approach for now)** – We add a `solver.neg_cur::Bool`. Since we already have `solver.rk` and `solver.x`, this flag would simply indicate if a negative curve is detected. The allocation cost is minimal, and we can later customize it for CG or CR if needed.  

@dpo and @amontoison, what do you think?

src/minres.jl
farhadrclass: yes I put that there for myself to code the logic but forgot to remove
Thanks

src/minres.jl
```diff
@@ -372,6 +418,7 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
     user_requested_exit && (status = "user-requested exit")
     overtimed           && (status = "time limit exceeded")
 
+    stats.nonposi_curv = zero_resid
```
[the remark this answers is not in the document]
farhadrclass: yes, but also I need to check if zero_resid has happened no? since 0 curve means stats.nonposi_curv no?
maybe I should name it neg_curv then ?

src/krylov_stats.jl
amontoison: But we already have this structure, it is `LanczosStats`:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/krylov_stats.jl#L112-L121

farhadrclass: we are missing some fields:
```julia
  inconsistent :: Bool
  residuals    :: Vector{T}
  Aresiduals   :: Vector{T}
```
which we use in MINRES

amontoison: I should finalize #876 asap such that it is easier for you to know what is already available or not in the statistics.

farhadrclass: That would be great,
in the mean time you good with my suggestion on 

> (My suggested approach for now) – We add a solver.neg_cur::Bool. Since we already have solver.rk and solver.x, this flag would simply indicate if a negative curve is detected. The allocation cost is minimal, and we can later customize it for CG or CR if needed.

src/minres.jl
```diff
@@ -148,12 +154,16 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
 
     # Set up workspace.
     allocate_if(!MisI, solver, :v, S, solver.x)  # The length of v is n
+    allocate_if(linesearch, solver, :rk, S, solver.x)  # The length of rk is n
```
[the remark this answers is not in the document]
farhadrclass: rk_dir ?
I also want to add solver.negcurv::bool so we can have access to it without stats?

src/krylov_stats.jl
dpo: @farhadrclass Just add a field `indefinite` to `SimpleStats` and a field `npc_dir` (non-positive curvature direction) to the MINRES solver object (it will hold what you currently call `rk`).

src/minres.jl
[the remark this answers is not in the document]
dpo: No, the flag in the stats is enough.

src/minres.jl
```diff
@@ -12,19 +12,27 @@
 # This implementation follows the original implementation by
 # Michael Saunders described in
 #
+# Negative curvature handling (following Liu, Yang, and Roosta, "MINRES: from negative curvature 
+# detection to monotonicity properties," SIAM Journal on Optimization 32, no. 4 (2022): 2636–2661):
+#
+# If linesearch is true, and if negative curvature is detected at any iteration k > 0, the solution from iteration k-1 is returned, while preserving the last computed residual as a potential search direction.
```
dpo, line 18:
```suggestion
# If linesearch is true and negative curvature is detected at iteration k > 0, the solution from iteration k-1 is returned, and `solver.npc_dir` contains the last computed residual, which is a direction of nonpositive curvature.
```

src/minres.jl
```diff
@@ -12,19 +12,27 @@
...
+# If negative curvature is detected at first iteration, the method returns the right-hand side (i.e., the negative gradient).
```
dpo, line 19:
```suggestion
# If linesearch is true and negative curvature is detected at iteration k = 0, the right-hand side is returned as solution (i.e., the negative gradient), and `solver.npc_dir` contains [WHAT?].
```

src/minres.jl
```diff
@@ -12,19 +12,27 @@
...
+#
 # C. C. Paige and M. A. Saunders, Solution of Sparse Indefinite Systems of Linear Equations,
 # SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
 #
 # Dominique Orban, <dominique.orban@gerad.ca>
 # Brussels, Belgium, June 2015.
 # Montréal, August 2015.
+#
+
```
dpo, line 28:
```suggestion
```

src/minres.jl
```diff
@@ -54,6 +62,10 @@ A is indefinite.
 
 MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr‖₂.
 
+If `linesearch` is true, and if negative curvature is detected at any iteration `k > 0`, the solution from iteration `k-1` is returned, while preserving the last computed residual as a potential search direction.
+If negative curvature is detected at iteration `0`, the method returns the right-hand side
+(i.e., the negative gradient).
+
```
dpo, line 68: Don’t duplicate this explanation. Remove the one from the comments above, and insert it here.

src/minres.jl
```diff
@@ -88,6 +101,8 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 #### Reference
```
dpo, line 101:
```suggestion
#### References
```

src/minres.jl
```diff
@@ -88,6 +101,8 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
...
 
 * C. C. Paige and M. A. Saunders, [*Solution of Sparse Indefinite Systems of Linear Equations*](https://doi.org/10.1137/0712047), SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
+
```
dpo, line 104:
```suggestion
```

src/minres.jl
```diff
@@ -169,22 +190,30 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
     # Initialize Lanczos process.
     # β₁ M v₁ = b.
     kcopy!(n, r2, r1)  # r2 ← r1
     MisI || mulorldiv!(v, M, r1, ldiv)
+
+    rNorm =  knorm_elliptic(n, r2, r1)  # = ‖r‖
+    history && push!(rNorms, rNorm)
+
+
```
dpo, line 203:
```suggestion
```

Please make sure you don’t introduce extra blank lines.

src/minres.jl
```diff
@@ -169,22 +190,30 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
     β₁ = kdotr(m, r1, v)
     β₁ < 0 && error("Preconditioner is not positive definite")
     if β₁ == 0
       stats.niter = 0
       stats.solved, stats.inconsistent = true, false
       stats.timer = start_time |> ktimer
-      stats.status = "x is a zero-residual solution"
+      stats.status = "b is a zero-curvature directions"
```
dpo, line 210: Undo this change.

src/minres.jl
```diff
@@ -169,22 +190,30 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
       history && push!(rNorms, β₁)
       history && push!(ArNorms, zero(T))
       history && push!(Aconds, zero(T))
       warm_start && kaxpy!(n, one(FC), Δx, x)
       solver.warm_start = false
+      stats.indefinite = true
```
dpo, line 216: You haven’t detected nonpositive curvature at this point, only that x = 0 is a solution.

src/minres.jl
```diff
@@ -283,6 +331,15 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
       ϕ = cs * ϕbar
       ϕbar = sn * ϕbar
 
+      if linesearch
+        # calculating the residual npc_dir  = sn*sn * npc_dir  - ϕbar * cs   * v
```
dpo, line 335:
```suggestion
        # calculate the residual npc_dir  = sn*sn * npc_dir  - ϕbar * cs   * v
```

src/minres.jl
```diff
@@ -169,22 +190,30 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
       kcopy!(n, r1, b)  # r1 ← b
     end
 
+    linesearch && kcopy!(n, npc_dir , r1)  # npc_dir  ← r1
+
     # Initialize Lanczos process.
     # β₁ M v₁ = b.
     kcopy!(n, r2, r1)  # r2 ← r1
     MisI || mulorldiv!(v, M, r1, ldiv)
```
dpo, line 198: I think `npc_dir` should contain the preconditioned initial residual here.

dpo (changes requested):

farhadrclass:
**@dpo** Here are the updates.  

I also noticed that we have this file: [[block_minres.jl](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/block_minres.jl)](https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/block_minres.jl), which doesn't seem fully finished. For example, it mentions that "Block-MINRES doesn't support preconditioning yet," and on line 118, there's a comment: `# Temporary buffers -- should be stored in the solver.`  

I believe we should open a separate PR to address these issues and **add linesearch.** What do you think?

amontoison:
See #939 for the buffers in block-MINRES.
It should not a big difference in terms of performance but I can set a higher priority if you use it.

For the preconditioner, it's quite complex to implement it because you need to perform the reduced QR of the block-Krylov basid matrix in an elliptic norm and we can't rely on Householder QR for that (and thus LAPACK).
I could add a fallback with MGS in elliptic norm but don't have bandwidth for that right now.

dpo:
@farhadrclass Don’t worry about line search in block MINRES. That’s an algorithm of a slightly different nature.

src/minres.jl
```diff
@@ -12,19 +12,23 @@
 # This implementation follows the original implementation by
 # Michael Saunders described in
 #
+# Negative curvature handling (following Liu, Yang, and Roosta, "MINRES: from negative curvature 
+# detection to monotonicity properties," SIAM Journal on Optimization 32, no. 4 (2022): 2636–2661):
```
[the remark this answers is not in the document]
dpo: Move this part after the reference below.

src/minres.jl
```diff
@@ -54,6 +58,9 @@ A is indefinite.
 
 MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr‖₂.
 
+If linesearch is true and negative curvature is detected at iteration k > 0, the solution from iteration k-1 is returned, and `solver.npc_dir` contains the last computed residual, which is a direction of nonpositive curvature.
```
dpo, line 61:
```suggestion
```

src/minres.jl
```diff
@@ -54,6 +58,9 @@ A is indefinite.
...
+If linesearch is true and negative curvature is detected at iteration k = 0, the right-hand side is returned as solution (i.e., the negative gradient), and `solver.npc_dir` contains the preconditioned initial residual.
```
dpo, line 62:
```suggestion
```

src/minres.jl
```diff
@@ -68,6 +75,7 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 * `M`: linear operator that models a Hermitian positive-definite matrix of size `n` used for centered preconditioning;
 * `ldiv`: define whether the preconditioner uses `ldiv!` or `mul!`;
 * `window`: number of iterations used to accumulate a lower bound on the error;
+* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If negative curvature is detected at iteration k > 0, the solution of iteration k-1 is returned. If negative curvature is detected at iteration 0, the right-hand side is returned (i.e., the negative gradient);
```
dpo, line 78:
```suggestion
* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch.
If linesearch is true and nonpositive curvature is detected at iteration k = 0, the right-hand side is returned as solution (i.e., the negative gradient), and `solver.npc_dir` contains the preconditioned initial residual.
If linesearch is true and nonpositive curvature is detected at iteration k = 0, the right-hand side is returned as solution (i.e., the negative gradient), and `solver.npc_dir` contains the preconditioned initial residual.
```

src/minres.jl
```diff
@@ -85,9 +93,10 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 * `x`: a dense vector of length `n`;
 * `stats`: statistics collected on the run in a [`SimpleStats`](@ref) structure.
 
-#### Reference
+#### References
 
 * C. C. Paige and M. A. Saunders, [*Solution of Sparse Indefinite Systems of Linear Equations*](https://doi.org/10.1137/0712047), SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
+* Liu, Yang & Roosta, Fred. (2022). A Newton-MR algorithm with complexity guarantees for nonconvex smooth unconstrained optimization. 10.48550/arXiv.2208.07095.
```
dpo, line 99: Fix this reference and make it look like the first one.

src/minres.jl
```diff
@@ -275,6 +296,24 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
+        if cγ ≥ 0
+          (verbose > 0) && @printf(iostream, "nonpositive curvature detected:  cs * γbar = %e\n", cγ)
+          stats.solved = true
+          stats.niter = iter
+          stats.inconsistent = false
+          stats.timer = start_time |> ktimer
+          stats.status = "nonpositive curvature"
+          iter == 1 && kcopy!(n, x, b)
+          solver.warm_start = false
+          # when we use the linesearch and encounter negative curvature, we return the last residual xk but user has access to npc_dir  from solver.npc_dir
```
dpo, line 311:
```suggestion
```

src/minres.jl
```diff
@@ -283,6 +322,15 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
       ϕ = cs * ϕbar
       ϕbar = sn * ϕbar
 
+      if linesearch
+        # calculating the residual npc_dir  = sn*sn * npc_dir  - ϕbar * cs   * v
```
dpo, line 326:
```suggestion
        # compute the residual vector and store it in npc_dir
```

src/minres.jl
```diff
@@ -283,6 +322,15 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
+        sn2 = sn * sn
+        kscal!(n, sn2, npc_dir  )  # npc_dir  = sn2 * npc_dir 
+        ϕ_c = -ϕbar * cs 
+        kaxpy!(n, ϕ_c, v, npc_dir )   # npc_dir  = npc_dir  + ϕ_c * v
```
dpo, line 330:
```suggestion
        kscal!(n, sn * sn, npc_dir  )  # npc_dir  = sn * sn * npc_dir 
        kaxpy!(n, -ϕbar * cs, v, npc_dir )   # npc_dir  = npc_dir  - ϕbar * cs * v
```

src/minres.jl
```diff
@@ -283,6 +322,15 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
...
+      end
+      
+
```
dpo, line 333:
```suggestion
```

test/test_minres.jl
```diff
@@ -68,6 +68,48 @@
...
+      minres!(solver, A, b, linesearch=true)
+      x, stats, npc_dir = solver.x, solver.stats, solver.npc_dir
+      @test stats.status == "nonpositive curvature"
+      @test stats.indefinite == true
+      # Verify that the returned direction indeed exhibits nonpositive curvature.
+      # For both real and complex cases, ensure to take the real part.
+      curvature = real(dot(npc_dir, A * npc_dir))
+      @test curvature <= 0
+
+
```
dpo, line 85:
```suggestion
```

test/test_minres.jl
```diff
@@ -68,6 +68,48 @@
...
+      # Test Linesearch which would stop on the first call since A is negative definite
+      A, b = symmetric_indefinite(FC=FC; shift = 5)
+      x, stats = minres(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+      @test stats.niter == 1 # in Minres they add 1 to the number of iterations first step
+      @test all(x .== b)
+      @test stats.solved == true
+      @test stats.indefinite == true
```
dpo, line 93: Verify that you have an npc direction.

test/test_minres.jl
```diff
@@ -68,6 +68,48 @@
...
+
+      # Test when b^TAb=0 and linesearch is true
+      A, b = system_zero_quad(FC=FC)
+      x, stats = minres(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+      @test all(x .== b)
+      @test stats.solved == true
+      @test stats.indefinite == true
+
+
```
dpo, line 103:
```suggestion
```

dpo (changes requested):
I have already asked to remove duplicate blank lines and to verify that you have found an npc direction in the unit tests several times.

src/minres.jl
farhadrclass: so I remove it since we have same explanation on linesearch or keep one ?

src/minres.jl
farhadrclass: @dpo  I just noticed this is repeating so I will reword it

src/minres.jl
```diff
@@ -68,6 +73,8 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 * `M`: linear operator that models a Hermitian positive-definite matrix of size `n` used for centered preconditioning;
 * `ldiv`: define whether the preconditioner uses `ldiv!` or `mul!`;
 * `window`: number of iterations used to accumulate a lower bound on the error;
+* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If `linesearch` is true and negative curvature is detected, the solution depends on the iteration: at  k > 1, the solution from k-1 is returned with `solver.npc_dir` as the last residual; at k = 1, the right-hand side is returned with `solver.npc_dir` as the preconditioned initial residual. [Note that the MINRES solver start at iteration 1, so the first iteration is k = 1];
```
dpo, line 76:
```suggestion
* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If `linesearch` is true and nonpositive curvature is detected, the solution depends on the iteration: at iteration k = 1, the right-hand side is returned with `solver.npc_dir` as the preconditioned initial residual; at iteration k > 1, the solution from iteration k-1 is returned with `solver.npc_dir` as the most recent residual. (Note that the MINRES solver starts at iteration 1, so the first iteration is k = 1);
```

src/minres.jl
```diff
@@ -85,9 +92,11 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 * `x`: a dense vector of length `n`;
 * `stats`: statistics collected on the run in a [`SimpleStats`](@ref) structure.
 
-#### Reference
+#### References
 
 * C. C. Paige and M. A. Saunders, [*Solution of Sparse Indefinite Systems of Linear Equations*](https://doi.org/10.1137/0712047), SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
+
+* Liu, Yang, and Roosta, MINRES: from negative curvature detection to monotonicity properties, SIAM Journal on Optimization, 32(4), pp. 2636--2661, 2022.
```
dpo, line 99: This is still not in the same format as the previous reference. Add a link to the DOI.

src/minres.jl
```diff
@@ -173,6 +188,11 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
     # β₁ M v₁ = b.
     kcopy!(n, r2, r1)  # r2 ← r1
     MisI || mulorldiv!(v, M, r1, ldiv)
+
+    linesearch && kcopy!(n, npc_dir , v)  # npc_dir  ← v; contain the preconditioned initial residual
+    rNorm =  knorm_elliptic(n, r2, r1)  # = ‖r‖
+    history && push!(rNorms, rNorm)
```
dpo, line 194:
```suggestion
```

dpo: This is already done below.

src/minres.jl
```diff
@@ -283,6 +321,12 @@ kwargs_minres = (:M, :ldiv, :λ, :atol, :rtol, :etol, :conlim, :itmax, :timemax,
       ϕ = cs * ϕbar
       ϕbar = sn * ϕbar
 
+      if linesearch
+        # compute the residual vector and store it in npc_dir
+        kscal!(n, sn * sn, npc_dir  )  # npc_dir  = sn * sn * npc_dir 
+        kaxpy!(n, -ϕbar * cs, v, npc_dir )   # npc_dir  = npc_dir  - ϕbar * cs * v
```
dpo, line 327:
```suggestion
        kscal!(n, sn * sn, npc_dir)  # npc_dir  = sn * sn * npc_dir 
        kaxpy!(n, -ϕbar * cs, v, npc_dir)  # npc_dir  = npc_dir  - ϕbar * cs * v
```

test/test_minres.jl
```diff
@@ -68,6 +68,50 @@
...
+      curvature = real(dot(npc_dir, A * npc_dir))
+      @test curvature <= 0
+
+      # Test when b^TAb=0 and linesearch is true
+      A, b = system_zero_quad(FC=FC)
+      x, stats = minres(A, b, linesearch=true)
+      @test stats.status == "nonpositive curvature"
+      @test all(x .== b)
+      @test stats.solved == true
+      @test stats.indefinite == true
```
dpo, line 104: Test that `real(dot(npc_dir, A * npc_dir)) ≈ 0` ?

test/test_minres.jl
```diff
@@ -68,6 +68,50 @@
...
+
+      # Test if warm_start and linesearch are both true, it should through an error
```
dpo, line 106:
```suggestion
      # Test if warm_start and linesearch are both true, it should throw an error
```

test/test_minres.jl
```diff
@@ -68,6 +68,50 @@
...
+      A, b = symmetric_indefinite(FC=FC)
+      @test_throws MethodError minres(A, b, warm_start = true, linesearch = true)      
+      
+      # Test the constructor warm_start and linesearch when both are true, it should through an error 
+       # Test linesearch
+       A, b = symmetric_indefinite(FC=FC)
+       solver = MinresSolver(A, b)
+       @test_throws MethodError minres!(solver, A, b, linesearch=true, warm_start = true)
```
dpo, line 114: This is already tested by the previous test, isn’t it?

dpo (changes requested):

src/minres.jl
```diff
@@ -85,9 +92,11 @@ MINRES produces monotonic residuals ‖r‖₂ and optimality residuals ‖Aᴴr
 * `x`: a dense vector of length `n`;
 * `stats`: statistics collected on the run in a [`SimpleStats`](@ref) structure.
 
-#### Reference
+#### References
 
 * C. C. Paige and M. A. Saunders, [*Solution of Sparse Indefinite Systems of Linear Equations*](https://doi.org/10.1137/0712047), SIAM Journal on Numerical Analysis, 12(4), pp. 617--629, 1975.
+
+* Y. Liu and F. Roosta, [*MINRES: From Negative Curvature Detection to Monotonicity Properties*](https://doi.org/10.1137/21M143666X), *SIAM Journal on Optimization*, 32(4), pp. 2636–2661, 2022.
```
dpo, line 99:
```suggestion
* Y. Liu and F. Roosta, [*MINRES: From Negative Curvature Detection to Monotonicity Properties*](https://doi.org/10.1137/21M143666X), SIAM Journal on Optimization, 32(4), pp. 2636--2661, 2022.
```

dpo:
Thank you.

PR #970, opened 2025-02-25
Fix Negative Curvature Status Overriding in CR.jl

farhadrclass:
**Description:**  
This PR addresses an issue in `CR.jl` where the "nonpositive curvature" status was never set as intended. The problem arose from the current evaluation order:

```julia
solved = resid_decrease || npcurv || on_boundary
...
npcurv  && (status = "nonpositive curvature")
solved  && (status = "solution good enough given atol and rtol")
```

Because `solved` is defined as `resid_decrease || npcurv || on_boundary`, even when `npcurv` is `true`, `solved` also becomes `true`, leading to the `"solution good enough given atol and rtol"` status being assigned afterward. This inadvertently overrides the negative curvature status, meaning the `"nonpositive curvature"` status would never actually be set.

**Changes Made:**  
- **Reordered and Refactored Conditions:**  
  Modified the logic so that the negative curvature condition (`npcurv`) is checked and assigned **before** the general `solved` condition. This ensures that if negative curvature is detected, the `"nonpositive curvature"` status takes precedence.

### DIFF
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -402,8 +402,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
     # Termination status
     tired               && (status = "maximum number of iterations exceeded")
     on_boundary         && (status = "on trust-region boundary")
-    npcurv              && (status = "nonpositive curvature")
     solved              && (status = "solution good enough given atol and rtol")
+    npcurv              && (status = "nonpositive curvature")
     user_requested_exit && (status = "user-requested exit")
     overtimed           && (status = "time limit exceeded")
 
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -335,7 +335,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
 
       kaxpy!(n, α, p, x)
       xNorm = knorm(n, x)
-      xNorm ≈ radius && (on_boundary = true)
+      xNorm ≈ radius > 0 && (on_boundary = true)
       kaxpy!(n, -α, Mq, r)  # residual
       if MisI
         rNorm² = kdotr(n, r, r)
@@ -401,11 +401,11 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
 
     # Termination status
     tired               && (status = "maximum number of iterations exceeded")
-    on_boundary         && (status = "on trust-region boundary")
     solved              && (status = "solution good enough given atol and rtol")
     user_requested_exit && (status = "user-requested exit")
     overtimed           && (status = "time limit exceeded")
     npcurv              && (status = "nonpositive curvature")
+    on_boundary         && (status = "on trust-region boundary")
 
     # Update x
     warm_start && kaxpy!(n, one(FC), Δx, x)
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -403,9 +403,9 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
     tired               && (status = "maximum number of iterations exceeded")
     on_boundary         && (status = "on trust-region boundary")
     solved              && (status = "solution good enough given atol and rtol")
-    npcurv              && (status = "nonpositive curvature")
     user_requested_exit && (status = "user-requested exit")
     overtimed           && (status = "time limit exceeded")
+    npcurv              && (status = "nonpositive curvature")
 
     # Update x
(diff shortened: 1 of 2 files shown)


src/cr.jl
```diff
@@ -402,8 +402,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
     # Termination status
     tired               && (status = "maximum number of iterations exceeded")
     on_boundary         && (status = "on trust-region boundary")
-    npcurv              && (status = "nonpositive curvature")
     solved              && (status = "solution good enough given atol and rtol")
+    npcurv              && (status = "nonpositive curvature")
```
dpo, line 406: Unit test please.

dpo (changes requested):

src/cr.jl
dpo: Also, put this test last.

farhadrclass:
One side note I noticed, is that we never test CR with radius > 0, I think we need a bug for that?

test/test_cr.jl
```diff
@@ -67,6 +67,11 @@
       x, stats = cr(A, b, linesearch=true)
       @test stats.status == "nonpositive curvature"
 
+      #test nonpositive curvature when radius > 0
+      A, b = symmetric_indefinite(FC=FC, shift = 5)
+      x, stats = cr(A, b, radius = one(Float64))
+      @test stats.status == "nonpositive curvature"
```
dpo, line 73: Right, but in trust-region methods, we would want “on trust-region boundary” to be the final status.

farhadrclass: @dpo so the issue here is that we want “on trust-region boundary”  be the status?
In CR the flag is only set once
```julia
 elseif pAp > 0 && ρ > 0  # no negative curvature
          (verbose > 0) && @printf(iostream, "positive curvatures along p and r. pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
          α = ρ / kdotr(n, q, Mq)
          if α ≥ t1
            α = t1
            on_boundary = true
          end
```
which is not the case in this test, since A is a diagonal with very -ve values

test/test_cr.jl
```diff
@@ -67,10 +67,10 @@
       x, stats = cr(A, b, linesearch=true)
       @test stats.status == "nonpositive curvature"
```
dpo, line 68: Check that `dot(x, A * x) ≤ 0`.

test/test_cr.jl
```diff
@@ -67,10 +67,10 @@
...
 
-      #test nonpositive curvature when radius > 0
+      #test on trust-region boundary when radius > 0
       A, b = symmetric_indefinite(FC=FC, shift = 5)
       x, stats = cr(A, b, radius = one(Float64))
-      @test stats.status == "nonpositive curvature"
+      @test stats.status == "on trust-region boundary"
```
dpo, line 73: Check that `dot(x, A * x) ≤ 0` and `norm(x) ≈ 1`.

dpo (changes requested):

farhadrclass:
@dpo 
When running the following test:

```julia
# Test linesearch
A, b = symmetric_indefinite(FC=FC)
x, stats = cr(A, b, linesearch=true)
@test stats.status == "nonpositive curvature"
@test real(dot(x, A * x)) ≤ 0   # <-- this test fails

# Test on trust-region boundary when radius > 0
A, b = symmetric_indefinite(FC=FC, shift=5)
x, stats = cr(A, b, radius=one(Float64))
@test stats.status == "on trust-region boundary"
@test real(dot(x, A * x)) ≤ 0
@test norm(x) ≈ 1.0
```

The first test is failing while the second one passes. I suspect this is because when `linesearch` is enabled, the solver returns the final `x` instead of the one that produced the negative curvature.

Once the PR for MINRES is accepted, I plan to return `npc_dir` for the CR solver and update the tests accordingly.

test/test_cr.jl
```diff
@@ -66,6 +66,14 @@
       A, b = symmetric_indefinite(FC=FC)
       x, stats = cr(A, b, linesearch=true)
       @test stats.status == "nonpositive curvature"
+      # @test real(dot(x, A * x)) ≤ 0
+
+      # Test on trust-region boundary when radius > 0
+      A, b = symmetric_indefinite(FC=FC, shift = 5)
+      x, stats = cr(A, b, radius = one(Float64))
+      @test stats.status == "on trust-region boundary"
+      @test real(dot(x, A * x)) ≤ 0
```
dpo, line 75: We will update this PR after #969 because we will then have access to the nonpositive curvature direction.

farhadrclass:
@dpo   I think we should close this and follow https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/985

PR #983, opened 2025-03-23
Add a generic interface for Krylov.jl

### LINKED ISSUE #979: Add a generic constructor for KrylovSolver and BlockKrylovSolver
It will be great to have function similar to `solve!` such that we can create a workspace based on a symbol or a string of solver.
For example:
```julia
function KrylovSolver(solver::Symbol, args..., kwargs...)
  if solver == :cg
     CgSolver(args..., kwargs...)
  elseif ...
  ...
  else
     error("Solver $solver is not supported by Krylov.jl")
end
```
We just need to be smart and find a way to use multiple dispatch.
As example; it will be easier to switch between Krylov methods for an optimization solver: 
```julia
function my_optimization_solver(linear_solver=:cg, args...; kwargs...)
   ...
   krylov_solver = KrylovSolver(linear_solver, ...)
   # Create a workspace for the optimization solver
   workspace = ...
   return workspace
end
```

amontoison:
close #979

- Remove the internal dictionaries `BLOCK_KRYLOV_SOLVERS` and `KRYLOV_SOLVERS`.
- Unexport the abstract types `BlockKrylovSolver`, `KrylovSolver`, and `KrylovStats`.
- Polish a few docstrings.
- Rename `solve!` to `krylov_solve!` (fix the conflict with `SolverCore.jl`).
- Add the functions `krylov_workspace` and `krylov_solve`.
- Add a new documentation page on the generic interface (`generic_interface.md`) -- [preview](https://jso.dev/Krylov.jl/previews/PR983/generic_interface/).
- Update the documentation page `inplace.md`.
- Reduce the default value of `memory` in `block_gmres`.
- Verify with `@inferred` in `test_solvers.jl` that the dispatch works as expected and that the result inference works with the three routines of the generic interface (`krylov_workspace`, `krylov_solve` and `krylov_solve!`).

cc @dpo @michel2323


dpo:
I’m all in favor. Please just make sure to preserve backward compatibility because we use various Krylov solvers inside Trunk and R2N.

amontoison:
The next release will be a breaking one (`0.10.0`), but all packages that use `Krylov.jl` will continue to work smoothly with version `0.9.10` (protected by the compat entry in the `Project.toml`) until we update them for `Krylov.jl` `0.10.0`.

amontoison:
@tmigot
I would like to get your point of view on the naming of the three routines in the generic interface (`krylov_workspace`, `krylov_solve!`, and `krylov_solve`) before merging the PR.
Since they will be used throughout the JSO organization, it is important to choose their names wisely.

docs/src/api.md
```diff
@@ -1,7 +1,7 @@
 ## Stats Types
 
 ```@docs
-Krylov.KrylovStats
+Krylov.Krylov.KrylovStats
```
tmigot, line 4: Krylov.Krylov?

docs/src/generic_interface.md
```diff
@@ -0,0 +1,73 @@
...
+The interface is designed to be common for all methods and contains three routines [`krylov_workspace`](@ref krylov_workspace), [`krylov_solve`](@ref krylov_solve) and [`krylov_solve!`](@ref krylov_solve!).
+They allow to build Krylov workspaces and call both in-place and out-of-place variants of the solvers with a unified API.
+
+```@docs
+krylov_workspace
+krylov_solve!
+krylov_solve
+```
+
+In-place solvers update the workspace `solver`, from which solutions and statistics can be retrieved.
```
tmigot, line 13:
```suggestion
In-place solvers update the workspace `workspace`, from which solutions and statistics can be retrieved.
```
I think it's better to be coherent and use the word workspace everywhere in this case. I think `solver` is just an old habit we had.

src/block_gmres.jl
```diff
@@ -61,6 +61,9 @@ function block_gmres end
 
 where `kwargs` are keyword arguments of [`block_gmres`](@ref).
 
+Note that the keyword argument `memory` is the only exception.
```
tmigot, line 64: the only exception compared to?

src/diom.jl
```diff
@@ -77,8 +77,8 @@ function diom end
 
 where `kwargs` are keyword arguments of [`diom`](@ref).
 
-Note that the `memory` keyword argument is the only exception.
-It's required to create a `DiomSolver` and can't be changed later.
+Note that the keyword argument `memory` is the only exception.
```
tmigot, line 80: same

src/dqgmres.jl
```diff
@@ -77,8 +77,8 @@ function dqgmres end
 
 where `kwargs` are keyword arguments of [`dqgmres`](@ref).
 
-Note that the `memory` keyword argument is the only exception.
-It's required to create a `DqgmresSolver` and can't be changed later.
+Note that the keyword argument `memory` is the only exception.
```
tmigot, line 80: same

src/fgmres.jl
```diff
@@ -79,8 +79,8 @@ function fgmres end
 
 where `kwargs` are keyword arguments of [`fgmres`](@ref).
 
-Note that the `memory` keyword argument is the only exception.
-It's required to create a `FgmresSolver` and can't be changed later.
+Note that the keyword argument `memory` is the only exception.
```
tmigot, line 82: same

src/fom.jl
```diff
@@ -72,8 +72,8 @@ function fom end
 
 where `kwargs` are keyword arguments of [`fom`](@ref).
 
-Note that the `memory` keyword argument is the only exception.
-It's required to create a `FomSolver` and can't be changed later.
+Note that the keyword argument `memory` is the only exception.
```
tmigot, line 75: same

test/test_mp.jl
```diff
@@ -1,65 +1,71 @@
-@testset "mp" begin
+function test_mp(FC)
```
tmigot, line 1: It's usually not a good practice in Julia to stuff all the tests in a function. @testset makes it easier to trace back to which test fails.

test/runtests.jl
```diff
@@ -1,5 +1,5 @@
 using Krylov, LinearAlgebra, SparseArrays, Printf, Random, Test
-import Krylov: KRYLOV_SOLVERS, solve!
+import Krylov: KrylovSolver, BlockKrylovSolver
```
tmigot, line 2: Tbh I find slightly weird the fact that these two are not exported by Krylov.jl

src/krylov_solve.jl
```diff
@@ -3,23 +3,52 @@
...
-# - `cg!(solver, A, b, x0; kwargs...)`
 # - `cg(A, b, x0; kwargs...)`
-# - `solve!(solver, A, b, x0; kwargs...)`
+# - `cg!(solver, A, b, x0; kwargs...)`
+# - `krylov_solve(Val{:cg}(), A, b, x0; kwargs...)`
+# - `krylov_solve!(solver, A, b, x0; kwargs...)`
+#
+# Generic constructors for each KrylovSolver and BlockKrylovSolver are also defined using metaprogramming.
+# The function `krylov_workspace` uses the first argument, `Val{solver}()`, where `solver` is a symbol used to dispatch to the appropriate method.
+# For example, the three constructors for `CgSolver` can be called using the following syntax:
```
tmigot, line 18:
```suggestion
# For example, the three constructors for `CgWorkspace` can be called using the following syntax:
```

src/krylov_solve.jl
```diff
@@ -3,23 +3,52 @@
...
+export krylov_workspace, krylov_solve, krylov_solve!
+
+"""
+    krylov_workspace(Val{method}(), args...; kwargs...)
+
+Generic function that dispatches to the appropriate workspace constructor for each subtype of [`KrylovSolver`](@ref) and [`BlockKrylovSolver`](@ref).
+The first argument `Val{method}()`, where `method` is a symbol (such as `:cg`, `:gmres` or `:block_minres`), specifies the (block) Krylov method for which a workspace is desired.
+The returned workspace can later be used by [`krylov_solve!`](@ref) to execute the (block) Krylov method in-place.
+"""
+function krylov_workspace end
```
tmigot, line 36: An idea also is to handle the case where the first argument in `Val{method}()` contains a method that does not exist and provide a better message than just Not implemented.

src/krylov_solve.jl
```diff
@@ -76,10 +124,12 @@ for (workspace, krylov, args, def_args, optargs, def_optargs, kwargs, def_kwargs
         solver.stats.timer += elapsed_time
         return results(solver)
       end
+
+      krylov_solve(::Val{Symbol($krylov)}, $(def_args...); $(def_kwargs...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}} = $(krylov)($(args...); $(kwargs...))
     end
   elseif krylov in (:diom, :dqgmres, :fom, :gmres, :fgmres, :gpmr)
     @eval begin
-      function $(krylov)($(def_args...); memory::Int=20, $(def_kwargs...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}}
+      function $(krylov)($(def_args...); memory::Int = 20, $(def_kwargs...)) where {T <: AbstractFloat, FC <: FloatOrComplex{T}}
```
tmigot, line 132: These magic numbers 20 in the meta-programming might be easier to track or document in a constant somewhere !?

tmigot (changes requested):

tmigot:
> @tmigot I would like to get your point of view on the naming of the three routines in the generic interface (`krylov_workspace`, `krylov_solve!`, and `krylov_solve`) before merging the PR. Since they will be used throughout the JSO organization, it is important to choose their names wisely.

Yes, this is tricky. I guess your motivation was to avoid the potential conflict import of solve and solve! that are also other packages?
At the same time, it's slightly less intuitive to have to write `krylov_` all the time.

We use to hack Krylov.solve! here https://github.com/JuliaSmoothOptimizers/LLSModels.jl/blob/main/src/solve_krylov.jl for linear least squares. How would you handle that?

amontoison:
What do you think of `kworkspace`, `ksolve` and `ksolve!` ?

dpo:
I don’t see anything wrong with writing Krylov.solve! when there is a potential ambiguity.

amontoison:
If I don't use the prefix `krylov_`, or `k`, I need to keep the routines unexported which lead to always write `Krylov.`.

dpo:
How about depending on SolveCore? We could trim it down if necessary.

amontoison:
Why do you want to add a dependency that we don't need?

dpo:
Because it defines the solve! method used everywhere else in JSO.

amontoison:
You can still have conflicts which other Julia packages that export `solve!` like JuMP.
It's only a partial fix.

src/krylov_solve.jl
amontoison: It is a good idea Tangi!
I could define a function `krylov_solve(::Val{<:Symbol}, args..., kwargs...)` that returns an error with a nice message (this method doesn't or the args / kwargs  provided are not supported by the solver).
It will a generic fallback and I can do the same thing for `krylov_workspace`.

test/runtests.jl
amontoison: It is abstract types, do you think it is still useful in practice?
In `Oceananigans.jl`, we want to define a `KrylovSolver` for solving problems and it conflicts with this exported type.

test/test_mp.jl
amontoison: I agree with you but I got many issues with memory allocation tests in the past because of special scopes in these `testset` that were hybrid between global and local.
I think that they solved that with recent releases of Julia.
I bumped the Julia version from 1.6 to 1.10 so we could remove all these functions now.

src/krylov_solve.jl
amontoison: We got this discussion a long time ago when we introduced the in-place methods and @dpo went to the conclusion that he prefered `CgSolver` than `CgWorkspace`.
Do you use `TronWorkspace` or something similar for optimization methods?
It is a big breaking change if we rename all solvers.

dpo:
At least, things will be consistent within JSO.

amontoison:
What you think of a package extension for Krylov in `SolverCore.jl`?
Because `solve!` is defined in `SolverCore`, an extension for Krylov can do an alias `SolverCore.solve!(...) = Krylov.krylov_solve!(...)` and any package that uses `Krylov.jl` and `SolverCore.jl` in the same time can rely on the same function `solve!`.
It gives the flexibility to use both `solve!` and `krylov_solve!` in the JSO ecosystem.
I also don't want to remove the prefix `krylov_` for the function `krylov_workspace` because `workspace` is more a name of a variable and I would like to be consistent with the two other routines of the generic interface.

amontoison:
@dpo --> https://github.com/JuliaSmoothOptimizers/SolverCore.jl/pull/88
If the user uses any packages that depends on `SolverCore.jl` (inside or outside JSO), then `solve!` can be also used.
```julia
julia> using Krylov

julia> A = rand(10,10)
10×10 Matrix{Float64}:
 0.0481886  0.499861  0.630684   0.889681   0.193121  0.541846   0.375265   0.262264   0.504869  0.72162
 0.0132571  0.44793   0.351687   0.526091   0.404355  0.793369   0.418777   0.103955   0.110093  0.191957
 0.239654   0.731971  0.17955    0.552337   0.296913  0.991215   0.0197108  0.977934   0.158278  0.341387
 0.292721   0.398115  0.659456   0.851409   0.491268  0.210392   0.110309   0.76524    0.635155  0.604136
 0.762262   0.818206  0.0959231  0.946439   0.236548  0.0449395  0.475013   0.88021    0.615216  0.324055
 0.15267    0.209481  0.857071   0.574549   0.128105  0.176248   0.251892   0.0285661  0.240964  0.0832387
 0.936432   0.803625  0.594675   0.182967   0.541813  0.525618   0.162202   0.708682   0.670923  0.36341
 0.185657   0.531753  0.297099   0.0567471  0.37648   0.416897   0.620143   0.0345962  0.207503  0.428115
 0.199313   0.239673  0.385852   0.10287    0.487413  0.335752   0.955628   0.739181   0.400595  0.397653
 0.257873   0.593973  0.182434   0.393184   0.708361  0.797949   0.751253   0.330009   0.321804  0.630196

julia> b = rand(10)
10-element Vector{Float64}:
 0.7837834085844175
 0.31500026993136987
 0.815330286371816
 0.30147821027766153
 0.6644592603333351
 0.06178820146617303
 0.4443295000896076
 0.030528227908818928
 0.5790167405231206
 0.5439416608009147

julia> solver = CgSolver(A, b)
┌──────────────────┬────────────────────┬──────────────────┐
│          CgSolver│           nrows: 10│         ncols: 10│
├──────────────────┼────────────────────┼──────────────────┤
│Precision: Float64│   Architecture: CPU│Storage: 363 bytes│
├──────────────────┼────────────────────┼──────────────────┤
│         Attribute│                Type│              Size│
├──────────────────┼────────────────────┼──────────────────┤
│                 m│               Int64│           8 bytes│
│                 n│               Int64│           8 bytes│
│                 x│     Vector{Float64}│          80 bytes│
│                 r│     Vector{Float64}│          80 bytes│
│                 p│     Vector{Float64}│          80 bytes│
│                Ap│     Vector{Float64}│          80 bytes│
│        warm_start│                Bool│            1 byte│
│             stats│SimpleStats{Float64}│          26 bytes│
└──────────────────┴────────────────────┴──────────────────┘

SimpleStats
 niter: 0
 solved: false
 inconsistent: false
 indefinite: false
 residuals: []
 Aresiduals: []
 κ₂(A): []
 timer: 0.00μs
 status: unknown

julia> krylov_solve!(solver, A, b)
julia> solve!(solver, A, b)
ERROR: UndefVarError: `solve!` not defined in `Main`
Suggestion: check for spelling errors or missing imports.
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1

julia> using MadNLP
julia> solve!(solver, A, b)
┌──────────────────┬────────────────────┬──────────────────┐
│          CgSolver│           nrows: 10│         ncols: 10│
├──────────────────┼────────────────────┼──────────────────┤
│Precision: Float64│   Architecture: CPU│Storage: 393 bytes│
├──────────────────┼────────────────────┼──────────────────┤
│         Attribute│                Type│              Size│
├──────────────────┼────────────────────┼──────────────────┤
│                 m│               Int64│           8 bytes│
│                 n│               Int64│           8 bytes│
│                 x│     Vector{Float64}│          80 bytes│
│                 r│     Vector{Float64}│          80 bytes│
│                 p│     Vector{Float64}│          80 bytes│
│                Ap│     Vector{Float64}│          80 bytes│
│        warm_start│                Bool│            1 byte│
│             stats│SimpleStats{Float64}│          56 bytes│
└──────────────────┴────────────────────┴──────────────────┘

SimpleStats
 niter: 20
 solved: false
 inconsistent: false
 indefinite: false
 residuals: []
 Aresiduals: []
 κ₂(A): []
 timer: 35.20μs
 status: maximum number of iterations exceeded

**Comment by tmigot** [comment]
I agree that the match is between krylov_solve and Krylov.solve!. Personally, if the latter is possible I would prefer this.
I found it weird that in the same code you have to do `using Krylov` and then repeat `krylov` every time you use one of the function. Having both `solve!` and `krylov_solve` is not really a simplifying option.
`solve!` if it works is making the code simpler.

@amontoison you have a point with SolverCore, it should not depend directly from NLPModels, and I will open a PR to make it a package extension. So that SolverCore should really have minimal dependencies as it would just define the `solve!` function and `AbstractExecutionStats`, `AbstractOptimizationSolver`, `AbstractSolver ` structures. We could also add a `workspace` function as well to adapt your needs in this package.
Then, being under the umbrella of SolverCore would avoid any warning in JSO. There will still be warnings outside with other packages, but then it's ok to write Krylov.solve!.

I think originally we coded SolverCore with NLPModels in mind but it doesn't have to be.

Regarding https://github.com/JuliaSmoothOptimizers/SolverCore.jl/pull/88/files I don't think you can make Krylov a (optional) dependency of SolverCore as it is the interface.

In conclusion, I am just worried that `krylov_solve` would add complexity for the user.

**Comment by amontoison** [comment]
It is more or less the same, do we want to do some type piracy and define `SolverCore.solve!` in Krylov.jl or extend Solver.solve! for objects of Krylov.jl with the package extension.

In theory the second approach is what we should do now. We used the first approach because package extensions arrived only recently.

**Comment by tmigot** [comment]
> It is more or less the same, do we want to do some type piracy and define `SolverCore.solve!` in Krylov.jl or extend Solver.solve! for objects of Krylov.jl with the package extension.
> 
> In theory the second approach is what we should do now. We used the first approach because package extensions arrived only recently.

It is more a question of responsability. SolverCore is defining an interface of solver usage. So, it's not supposed to be dependent of solvers, while extending SolverCore.solve! in Krylov is just implementing the interface.
To the contrary JSOSuite is a wrapper of solvers, so it would be fine there.

If you think it's blocking to have SolverCore as a dependency here, we can always keep the current behavior, which is Krylov defines solve! and whenever we use it then we have to write Krylov.solve! which is OK I think.

File: docs/src/api.md (hunk 1 above)
Comment (amontoison) [inline docs/src/api.md:4]:
```suggestion
Krylov.KrylovStats
```

docs/src/generic_interface.md
```diff
@@ -0,0 +1,73 @@
...
+The interface is designed to be common for all methods and contains three routines [`krylov_workspace`](@ref krylov_workspace), [`krylov_solve`](@ref krylov_solve) and [`krylov_solve!`](@ref krylov_solve!).
+They allow to build Krylov workspaces and call both in-place and out-of-place variants of the solvers with a unified API.
+
+```@docs
+krylov_workspace
+krylov_solve!
+krylov_solve
+```
+
+In-place solvers update the workspace `workspace`, from which solutions and statistics can be retrieved.
```
amontoison, line 13:
```suggestion
In-place solvers update the workspace, from which solutions and statistics can be retrieved.
```

test/test_mp.jl
amontoison: I opened an issue #986

src/krylov_solve.jl
amontoison: I think you are right Tangi, we should use `KrylovWorkspace` instead of `KrylovSolver`.
The next release is breaking so I will also change that.

PR #985, opened 2025-03-27
implementation Line Search with Negative Curvature Detection with direction for CR 

farhadrclass:
The algorithm adapts its behavior upon detecting nonpositive curvature:

    At iteration k = 0: Returns the right-hand side b with solver.npc_dir set to the preconditioned initial residual.

    At iteration k > 0: Returns the solution from iteration k-1 with solver.npc_dir updated to the most recent residual.

This modification ensures that the CR algorithm can be effectively utilized within inexact Newton methods that employ a linesearch strategy.

### DIFF
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -62,7 +62,7 @@ For an in-place variant that reuses memory across solves, see [`cr!`](@ref).
  – at iteration k > 0,
    - if the residual from iteration k-1 is a nonpositive curvature direction but `workspace.p` is not, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 1;
    - if `workspace.p` is a nonpositive curvature direction but the residual is not, `workspace.p` is copied into `stats.npc_dir` and `stats.npcCount` is set to 1;
-   - if both are nonnegative curvature directions, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 2.
+   - if both are nonpositive curvature directions, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 2.
 * `γ`: tolerance to determine that the curvature of the quadratic model is nonpositive;
 * `atol`: absolute stopping tolerance based on the residual norm;
 * `rtol`: relative stopping tolerance based on the residual norm;
@@ -280,11 +280,10 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         if abspAp ≤ γ * pNorm * knorm(n, q)  # pᴴAp ≃ 0
           npcurv = true  # nonpositive curvature
           stats.indefinite = true
-          if iter == 0
+          stats.npcCount = 1
+          if iter > 0
             kcopy!(n, npc_dir, r)
-            stats.npcCount = 1
           else
-            stats.npcCount = 1
             kcopy!(n, npc_dir, p)
           end
           (verbose > 0) && @printf(iostream, "pᴴAp = %8.1e ≃ 0\n", pAp)
@@ -335,13 +334,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         elseif pAp > 0 && ρ < 0
           npcurv = true
           stats.indefinite = true
-          if iter == 0
-            kcopy!(n, npc_dir, r)
-            stats.npcCount = 1
-          else
-            stats.npcCount = 1
-            kcopy!(n, npc_dir, r)
-          end
+          stats.npcCount = 1
+          kcopy!(n, npc_dir, r)
           (verbose > 0) && @printf(iostream, "pᴴAp = %8.1e > 0 and rᴴAr = %8.1e < 0\n", pAp, ρ)
           # q_p is minimal for α_p = rᴴp / pᴴAp
           α = descent ?  min(t1, pr / pAp) : max(t2, pr / pAp)
@@ -359,13 +353,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         elseif pAp < 0 && ρ > 0
           npcurv = true
           stats.indefinite = true
-          if iter == 0
-            kcopy!(n, npc_dir, r)
-            stats.npcCount = 1
-          else
-            stats.npcCount = 1
-            kcopy!(n, npc_dir, p)
-          end
+          stats.npcCount = 1
+          kcopy!(n, npc_dir, p)
           (verbose > 0) && @printf(iostream, "pᴴAp = %8.1e < 0 and rᴴAr = %8.1e > 0\n", pAp, ρ)
           α = descent ? t1 : t2
           tr = min(tr, rNorm² / ρ)
@@ -383,13 +372,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         elseif pAp < 0 && ρ < 0
           npcurv = true
           stats.indefinite = true
-          if iter == 0
-            kcopy!(n, npc_dir, r)
-            stats.npcCount = 1
-          else
-            stats.npcCount = 2
-            kcopy!(n, npc_dir, r)
-          end
+          stats.npcCount = 2
+          kcopy!(n, npc_dir, r)
           (verbose > 0) && @printf(iostream, "negative curvatures along p and r. pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
           α = descent ? t1 : t2
           Δ = -α * pr + tr * rNorm² + (α^2 * pAp - (tr)^2 * ρ) / 2
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -197,8 +197,9 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
       workspace.warm_start = false
       if linesearch || radius > 0
         kcopy!(n, x, p)  # x ← M⁻¹ b
-        kcopy!(n, npc_dir, r)
+        kcopy!(n, npc_dir, p)
         stats.npcCount = 1
+        stats.indefinite = tru
       end
       return workspace
     end
@@ -235,8 +236,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
 
     while ! (solved || tired || user_requested_exit || overtimed)
       if linesearch
-        p_curv = (pAp ≤ γ * pNorm^2)
-        r_curv = (ρ   ≤ γ * rNorm^2)
+        p_curv = pAp ≤ γ * pNorm^2
+        r_curv = ρ   ≤ γ * rNorm^2
         if p_curv || r_curv
           npcurv = true
           (verbose > 0) && @printf(iostream, "nonpositive curvature detected: pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
@@ -248,7 +249,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
           workspace.warm_start = false
           stats.indefinite = true
           if iter == 0
-            kcopy!(n, npc_dir, r)
+            kcopy!(n, npc_dir, p)
             stats.npcCount = 1
           else
             if r_curv
@@ -281,11 +282,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
           npcurv = true  # nonpositive curvature
           stats.indefinite = true
           stats.npcCount = 1
-          if iter > 0
-            kcopy!(n, npc_dir, r)
-          else
-            kcopy!(n, npc_dir, p)
-          end
+          kcopy!(n, npc_dir, p)
           (verbose > 0) && @printf(iostream, "pᴴAp = %8.1e ≃ 0\n", pAp)
           if abspr ≤ γ * pNorm * rNorm  # pᴴr ≃ 0
             (verbose > 0) && @printf(iostream, "pᴴr = %8.1e ≃ 0, redefining p := r\n", pr)
@@ -324,7 +321,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
           end
 
         elseif pAp > 0 && ρ > 0  # no negative curvature
-          (verbose > 0) && @printf(iostream, "positive curvatures along p and r. pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
+          (verbose > 0) && @printf(iostream, "positive curvature along p and r. pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
           α = ρ / kdotr(n, q, Mq)
           if α ≥ t1
             α = t1
@@ -374,7 +371,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
           stats.indefinite = true
           stats.npcCount = 2
           kcopy!(n, npc_dir, r)
-          (verbose > 0) && @printf(iostream, "negative curvatures along p and r. pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
+          (verbose > 0) && @printf(iostream, "negative curvature along p and r. pᴴAp = %8.1e and rᴴAr = %8.1e\n", pAp, ρ)
           α = descent ? t1 : t2
           Δ = -α * pr + tr * rNorm² + (α^2 * pAp - (tr)^2 * ρ) / 2
           if Δ > 0
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -198,7 +198,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         kcopy!(n, x, p)  # x ← M⁻¹ b
         kcopy!(n, npc_dir, p)
         stats.npcCount = 1
-        stats.indefinite = tru
+        stats.indefinite = true
       end
       return workspace
     end
@@ -249,6 +249,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
           stats.indefinite = true
           if iter == 0
             kcopy!(n, npc_dir, p)
+            kcopy!(n, x, p)  # x ← M⁻¹ b
             stats.npcCount = 1
           else
             if r_curv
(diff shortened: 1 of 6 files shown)


src/cr.jl
```diff
@@ -51,7 +51,7 @@ M also indicates the weighted norm in which residuals are measured.
 * `M`: linear operator that models a Hermitian positive-definite matrix of size `n` used for centered preconditioning;
 * `ldiv`: define whether the preconditioner uses `ldiv!` or `mul!`;
 * `radius`: add the trust-region constraint ‖x‖ ≤ `radius` if `radius > 0`. Useful to compute a step in a trust-region method for optimization;
-* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If negative curvature is detected at iteration k > 0, the solution of iteration k-1 is returned. If negative curvature is detected at iteration 0, the right-hand side is returned (i.e., the negative gradient);
+* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If `linesearch` is true and nonpositive curvature is detected, the solution depends on the iteration: at iteration k = 0, the right-hand side is returned with `solver.npc_dir` as the preconditioned initial residual; at iteration k > 0, the solution from iteration k-1 is returned with `solver.npc_dir` as the most recent residual;
```
dpo, line 54: Why return the most recent residual in npc_dir when k > 0?

src/cr.jl
```diff
@@ -210,6 +216,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
     overtimed = false
 
     while ! (solved || tired || user_requested_exit || overtimed)
+      linesearch && kcopy!(n, npc_dir , r) # npc_dir ← r; contain the last residual
```
dpo, line 219: You’re doing this operation twice at the first iteration.

src/cr.jl
```diff
@@ -401,12 +409,15 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
...
-    on_boundary         && (status = "on trust-region boundary")
-    npcurv              && (status = "nonpositive curvature")
     solved              && (status = "solution good enough given atol and rtol")
     user_requested_exit && (status = "user-requested exit")
     overtimed           && (status = "time limit exceeded")
+    npcurv              && (status = "nonpositive curvature")
+    on_boundary         && (status = "on trust-region boundary")
 
+    if npcurv && linesearch
+       stats.indefinite = true
```
dpo, line 419: This was already done in the main loop, wasn’t it?

test/test_cr.jl
```diff
@@ -11,6 +11,8 @@
       resid = norm(r) / norm(b)
       @test(resid ≤ cr_tol)
       @test(stats.solved)
+      @test stats.indefinite == false
+
```
dpo, line 15:
```suggestion
```

test/test_cr.jl
```diff
@@ -64,42 +66,77 @@
...
+      solver = CrSolver(A, b)
+      cr!(solver, A, b, linesearch=true)
+      x, stats, npc_dir = solver.x, solver.stats, solver.npc_dir
       @test stats.status == "nonpositive curvature"
+      @test stats.indefinite == true
+      # Verify that the returned direction indeed exhibits nonpositive curvature.
+      # For both real and complex cases, ensure to take the real part.
+      curvature = real(dot(npc_dir, A * npc_dir))
+      @test curvature <= 0
+
```
dpo, line 78: Extra blank lines!!!

dpo (changes requested):

src/cr.jl
farhadrclass: residual is returned no?
I mean the one we just calculated in the iteration k (not k-1)

src/cr.jl
farhadrclass: yes, I need to put it in the last line of the loop or should I use (iter != 0  and linesearch) ....

src/cr.jl
farhadrclass: Curvature Tests
---------------
1. Search-Direction Curvature Test
   At each CR step, compute the step length

       α_i = (r_iᵀ r_i) / (p_iᵀ H p_i)

   Before updating p_i, check

       p_iᵀ H p_i ≤ 0

   If this holds, p_i is a nonpositive‐curvature direction. CR terminates immediately, returning:

       d_curv = p_i

2. Residual Curvature Test
   After the update x_{i+1} = x_i + α_i p_i, compute the new residual

       r_{i+1} = r_i − α_i H p_i

   Then form the next coefficient

       β_i = (r_{i+1}ᵀ r_{i+1}) / (r_iᵀ r_i)

   Before defining the next search direction, check

       r_{i+1}ᵀ H r_{i+1} ≤ 0

   If this holds, r_{i+1} is a nonpositive‐curvature direction and CR halts, returning:

       d_resid = r_{i+1}

Returns
-------
npc_dir : array-like
    The nonpositive‐curvature direction:
      - `p_i` if the search-direction test triggered,
      - `r_{i+1}` if the residual test triggered,
      - otherwise the current residual `r`.

p : array-like
    The (possibly updated) search direction.

stats.num_neg_dir : int
    Number of negative‐curvature directions detected:
      - 0: none
      - 1: one test triggered
      - 2: both tests triggered

farhadrclass:
@dpo  I’ve updated the PR.

I introduced `num_dir` in `stats` to indicate how many negative-curvature directions we’re using. However, I’m not entirely confident in the approach or whether I placed everything correctly (some lines are marked with `#TODO`). Would you be able to take a look now so I can finalize it and then write the test cases?

The idea is that we can detect 0, 1, or 2 negative directions. `npc_dir` is set to either `r` or `p`, and if both directions are needed, we use `solver.p` as well.

Let me know what you think!

src/cr.jl
```diff
@@ -51,7 +51,9 @@ M also indicates the weighted norm in which residuals are measured.
...
 * `radius`: add the trust-region constraint ‖x‖ ≤ `radius` if `radius > 0`. Useful to compute a step in a trust-region method for optimization;
-* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If negative curvature is detected at iteration k > 0, the solution of iteration k-1 is returned. If negative curvature is detected at iteration 0, the right-hand side is returned (i.e., the negative gradient);
+* `linesearch`: if `true` and nonpositive curvature is detected, behavior depends on the iteration.
+ – At k = 0, return the right-hand side with `solver.npc_dir` set to the preconditioned initial residual.
+ – At k > 0, return the solution from iteration k–1 with `solver.npc_dir` set to the detected nonpositive-curvature direction—either the previous search direction if the search-direction test fails, or the latest residual if the residual-curvature test fails. If both test fails, set `solver.npc_dir` to the current residual, update `solver.p` to the most recent search direction, and record the number of negative-curvature directions in `stats.num_neg_dir`;
```
dpo, line 56:
```suggestion
* `linesearch`: if `true` and nonpositive curvature is detected, the behavior depends on the iteration:
 – at iteration k = 0, return the preconditioned initial search direction in `solver.npc_dir`;
 – at iteration k > 0,
   - if the residual from iteration k-1 is a nonpositive curvature direction but `solver.p` is not, the residual is stored in `stats.npc_dir` and `stats.num_neg_dir` is set to 1;
   - if `solver.p` is a nonpositive curvature direction but the residual is not, `solver.p` is copied into `stats.npc_dir` and `stats.num_neg_dir` is set to 1;
   - if both are nonnegative curvature directions, the residual is stored in `stats.npc_dir` and `stats.num_neg_dir` is set to 2.
```

src/cr.jl
```diff
@@ -219,8 +227,11 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
           stats.inconsistent = false
           stats.timer = start_time |> ktimer
           stats.status = "nonpositive curvature"
+          kcopy!(n, npc_dir , r) # npc_dir ← r; contain the last residual
```
dpo, line 230: Here, you're not sure r is nonpositive curvature direction. The test on line 214 is an OR.

src/cr.jl
dpo: This should tell you how to adjust your code.

dpo (changes requested):

src/cr.jl
```diff
@@ -220,16 +234,33 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
...
+              kcopy!(n, npc_dir, r)
+              stats.npcCount = 1
+            elseif p_dir && !r_dir
+              kcopy!(n, npc_dir, p)
+              stats.npcCount = 1
+            else
+              kcopy!(n, npc_dir, r)
+              stats.npcCount = 2
+            end
+          end
```
dpo, line 263: Wouldn't it be shorter and equivalent to do
```julia
if r_curv
  kcopy!(n, npc_dir, r)
  stats.npcCount += 1
end
if p_curv
  stats.npcCount += 1
  rcurv || kcopy!(n, npc_dir, p)
end
```

src/cr.jl
```diff
@@ -263,6 +294,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
             else # case 2
               (verbose > 0) && @printf(iostream, "r is a direction of nonpositive curvature: %8.1e\n", ρ)
               α = tr
+              #TODO what to do in this case?
```
dpo, line 297: This is the case where both p and r are directions of nonnegative curvature. So you can do the same as in the linesearch case (and update the docstring).

test/test_cr.jl
```diff
@@ -63,24 +64,36 @@
...
+      # Iter=0: b^T A b = 0  → zero-curvature at k=0, with γ
+      A, b = system_zero_quad(FC=Float64)   # ensures bᵀ A b == 0
+      solver = CrWorkspace(A, b)
+      cr!(solver, A, b; linesearch=true, γ=γ_test)
+      x, stats, npc_dir = solver.x, solver.stats, solver.npc_dir
+      @test stats.niter == 0
+      @test stats.status == "b is a zero-curvature direction"
+      @test real(dot(npc_dir, A * npc_dir)) ≈ 0
+      
+      # Test when b^TAb=0 and linesearch is true, without γ
```
dpo, line 76: I don't think this test adds anything.

test/test_cr.jl
```diff
@@ -63,24 +64,36 @@
       @test(stats.solved)
 
       # Test linesearch
-      A, b = symmetric_indefinite(FC=FC)
-      x, stats = cr(A, b, linesearch=true)
-      @test stats.status == "nonpositive curvature"
+      # Iter=0: b^T A b = 0  → zero-curvature at k=0, with γ
```
dpo, line 67:
```suggestion
      # Iter=0: bᵀ Ab = 0 → zero-curvature at k=0
```

test/test_cr.jl
```diff
@@ -1,16 +1,17 @@
 @testset "cr" begin
   cr_tol = 1.0e-6
+  γ_test = sqrt(eps(Float64))
```
dpo, line 3: This is the value by default. I don't think it's useful to change it.

test/test_cr.jl
```diff
@@ -63,24 +64,36 @@
       @test(stats.solved)
 
       # Test linesearch
-      A, b = symmetric_indefinite(FC=FC)
-      x, stats = cr(A, b, linesearch=true)
-      @test stats.status == "nonpositive curvature"
+      # Iter=0: b^T A b = 0  → zero-curvature at k=0, with γ
+      A, b = system_zero_quad(FC=Float64)   # ensures bᵀ A b == 0
+      solver = CrWorkspace(A, b)
+      cr!(solver, A, b; linesearch=true, γ=γ_test)
```
dpo, line 70:
```suggestion
      cr!(solver, A, b; linesearch=true)
```

test/test_cr.jl
```diff
@@ -63,24 +64,36 @@
...
+      A, b = system_zero_quad(FC=FC)
+      solver = CrWorkspace(A, b)
+      cr!(solver, A, b, linesearch=true)
+      x, stats, npc_dir = solver.x, solver.stats, solver.npc_dir
+      @test stats.status == "b is a zero-curvature direction"
+      @test all(x .== b)
+      @test stats.solved == true
+      @test real(dot(npc_dir, A * npc_dir)) ≈ 0.0
 
       # Test Linesearch which would stop on the first call since A is negative definite
```
dpo, line 86: Is it negative definite or indefinite?

test/test_cr.jl
```diff
@@ -63,24 +64,36 @@
...
-
-      # Test when b^TAb=0 and linesearch is true
-      A, b = system_zero_quad(FC=FC)
-      x, stats = cr(A, b, linesearch=true)
-      @test stats.status == "b is a zero-curvature direction"
-      @test all(x .== b)
       @test stats.solved == true
+      @test stats.indefinite == true
+      curvature = real(dot(npc_dir, A * npc_dir))
+      @test curvature <= 0
```
dpo, line 96: What should the solution be here? You deleted that part.

test/test_cr.jl
```diff
@@ -89,17 +102,74 @@
...
+      @test stats.npcCount == 1
+      @test real(dot(npc_dir, A*npc_dir)) ≤ cr_tol
+      p1 = solver.p
+      @test real(dot(p1, A*p1)) < 0
+      
+      # Warm-start + linesearch must error
+      A, b = symmetric_indefinite(FC=Float64)
+      @test_throws MethodError cr(A, b; warm_start=true, linesearch=true)
+
+      # npc_dir is not a zero vector, however we don't have γ
```
dpo, line 133: What do you mean, "however we don't have γ"?

test/test_cr.jl
```diff
@@ -89,17 +102,74 @@
...
       cb_n2 = TestCallbackN2(A, b, tol = tol)
-      cr!(workspace, A, b, callback = cb_n2)
-      @test workspace.stats.status == "user-requested exit"
-      @test cb_n2(workspace)
+      cr!(solver, A, b, callback = cb_n2)
+      @test stats.status == "user-requested exit"
+      @test cb_n2(solver)
+      @test_throws TypeError cr(A, b, callback = solver -> "string", history = true)
+
+
```
dpo, line 157:
```suggestion
```

dpo (changes requested):

src/cr.jl
```diff
@@ -184,7 +194,11 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
       stats.status = "b is a zero-curvature direction"
       history && push!(ArNorms, zero(T))
       workspace.warm_start = false
-      linesearch && kcopy!(n, x, b)  # x ← b
+      if linesearch
+        kcopy!(n, x, b)  # x ← b
+        kcopy!(n, npc_dir, r)
+        stats.npcCount = 1
+      end
```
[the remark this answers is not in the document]
farhadrclass: We have kcopy!(n, npc_dir, r), should we also return x=r ?

src/cr.jl
farhadrclass: The issue is that 
`linesearch && (radius > 0) && error("'linesearch' set to 'true' but radius > 0")`

Do you want to have npc_dir without the linesearch as well?

src/cr.jl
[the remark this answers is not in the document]
dpo: No... x will be the descent direction (preconditioned b).

src/cr.jl
dpo: Yes, you can do the same for trust region.

farhadrclass:
added npc_dir logic for when  `radius > 0` and nonpositive curvature is detected, the behavior depends on the iteration and follow similar logic as linesearch;

farhadrclass:
@dpo is the changes I suggested for the radius aligned with what you were considering ?

test/test_cr.jl
```diff
@@ -100,6 +103,134 @@
...
+      curvature = real(dot(npc_dir, A * npc_dir))
+      @test curvature <= 0
+      @test stats.npcCount == 2
+
+      # radius > 0
+      # bᵀ Ab ≈  0 at first iteration
+      A = FC[1.0      1e-10; 1e-10    1e-10]
+      b = FC[0.0, 0.99]
+      solver = CrWorkspace(A, b)
+      cr!(solver, A, b; radius = 10 * one(Float64),  γ = 10.0)
```
dpo, line 153: `radius = 10 * one(FC)` and `γ = 10 * one(FC)`?

test/test_cr.jl
```diff
@@ -100,6 +103,134 @@
...
+      x, stats, npc_dir = solver.x, solver.stats, solver.npc_dir
+      
+      @test stats.npcCount == 1
+      @test stats.status == "on trust-region boundary"
+      @test stats.indefinite == true
+      curvature = real(dot(npc_dir, A * npc_dir))
+      @test curvature <= 0.01 # almost zero curvature
```
dpo, line 160:
```suggestion
      @test real(dot(npc_dir, A * npc_dir)) <= 0.01  # almost zero curvature
```

test/test_cr.jl
farhadrclass: can we have it if FC = Complex?

farhadrclass: @dpo  Here FC = Float64 vs Complex I think

test/test_cr.jl
```diff
@@ -1,236 +1,267 @@
...
-      x, stats, npc_dir = solver.x, solver.stats, solver.npc_dir
-      @test stats.status == "on trust-region boundary"
-      @test norm(x) ≈ 1.0
-      curvature = real(dot(x, A * x))
-      @test curvature <= 0
-
-      # Test on trust-region boundary when radius = 1 and linesearch is true
-      A, b = symmetric_indefinite(FC=FC, shift = 5)
-      @test_throws ErrorException cr(A, b, radius = one(Float64), linesearch = true) 
+    cr_tol = 1.0e-6
```
dpo, line 2: It looks like you screwed up the indentation here. Please fix.

dpo (changes requested):

src/krylov_stats.jl
```diff
@@ -24,6 +25,7 @@ mutable struct SimpleStats{T} <: KrylovStats{T}
   solved       :: Bool
   inconsistent :: Bool
   indefinite   :: Bool
+  npcCount  :: Int
```
amontoison, line 28: Can you align the types?

farhadrclass: yes ofc,
@amontoison  can we include .JuliaFormatter.toml file so when we run the formatter it would always be the same ( We need to run it for Krylov_stats btw)

amontoison: I don't want to use the Julia formatter in this repository because I would need to add custom blocks in many places to protect the code from being reformatted.
I used custom alignments intentionally, especially when using metaprogramming, and I want to keep them.
Also, JuliaFormatter was initially failing in this repository a few years ago, and I didn’t want to spend time fixing it (nor did the other JSO developers).
It was a low priority back then, and it's even more so now.

dpo: We really do need a formatter config here. It's come a long way since you last tried it.

dpo:
PHEW!!!!

PR #990, opened 2025-04-13
Use the suffix Workspace instead of Solver

### LINKED ISSUE #987: Rename the suffix of the workspaces *Solver -> *Workspace
I think `solver` is misleading for the user, as it could be confused with the Krylov method itself.  
 In our case, we only store data for the method, without any options.
`KrylovWorkspace` seems the most appropriate name for now.

amontoison:
close #987


amontoison:
@tmigot 
I think I should add aliases like this one for the workspaces:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/interface.jl#L53-L55
It will help users to switch smoothly to the new API.

What do you think?

docs/src/warm-start.md
```diff
@@ -17,25 +18,27 @@ If the user has an initial guess `x0`, it can be provided directly.
 cg(A, b, x0)
 ```
 
-It is also possible to use the `warm_start!` function to feed the starting point into the solver.
+It is also possible to use the `warm_start!` function to feed the starting point into the workspace.
```
amontoison, line 21:
```suggestion
It is also possible to use the function `warm_start!` to feed the starting point into the workspace.
```

amontoison:
@tmigot 
I will wait your comments before the new release `v0.10` but I merge the PR to be able to do other modifications.

tmigot:
Thanks @amontoison for the changes. To be honest, I am not too fond of the alias. True, it keeps compatibility, but anyway next release will be breaking and I think it is confusing to have the two functions. We can update all the deps of Krylov when upgrading to the new release.
Unless, there is another reason I am missing, we should remove it.

tmigot:
Same I think it makes things more complicated, and I am suspecting we will run into strange errors at some points

PR #991, opened 2025-04-15
Return a nice error if we detect the preconditioner is not SPD in CG

amontoison:
Requested in https://github.com/CliMA/Oceananigans.jl/pull/4387


amontoison:
It seems that they broke something one Mac with the release 1.11.5 of Julia.
I opened an issue: https://github.com/JuliaLang/julia/issues/58132

PR #1006, opened 2025-05-09
fix docs for CSR

oscardssmith:
With
```
A_gpu = CuSparseMatrixCSR(sprand(T, n, n, 0.05) + I)
b_cpu = CuVector(rand(T, n))
julia> x, stats = bicgstab(A_gpu, b_gpu, M=opM)
```
we before got
```
(Float32[1.0198809, 4.0956926, 1.28241, -1.7407427, -2.9416926, -0.32771796, 0.4703165, 5.368891, -3.1246433, -1.7737565  …  0.97625387, 3.7136111, -1.0120717, 2.8906355, -1.1602651, 0.46383825, 2.538051, 1.448916, 2.4476223, -3.592498], SimpleStats
 niter: 54
 solved: false
 inconsistent: false
 indefinite: false
 residuals: []
 Aresiduals: []
 κ₂(A): []
 timer: 48.24ms
 status: breakdown αₖ == 0
)
```
but with this, get
```
(Float32[-0.4150291, 2.639933, -1.3929893, -1.1049638, -1.2597414, 0.0031488854, -1.918298, 5.812737, -1.3443595, -1.2016623  …  1.6825913, 3.0281372, -1.4967421, 3.0308955, -1.1063535, 0.2894782, 1.1083188, -0.1900546, 2.3867438, 1.4211323], SimpleStats
 niter: 66
 solved: true
 inconsistent: false
 indefinite: false
 residuals: []
 Aresiduals: []
 κ₂(A): []
 timer: 57.85ms
 status: solution good enough given atol and rtol
)
```

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -138,8 +138,8 @@ if CUDA.functional()
 
   # Solve Py = x
   function ldiv_ilu0!(P::CuSparseMatrixCSR, x, y, z)
-    ldiv!(z, UnitLowerTriangular(P), x)  # Forward substitution with L
-    ldiv!(y, UpperTriangular(P), z)      # Backward substitution with U
+    ldiv!(z, LowerTriangular(P), x)  # Forward substitution with L
+    ldiv!(y, UnitUpperTriangular(P), z)      # Backward substitution with U
     return y
   end
 


amontoison:
@oscardssmith The breakdown is independent of the good / bad preconditioner. What we should check is if we converge in one iteration if we do a ILU preconditioner of a dense matrix stored in sparse CSR format.

oscardssmith:
> What we should check is if we converge in one iteration with an ILU preconditioner of a dense matrix stored in sparse CSR format.

This is true for the old, but not the new one.

> The breakdown is independent of the good / bad preconditioner

What is it dependent on? Randomly failing to solve seems like a fairly large problem.

amontoison:
> > What we should check is if we converge in one iteration with an ILU preconditioner of a dense matrix stored in sparse CSR format.
> 
> This is true for the old, but not the new one.

Ok, so it means that the old preconditioner is correct.

> > The breakdown is independent of the good / bad preconditioner
> 
> What is it dependent on? Randomly failing to solve seems like a fairly large problem.

Yes, it is why we should never use `bicgstab` :)
Everybody wanted this method implemented in `Krylov.jl` but it can breakdown at any moment.
It is cheap but not robust.

oscardssmith:
In that case, should the docs switch to `gmres` or some other better algorithm?

PR #1011, opened 2025-06-15
implementation Line Search with Negative Curvature Detection with direction for MINRES

farhadrclass:
Similar to https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/985

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -19,6 +19,9 @@
 # Liu, Yang, and Roosta, MINRES: from negative curvature detection to monotonicity properties,
 # SIAM Journal on Optimization, 32(4), pp. 2636--2661, 2022.
 #
+# M-A. Dahito and D. Orban, The Conjugate Residual Method in Linesearch and Trust-Region Methods.
+# SIAM Journal on Optimization, 29(3), pp. 1988--2025, 2019.
+#
 # Dominique Orban, <dominique.orban@gerad.ca>
 # Brussels, Belgium, June 2015.
 # Montréal, August 2015.
@@ -80,10 +83,10 @@ For an in-place variant that reuses memory across solves, see [`minres!`](@ref).
 * `ldiv`: define whether the preconditioner uses `ldiv!` or `mul!`;
 * `window`: number of iterations used to accumulate a lower bound on the error;
 * `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If `true` and nonpositive curvature is detected, the behavior depends on the iteration:
- – at iteration k = 0, return the preconditioned initial search direction in `workspace.npc_dir`;
- – at iteration k > 0,
-   - if the residual from iteration k-1 is a nonpositive curvature direction but the search direction at iteration k, is not, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 1;
-   - if both are nonpositive curvature directions, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 2. (Note that the MINRES solver starts at iteration 1, so the first iteration is k = 1);
+ – at iteration k = 0, the solver takes the right-hand side (i.e., the preconditioned negative gradient) as the current solution. The same search direction is returned in `workspace.npc_dir`, and `stats.npcCount` is set to 1;
+ – at iteration k > 0, the solver returns the solution from iteration k – 1,
+   - if the residual from iteration k is a nonpositive curvature direction but the search direction at iteration k, is not, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 1;
+   - if both are nonpositive curvature directions, the residual is stored in `stats.npc_dir`, the search direction is stored in `workspace.w1`, and `stats.npcCount` is set to 2. (Note that the MINRES solver starts at iteration 1, so the first iteration is k = 1);
 * `λ`: regularization parameter;
 * `atol`: absolute stopping tolerance based on the residual norm;
 * `rtol`: relative stopping tolerance based on the residual norm;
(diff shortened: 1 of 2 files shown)


src/minres.jl
```diff
@@ -79,8 +79,12 @@ For an in-place variant that reuses memory across solves, see [`minres!`](@ref).
...
-
+* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If `true` and nonpositive curvature is detected, the behavior depends on the iteration:
+ – at iteration k = 0, return the preconditioned initial search direction in `workspace.npc_dir`;
+ – at iteration k > 0,
+   - if the residual from iteration k-1 is a nonpositive curvature direction but `workspace.w1`, the search direction at iteration k, is not, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 1;
+   - if `workspace.w1` is a nonpositive curvature direction but the residual is not, `workspace.w1` is copied into `stats.npc_dir` and `stats.npcCount` is set to 1;
```
dpo, line 86: This case cannot happen, right?

src/minres.jl
```diff
@@ -304,26 +314,47 @@ kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itm
       ArNorm = ϕbar * root  # = ‖Aᴴrₖ₋₁‖
       history && push!(ArNorms, ArNorm)
 
+      # Compute the next plane rotation (part 1).
+      γ = sqrt(γbar * γbar + β * β)
```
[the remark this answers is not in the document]
dpo: He just moved this line from line 325 below. We should fix those everywhere in separate PRs.

src/minres.jl
```diff
@@ -304,26 +314,47 @@ kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itm
...
+      γ = max(γ, ϵM)
+      
+      # Final update of w.
+      kdiv!(n, w, γ)
+
       # Check for nonpositive curvature
       if linesearch
         cγ = cs * γbar
+        ξ_w_2 = ξ_w
```
dpo, line 327: Let's call these `ζ_w` (`\zeta`) as in Dahito's paper.

src/minres.jl
```diff
@@ -304,26 +314,47 @@ kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itm
...
+            kcopy!(n, x, b)
+            w1 = w2
+          else
+            # check w direction
+            if δ_w < 0
+              # w is also a nonpositive curvature direction #TODO confirm, since I moved around the calculation of w
+              w1 = w
+              stats.npcCount = 2
+            end
+          end
```
dpo, line 346: This is quite confusing. I'll need some explanations.

dpo (changes requested):

src/minres.jl
farhadrclass: yes, I forgot to remove it

src/minres.jl
farhadrclass: @dpo 
1. **Tracking curvature in the full space with `ξ_w` and `δ_w`.**

   * `ξ_w = -cγ * ‖r‖²` intermittently estimates the Rayleigh quotient on the residual, and `δ_w` accumulates that information into a test of $w^T A w$.
   * If `δ_w` is negative, then the newly formed Lanczos direction `w` also exhibits negative curvature.

2. **Roles of `w`, `w1`, and `w2`:**

   * **`w1`** is always the first nonpositive‑curvature direction we’ll hand back to the line search.
   * **`w2`** seeds the second direction (on iteration 1 it’s just the initial residual, on later iterations it holds the previous `w1`).
   * **`w`** is a temporary alias for whichever slot we’re about to overwrite (`w2` on the very first iteration, thereafter `w1`), and we only promote it into `w1` if its curvature test (`δ_w < 0`) passes.

3. **How we terminate and hand off to the line search:**

   * On **iteration 1**, we have no previous Krylov direction, so we simply return the scaled residual.
   * On **later iterations**,   We fill `w1` with w accordingly.


does this make sense or I did it wrong?

farhadrclass:
@dpo ready for review

src/minres.jl
dpo: The explanations should be in the comments in the code, not in the pull request.

src/minres.jl
```diff
@@ -79,8 +79,11 @@ For an in-place variant that reuses memory across solves, see [`minres!`](@ref).
 * `M`: linear operator that models a Hermitian positive-definite matrix of size `n` used for centered preconditioning;
 * `ldiv`: define whether the preconditioner uses `ldiv!` or `mul!`;
 * `window`: number of iterations used to accumulate a lower bound on the error;
-* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If `linesearch` is true and nonpositive curvature is detected, the solution depends on the iteration: at iteration k = 1, the right-hand side is returned with `workspace.npc_dir` as the preconditioned initial residual; at iteration k > 1, the solution from iteration k-1 is returned with `workspace.npc_dir` as the most recent residual. (Note that the MINRES solver starts at iteration 1, so the first iteration is k = 1);
-
+* `linesearch`: if `true`, indicate that the solution is to be used in an inexact Newton method with linesearch. If `true` and nonpositive curvature is detected, the behavior depends on the iteration:
+ – at iteration k = 0, return the preconditioned initial search direction in `workspace.npc_dir`;
```
dpo, line 83: and also as the solution

src/minres.jl
```diff
@@ -79,8 +79,11 @@ For an in-place variant that reuses memory across solves, see [`minres!`](@ref).
...
+ – at iteration k > 0,
+   - if the residual from iteration k-1 is a nonpositive curvature direction but `workspace.w1`, the search direction at iteration k, is not, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 1;
```
dpo, line 85:
```suggestion
   - if the residual from iteration k-1 is a nonpositive curvature direction but the search direction at iteration k, is not, the residual is stored in `stats.npc_dir` and `stats.npcCount` is set to 1;
```

dpo (changes requested):
Please add comments to explain the recurrence to compute the curvature along w.

farhadrclass:
@dpo Updated, 
Also I added M-A. Dahito and D. Orban's paper as a reference since we are adapting their calculation for search directions.

dpo (approved):
Thank you!

PR #1013, opened 2025-07-06
Add Curvature Metrics (dAd and rAr) to MINRES, CR, and CG Solvers

farhadrclass:
**Summary of Changes**

This update adds computation of the following quantities for the **MINRES**, **CR**, and **CG** solvers:

* **`dAd`**: The dot product of the direction vector with `A` times the direction vector.
* **`rAr`**: The dot product of the residual vector with `A` times the residual vector.

These metrics can be useful for diagnostics, convergence analysis, and monitoring solver behavior.

### DIFF
--- a/src/cg.jl
+++ b/src/cg.jl
@@ -170,6 +170,8 @@ kwargs_cg = (:M, :ldiv, :radius, :linesearch, :atol, :rtol, :itmax, :timemax, :v
       stats.status = "x is a zero-residual solution"
       warm_start && kaxpy!(n, one(FC), Δx, x)
       workspace.warm_start = false
+      stats.dAd = zero(T)
+      stats.rAr = γ
       return workspace
     end
 
@@ -287,6 +289,8 @@ kwargs_cg = (:M, :ldiv, :radius, :linesearch, :atol, :rtol, :itmax, :timemax, :v
     stats.inconsistent = inconsistent
     stats.timer = start_time |> ktimer
     stats.status = status
+    stats.dAd = pAp
+    stats.rAr = γ
     return workspace
   end
 end
--- a/src/cr.jl
+++ b/src/cr.jl
@@ -181,6 +181,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
       history && push!(ArNorms, zero(T))
       warm_start && kaxpy!(n, one(FC), Δx, x)
       workspace.warm_start = false
+      stats.dAd = zero(T)
+      stats.rAr = zero(T)
       return workspace
     end
 
@@ -200,6 +202,8 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
         stats.npcCount = 1
         stats.indefinite = true
       end
+      stats.dAd = zero(T)
+      stats.rAr = zero(T)
       return workspace
     end
 
@@ -251,6 +255,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
             kcopy!(n, npc_dir, p)
             kcopy!(n, x, p)  # x ← M⁻¹ b
             stats.npcCount = 1
+            stats.dAd = pAp
           else
             if r_curv
               kcopy!(n, npc_dir, r)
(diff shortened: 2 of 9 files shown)


farhadrclass:
@dpo  this is the very first idea, 
I am open for suggestions such as if you want to collect them in a tuple then assign them or calculate them only when we have negative curvature (also any name suggestions would be appreciated )

src/cg.jl
```diff
@@ -170,6 +170,8 @@ kwargs_cg = (:M, :ldiv, :radius, :linesearch, :atol, :rtol, :itmax, :timemax, :v
       stats.status = "x is a zero-residual solution"
       warm_start && kaxpy!(n, one(FC), Δx, x)
       workspace.warm_start = false
+      stats.dAd = zero(T)
+      stats.rAr = γ
```
dpo, line 174: `γ` is not $r^T A r$ in CG; it is $r^T M r$, where $M$ is the preconditioner.

src/cr.jl
```diff
@@ -251,6 +255,7 @@ kwargs_cr = (:M, :ldiv, :radius, :linesearch, :γ, :atol, :rtol, :itmax, :timema
             kcopy!(n, npc_dir, p)
             kcopy!(n, x, p)  # x ← M⁻¹ b
             stats.npcCount = 1
+            stats.dAd = pAp
```
dpo, line 258: Why not set both variables at each iteration after line 243?

src/krylov_stats.jl
```diff
@@ -14,6 +14,8 @@ The fields are as follows:
 - `inconsistent`: Flags whether the system was detected as inconsistent (i.e., when `b` is not in the range of `A`);
 - `indefinite`: Flags whether the system was detected as indefinite (i.e., when `A` is not positive definite);
 - `npcCount`: The number of nonpositive curvature directions encountered during the solve;
+- `dAd`: The dot product of the direction with the matrix `A` times the direction;
```
dpo, line 17:
```suggestion
- `dAd`: The dot product of the direction with the matrix `A` times the search direction;
```

src/block_krylov_workspaces.jl
```diff
@@ -115,7 +115,7 @@ function BlockGmresWorkspace(m::Integer, n::Integer, p::Integer, SV::Type, SM::T
   τ  = SV[SV(undef, p) for i = 1 : memory]
   SV = isconcretetype(SV) ? SV : typeof(τ)
   SM = isconcretetype(SM) ? SM : typeof(X)
-  stats = SimpleStats(0, false, false, false, 0, T[], T[], T[], 0.0, "unknown")
+  stats = SimpleStats(0, false, false, false, 0, zero(T), zero(T), T[], T[], T[], 0.0, "unknown")
```
dpo, line 118: Why do you need to modify the block Krylov workspaces?

src/minres.jl
```diff
@@ -361,6 +364,10 @@ kwargs_minres = (:M, :ldiv, :linesearch ,:λ, :atol, :rtol, :etol, :conlim, :itm
           stats.status = "nonpositive curvature"
           workspace.warm_start = false
           stats.indefinite = true
+          mul!(r2, A, npc_dir) # r2 = A * npc_dir
+          mul!(w2, A, w1)  # w2 = A * w1
+          stats.dAd = kdotr(n, w1, w2)  # stats.dAd = w1^T * A * w1
+          stats.rAr = kdotr(n, npc_dir, r2)  # stats.rAr = npc_dir^T * A * npc_dir
```
dpo, line 370: We don't want additional products with $A$. `stats.dAd` and `stats.rAr` should contain the estimates.

dpo (changes requested):

src/block_krylov_workspaces.jl
farhadrclass: Since I update the SimpleStats, it needs to be update here where we create a new one

src/minres.jl
farhadrclass: I can provide the estimates, but they won’t be fully accurate—only approximately correct in terms of the sign of the product. The actual values will differ, though we can still tell whether something is significantly negative or not.
is that good enough for us?

farhadrclass: @dpo  I need your input on this to go forward

dpo: @farhadrclass In Krylov methods, all "estimates" coincide with the actual value, but only in exact arithmetic. What I'm saying is that we have "estimates" of rAr and dAd that are also exact in exact arithmetic. They are the same in MINRES and CR.

PR #1033, opened 2025-10-28
Loosen type-constraints don't enforce `<:AbstractVector`

vchuravy:
@lcw and I just run into a case where we have a data-structure `GridArray` that for the purpose of Krylov.jl acts like a Vector and we have defined the Krylov custom workspace API for, 
but for other reasons presents as `AbstractArray{Float64, 2}`.

IIRC the last time we spoke about this, there was an argument that the reason why Krylov.jl has this type constraint is so that the block-gmres API can use Matrix as an input.

### DIFF
--- a/src/gmres.jl
+++ b/src/gmres.jl
@@ -89,9 +89,9 @@ and [`krylov_solve!`](@ref) to run the Krylov method in-place.
 function gmres! end
 
 def_args_gmres = (:(A                    ),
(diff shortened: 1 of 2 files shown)


amontoison:
I can understand that but can you show me what is doing `GridArray` such that I better understand the type?
I think I have an alternative solution that you will like 🙂

PR #1040, opened 2025-11-23
Fix warm-start of TriCG and TriMR

timholy:
The three commits here:

- improve the docstrings for `tricg` and `trimr`
- add tests that fail on `main` due to a bug in the size-checks for the supplied warmstart vectors
- fix the bug

Feel free to either squash or, if you want the docstring changes separately from the rest, merge as separate commits. I can also squash the latter two commits into one if you prefer.

### DIFF
--- a/src/workspace_accessors.jl
+++ b/src/workspace_accessors.jl
@@ -153,15 +153,27 @@ for (KS, fun, nsol, nA, nAt, warm_start) in [
       issolved(workspace :: $KS) = workspace.stats.solved
     end
     if $warm_start
-      if $KS in (BilqrWorkspace, TrilqrWorkspace, TricgWorkspace, TrimrWorkspace, GpmrWorkspace)
+      if $KS in (BilqrWorkspace, TrilqrWorkspace)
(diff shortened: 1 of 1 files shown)


amontoison:
> Changes implemented. Note the two marked `@test_broken` for `trilqr`, which can be fixed separately someday.

Did you added the test for rectangular matrices? You need to ensure that `Ax = b` and `A'y = c` are consistent, which require `b` to be in the range of A and `c` in the range of A'.

timholy:
> Did you added the test for rectangular matrices? You need to ensure that Ax = b and A'y = c are consistent, which require b to be in the range of A and c in the range of A'.

Yes. Perhaps I should rename
```julia
A5, b5, c5, M5, N5 = small_sqd(false; FC)
```
to be
```julia
A3x2, b3, c2, M3x3, N2x2 = small_sqd(false; FC)
```
? (The overall system is of size 5x5, which is where the 5 comes from.)

Thanks for reviewing so quickly! Do get some sleep!

amontoison:
@timholy I tagged a new release `0.10.3`. It contains your fixes.

PR #1043, opened 2025-11-25
TRICG & TRIMR: support warm-start with ldiv=true

timholy:
This addresses some potential inconsistencies in `tricg` and `trimr` in
the case `ldiv=true`:

- docs: fixes the statement about the relationship between `E`/`F` and
  `M`/`N`
- docs: describes the residual-norm in terms of `E`/`F` rather than
  `M`/`N` (only the former are unambiguous, given the problem statement)
- adds support for warm-starting when `ldiv=true` for arbitrary `M`/`N`

~NOTE: this is a WIP as `trimr` has not yet been modified. Once we settle on the `tricg` changes, I'll port the same changes to `trimr`.~

### DIFF
--- a/src/tricg.jl
+++ b/src/tricg.jl
@@ -75,7 +75,7 @@ For an in-place variant that reuses memory across solves, see [`tricg!`](@ref).
 * `y0`: a vector of length `n` that represents an initial guess of the solution `y`.
 
 Warm-starting is supported only when:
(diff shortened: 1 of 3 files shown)


timholy:
I believe this is ready for merging. Since it is about warm-starting, I also went ahead and implemented a fix for #1076 for these two methods.

PR #1045, opened 2025-11-30
Improve inference for Double64

timholy:
`sym_givens` inferred return type is `Tuple{Union{Float64, Double64}, Union{Float64, Double64}, Double64}` when passed `Double64` arguments. This commit reworks it to ensure deterministic return types. It also modernizes the implementation to avoid unnecessary type conversions, branches, etc.

## Related issues

There is no related issue.

## Checklist

- [X] I am following the contributing guidelines
- [X] Tests are passing
- [NA] Lint workflow is passing
- [NA] Docs were updated and workflow is passing

### DIFF
--- a/src/krylov_utils.jl
+++ b/src/krylov_utils.jl
@@ -24,29 +24,25 @@ function sym_givens(a :: T, b :: T) where T <: AbstractFloat
   # http://www.stanford.edu/group/SOL/dissertations/sou-cheng-choi-thesis.pdf
   # D. Orban, Montreal, May 2015.
 
-  if b == 0
-    if a == 0
-      c = one(T)
-    else
-      c = sign(a)  # In Julia, sign(0) = 0.
-    end
+  if iszero(b)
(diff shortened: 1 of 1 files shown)


amontoison:
@timholy Is it you that called copilot?
I don't know who setup all these extensions...

src/krylov_utils.jl
```diff
@@ -24,29 +24,25 @@ function sym_givens(a :: T, b :: T) where T <: AbstractFloat
...
   # D. Orban, Montreal, May 2015.
 
-  if b == 0
-    if a == 0
-      c = one(T)
-    else
-      c = sign(a)  # In Julia, sign(0) = 0.
-    end
+  if iszero(b)
+    c = T(sign(a) + iszero(a)) # In Julia, sign(0) = 0.
```
amontoison: Can you confirm that `sign(0)` is also `0` for other types like Double64?
I never checked with it.

amontoison (approved):
LGTM 👍
I will merge the PR after the CI tests.

PR #1079, opened 2026-04-07
Use combined tolerance for solved check

farhadrclass:
Replace the previous rNorm ≤ rtol check with rNorm ≤ ε (where ε = atol + rtol * β₁) when setting solved, solved_mach, and solved_lim. This makes the convergence/zero-residual detection consistent with the computed combined tolerance (absolute+relative) and prevents incorrect termination when atol or the initial scale β₁ are nonzero.

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -267,7 +267,7 @@ kwargs_workspace_minres = (:window,)
     kdisplay(iter, verbose) && @printf(iostream, "%5d  %7.1e  %7.1e  %7.1e  %8.1e  %8.1e  %7.1e  %7.1e  %7s  %7s  %.2fs\n", iter, rNorm, ArNorm, β, cs, sn, ANorm, Acond, "✗ ✗ ✗ ✗", "✗ ✗ ✗ ✗", start_time |> ktimer)
 
     ε = atol + rtol * β₁
-    solved = solved_mach = solved_lim = (rNorm ≤ ε)
+    solved = solved_mach = solved_lim = false
     tired  = iter ≥ itmax
     ill_cond = ill_cond_mach = ill_cond_lim = false
     zero_resid = zero_resid_mach = zero_resid_lim = (rNorm ≤ ε)
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -267,7 +267,7 @@ kwargs_workspace_minres = (:window,)
     kdisplay(iter, verbose) && @printf(iostream, "%5d  %7.1e  %7.1e  %7.1e  %8.1e  %8.1e  %7.1e  %7.1e  %7s  %7s  %.2fs\n", iter, rNorm, ArNorm, β, cs, sn, ANorm, Acond, "✗ ✗ ✗ ✗", "✗ ✗ ✗ ✗", start_time |> ktimer)
 
     ε = atol + rtol * β₁
-    solved = solved_mach = solved_lim = (rNorm ≤ rtol)
+    solved = solved_mach = solved_lim = (rNorm ≤ ε)
     tired  = iter ≥ itmax
     ill_cond = ill_cond_mach = ill_cond_lim = false
     zero_resid = zero_resid_mach = zero_resid_lim = (rNorm ≤ ε)


farhadrclass:
@dpo  this is the issue, 
It was a small fix,

dpo:
That seems right. It's also the test used in MINRES-QLP: https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/minres_qlp.jl#L480

src/minres.jl
```diff
@@ -267,7 +267,7 @@ kwargs_workspace_minres = (:window,)
     kdisplay(iter, verbose) && @printf(iostream, "%5d  %7.1e  %7.1e  %7.1e  %8.1e  %8.1e  %7.1e  %7.1e  %7s  %7s  %.2fs\n", iter, rNorm, ArNorm, β, cs, sn, ANorm, Acond, "✗ ✗ ✗ ✗", "✗ ✗ ✗ ✗", start_time |> ktimer)
 
     ε = atol + rtol * β₁
-    solved = solved_mach = solved_lim = (rNorm ≤ rtol)
+    solved = solved_mach = solved_lim = (rNorm ≤ ε)
     tired  = iter ≥ itmax
     ill_cond = ill_cond_mach = ill_cond_lim = false
     zero_resid = zero_resid_mach = zero_resid_lim = (rNorm ≤ ε)
```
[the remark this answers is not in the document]
farhadrclass: I don't agree with this review

farhadrclass:
@dpo  I agree and also the same test in CR and CG,

Also I don't agree with co-pilot review so I ignored it 

If it is good for you, should we merge it ?

src/minres.jl
```diff
@@ -267,7 +267,7 @@ kwargs_workspace_minres = (:window,)
     kdisplay(iter, verbose) && @printf(iostream, "%5d  %7.1e  %7.1e  %7.1e  %8.1e  %8.1e  %7.1e  %7.1e  %7s  %7s  %.2fs\n", iter, rNorm, ArNorm, β, cs, sn, ANorm, Acond, "✗ ✗ ✗ ✗", "✗ ✗ ✗ ✗", start_time |> ktimer)
 
     ε = atol + rtol * β₁
-    solved = solved_mach = solved_lim = (rNorm ≤ rtol)
+    solved = solved_mach = solved_lim = (rNorm ≤ ε)
```
amontoison: We use 6 variables for the same stopping condition, maybe we should just removed a few of them.
What you want is already at line 273:
`zero_resid = zero_resid_mach = zero_resid_lim = (rNorm ≤ ε)`.

I don't understand why we have so many variables.

farhadrclass: @dpo  and @amontoison 
I agree but the goal of this PR is just to fix the current issue that is blocking JSO solver from using MINREs

dpo: Is it a typo? Should it be `ArNorm` in the line above?

amontoison: It will change the returned `status` at the end. With your modifications, we have two times the same stopping conditions but the related variables are used for different status.

dpo: At the end of the main loop, `solved` is set based on $$\Vert A r \Vert$$, not $$\Vert r \Vert$$: https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/1079/changes#diff-65c1dbcc771522673d6f0c4e47c90e950b2bcb251ceda42f6de20901c3055021R449

So I believe the line that @farhadrclass modified should instead be
```julia
solved = solved_mach = solved_lim = (ArNorm ≤ ε)
```
There could also be an additional test
```julia
zero_resid = zero_resid_mach = zero_resid_linm = (rNorm ≤ ε)
```
Those are stopping conditions of different natures.

amontoison: I checked quickly the code and we don't have any stopping condition of `ArNorm`.
I think you are right Dominique.
The tolerance should just be based on `atol + rtol * ||Ab||` like in MINRES-QLP or USYMQR and not `ε = atol + rtol * ||b||`.

amontoison: - https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/usymqr.jl#L336
- https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/usymqr.jl#L339

dpo: The condition with test2 is about ArNorm.

farhadrclass: > At the end of the main loop, `solved` is set based on ‖ A r ‖ , not ‖ r ‖ : https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/1079/changes#diff-65c1dbcc771522673d6f0c4e47c90e950b2bcb251ceda42f6de20901c3055021R449
> 
> So I believe the line that @farhadrclass modified should instead be
> 
> ```julia
> solved = solved_mach = solved_lim = (ArNorm ≤ ε)
> ```
> 
> There could also be an additional test
> 
> ```julia
> zero_resid = zero_resid_mach = zero_resid_linm = (rNorm ≤ ε)
> ```
> 
> Those are stopping conditions of different natures.

yes I agree but then why in minres_qlp we have 
`    solved = zero_resid = zero_resid_lim = rNorm ≤ ε`

farhadrclass: should we have `solved = zero_resid`?

farhadrclass: Also ArNorm is zero before the loop at line 254

farhadrclass: @dpo @amontoison I agree with  about the difference between the exact solution (`rNorm`) and the least-squares solution (`ArNorm`). 

@dpo, However, to handle the pre-loop initialization, I can't check `(ArNorm ≤ ε)` because `ArNorm` is initialized to `0` at line 254. Evaluating it before the loop would cause the solver to instantly terminate on every problem!

Instead of hardcoding the flags to `false` or calculating `‖Ab‖` before the loop, I think the cleanest fix is to just use `(rNorm ≤ ε)` for the pre-loop check. Mathematically, if the initial exact residual is below tolerance, it is also a valid least-squares solution ($r=0 \implies A^T r = 0$), no? 

@dpo  what do you think?
I also think if we want to rethink the variables, that would be a bigger PR and should include more solvers not just MINRES and that would my attention away from JSOSolvers now

dpo:
```suggestion
    solved = solved_mach = solved_lim = false
```
I'm making this suggestion because:

- the `solved` stopping test is really based on $\Vert A r \Vert$
- we don't have $\Vert A r \Vert$ before the main loop.

In your case @farhadrclass, the change should force one iteration and should solve your issue.

dpo (changes requested):

src/minres.jl
farhadrclass: that is actually what I was thinking, 
This way it force it and make more sense for future edits

dpo:
Why is documentation failing?

amontoison:
> Why is documentation failing?

It is related to a tutorial with FFTW.jl for linear operators.
It is not related to this PR.

I will fix it when I will be allowed to commit again.

farhadrclass:
@dpo  I think @amontoison is right 

I think the issue with LaplacianOperator being negative defined no?

dpo (approved):

dpo:
Thank you!

PR #1080, opened 2026-04-08
Initialize solved/zero_resid flags to false

farhadrclass:
Set solved, zero_resid, and zero_resid_lim to false during initialization instead of computing them from rNorm ≤ ε. This avoids marking the problem as solved at startup (and potential use of rNorm before it's set), ensuring stopping flags are only updated after the residual is computed.

### DIFF
--- a/src/minres_qlp.jl
+++ b/src/minres_qlp.jl
@@ -225,7 +225,8 @@ kwargs_minres_qlp = (:M, :ldiv, :linesearch, :λ, :atol, :rtol, :Artol, :itmax, 
 
     # Stopping criterion.
     breakdown = false
-    solved = zero_resid = zero_resid_lim = false
+    solved = false
+    zero_resid = zero_resid_lim = rNorm ≤ ε
     zero_resid_mach = false
     inconsistent = false
     ill_cond_mach = false
--- a/src/minres_qlp.jl
+++ b/src/minres_qlp.jl
@@ -225,7 +225,7 @@ kwargs_minres_qlp = (:M, :ldiv, :linesearch, :λ, :atol, :rtol, :Artol, :itmax, 
 
     # Stopping criterion.
     breakdown = false
-    solved = zero_resid = zero_resid_lim = rNorm ≤ ε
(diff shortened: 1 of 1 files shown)


dpo:
Set `zero_resid = zero_resid_lim = rNorm ≤ ε`, but `solved = false`.

amontoison:
I don't understand your modification @farhadrclass.
Why do you want to do an iteration in all cases if the residual is small enough?
It includes cases like warm-start and we want to stop if `||r_0|| = ||b - A x_0||` is already very small.

farhadrclass:
@amontoison  it should be as @dpo  mentioned, I made a mistake and asssign all of them

amontoison:
You now go in the `while` loop with your modification:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/minres_qlp.jl#L238

We don't want that.

amontoison:
> Set solved, zero_resid, and zero_resid_lim to false during initialization instead of computing them from rNorm ≤ ε. This avoids marking the problem as solved at startup (and potential use of rNorm before it's set), ensuring stopping flags are only updated after the residual is computed.

I don't understand the goal of the PR.
You want to check `rNorm == 0` or `rNorm ≤ ε` before the main loop in many methods.
`rNorm` is well-defined here before the main loop and it seems to me that we don't need any modification.

dpo:
Sorry; the flags in this solver and others are a bit confusing. MINRES and MINRES-QLP should stop when $r$ is small (`zero_resid`) or $Ar$ is small (`solved`). But currently, for inconsistent systems, it seems that `solved` will be `false` and `inconsistent` will be `true`. That conveys the idea that the solve was not successful when it was. I think that, instead, we should have

* `solved = true` when $Ar$ is small
* `inconsistent = true` when $Ar$ is small but $r$ is not small
* `zero_resid = true` and `inconsistent = false` when $r$ is small.

I think this would affect several other solvers, and so should be done in a separate PR. I’m going to close this for now. Thanks @farhadrclass.

PR #1082, opened 2026-05-08
Fix premature termination in `minres!` due to dimensional mismatch in stopping criteria

farhadrclass:
**Summary**
This PR fixes a bug in the `minres!` loop where the solver would terminate prematurely (often on the first iteration) when solving systems with a large initial residual.

**Details**
In the stopping criteria for `minres!`, the variable `test2` is defined as `root / ANorm` (representing the dimensionless ratio $\|A^H r\| / \|A\|$). However, it was being compared against `ε`, which is a dimensionful threshold scaled by the initial residual `β₁`:

```julia
ε = atol + rtol * β₁
# ...
solved_lim = (test2 ≤ ε)

```

When used inside outer non-linear solvers (like Trust-Region or Regularized Newton methods), the initial gradient/residual `β₁` is often quite large. If `rtol * β₁ > 1.0`, `ε` becomes greater than 1. Because `test2` is inherently bounded by 1, the condition `test2 ≤ ε` evaluates to `true` instantly.

This forces MINRES to exit with `status = "found approximate minimum least-squares solution"` without actually solving the system.

**Impact**

* **Suppressed Curvature Detection:** Because the solver exited on iteration 1, it completely bypassed the Armijo-Goldstein negative curvature handlers (`cγ = cs * γbar >= 0`).
* **Outer Solver Stagnation:** In non-convex benchmarks (like the Extended Rosenbrock), handing back an unsolved step caused the outer R2N loop to reject the step, skyrocketing the regularization parameter $\sigma$ and stalling the algorithm.

**The Fix**
Decoupled the dimensionless `test2` check from the dimensionful `ε`. `test2` is now evaluated against the dimensionless tolerances directly:

```julia
solved_lim = (test2 ≤ rtol + atol)

```

This simple correction brings the behavior of `minres!` perfectly in line with `minres_qlp!` and CR, restoring its ability to reliably catch non-positive curvature directions on highly indefinite systems.

### DIFF
--- a/src/minres.jl
+++ b/src/minres.jl
@@ -446,7 +446,7 @@ kwargs_workspace_minres = (:window,)
       # Stopping conditions based on user-provided tolerances.
       tired = iter ≥ itmax
       ill_cond_lim = (inv(Acond) ≤ ctol)
-      solved_lim = (test2 ≤ ε)
+      solved_lim = (test2 ≤ rtol + atol)
       zero_resid_lim = MisI && (test1 ≤ eps(T))
       resid_decrease_lim = (rNorm ≤ ε)
       iter ≥ window && (fwd_err = err_lbnd ≤ etol * sqrt(xENorm²))


amontoison:
Thanks ChatGPT but the explanation is totally wrong and doesn't make any sense...

farhadrclass:
@dpo  @amontoison This is actually the case,
I edit the text with AI but the issue is that MINRES takes way too much step to solve a sub-problem vs CR or even CG

The issue is that MINRES would stop before it even check the problem because the test2 fails on the first iteration while CR, minres_QLP, CG all pass (I am pretty sure MINRES needs re-ordering and why are we having too many tests and not following a same structures as minres_QLP or CR)

farhadrclass:
why is `ε = atol + rtol * β₁` has  β₁ in it here and in MINRES_QLP doesn't

I am open to suggestion on why the MINRES is doing so poor compare with CR or MINRES_QLP when I use it in my R2N solver

amontoison:
> Member

`β₁` is the norm of the right-hand side, it is the same for all solvers.
We just call it `rNorm` in MINRES-QLP:
https://github.com/JuliaSmoothOptimizers/Krylov.jl/blob/main/src/minres_qlp.jl#L206

I also don't know why R2N is not working well with MINRES but it doesn't make any sense to add an ABSOLUTE tolerance with a RELATIVE tolerance. The goal of a RELATIVE tolerance is to be relative to something, which is the initial residual (== right-hand side without warm-start).

Did you play with the tolerances? Did you try to set `rtol=0.0` and just rely on an absolute tolerance?

You also said that "`test2` is inherently bounded by `1`." But it is an hallucination from AI...

amontoison:
> @dpo @amontoison This is actually the case, I edit the text with AI but the issue is that MINRES takes way too much step to solve a sub-problem vs CR or even CG
> 
> The issue is that MINRES would stop before it even check the problem because the test2 fails on the first iteration while CR, minres_QLP, CG all pass (I am pretty sure MINRES needs re-ordering and why are we having too many tests and not following a same structures as minres_QLP or CR)

The modification is this PR is not valid and the explication doesn't make any sense.
Open an issue and a provide a reproducer if you think that we have something to fix / modify in `Krylov.jl`.

I also don't see any "dimensional mismatch" in the computation of `||A^H r|| / ||A||` (wrong title?)
I will add a contribution rule like in JuMP to not use AI if it leads to not understand the content of your PR.

farhadrclass:
why are we doing this 
https://github.com/Farhad-phd/Krylov.jl/blob/315141e3f154f91c3638be9739bb43a1951512c8/src/minres.jl#L419

and for R2N, we do not change anything when we change the subsolver, just the name, all tolerances are the same

amontoison:
@farhadrclass Stop opening the PR, I said it is wrong and I asked to opened an issue if needed...

farhadrclass:
@amontoison please do not close this PR, I will do it once we agree, It is very challenging to answer your earlier message once you keep closing it 

I am trying to find the cause why MINRES does not follow the same same CR

amontoison:
> @amontoison please do not close this PR, I will do it once we agree, It is very challenging to answer your earlier message once you keep closing it

I don't see why it is challenging. I close the PR because the content is wrong (title, modification and description).
I also spent time to explain you why.
You can continue to write messages, it is not a valid argument...

> I am trying to find the cause why MINRES does not follow the same same CR

Please open an issue and provide a reproducer with a linear system.
It is how we could help you.
Once again, this PR is not a correct solution.

farhadrclass:
@amontoison here is the issue https://github.com/JuliaSmoothOptimizers/Krylov.jl/issues/1083

PR #1100, opened 2026-07-06
Support for any floating-point data type in block Krylov methods

amontoison:
Add a generic Julia implementation for `larfg`, `geqrf`, `orgqr` and `ormqr` (LAPACK routines).
It allows support for any precision with the block Krylov solvers implemented in `Krylov.jl` (block-MINRES / block-GMRES).

@langou It could be relevant for you if we want one day we use what you presented at Saint-Girons.

### DIFF
--- a/src/block_krylov_utils.jl
+++ b/src/block_krylov_utils.jl
@@ -194,7 +194,7 @@ function householder!(Q::AbstractMatrix{FC}, R::AbstractMatrix{FC}, τ::Abstract
   kfill!(R, zero(FC))
   kgeqrf!(Q, τ)
   copy_triangle(Q, R, k)
-  !compact && korgqr!(Q, τ)
+  !compact && kungqr!(Q, τ)
   return Q, R
 end
 
@@ -203,11 +203,11 @@ function householder!(Q::AbstractMatrix{FC}, R::AbstractMatrix{FC}, τ::Abstract
   kfill!(R, zero(FC))
   kgeqrf!(Q, τ, buffer)
   copy_triangle(Q, R, k)
-  !compact && korgqr!(Q, τ, buffer)
+  !compact && kungqr!(Q, τ, buffer)
   return Q, R
 end
 
-for (Xgeqrf, Xorgqr, Xormqr, T) in ((:sgeqrf_, :sorgqr_, :sormqr_, :Float32   ),
+for (Xgeqrf, Xungqr, Xunmqr, T) in ((:sgeqrf_, :sorgqr_, :sormqr_, :Float32   ),
                                     (:dgeqrf_, :dorgqr_, :dormqr_, :Float64   ),
                                     (:cgeqrf_, :cungqr_, :cunmqr_, :ComplexF32),
                                     (:zgeqrf_, :zungqr_, :zunmqr_, :ComplexF64))
@@ -235,57 +235,57 @@ for (Xgeqrf, Xorgqr, Xormqr, T) in ((:sgeqrf_, :sorgqr_, :sormqr_, :Float32   ),
             return nothing
         end
 
-        function $Xorgqr(m, n, k, a, lda, tau, work, lwork, info)
-            return ccall((@blasfunc($Xorgqr), libblastrampoline), Cvoid,
+        function $Xungqr(m, n, k, a, lda, tau, work, lwork, info)
+            return ccall((@blasfunc($Xungqr), libblastrampoline), Cvoid,
                          (Ref{BlasInt}, Ref{BlasInt}, Ref{BlasInt}, Ptr{$T},
                           Ref{BlasInt}, Ptr{$T}, Ptr{$T}, Ref{BlasInt}, Ref{BlasInt}),
                           m, n, k, a, lda, tau, work, lwork, info)
         end
 
-        function korgqr_buffer!(A::Matrix{$T}, tau::Vector{$T})
+        function kungqr_buffer!(A::Matrix{$T}, tau::Vector{$T})
             m, n = size(A)
             k = length(tau)
             work = Ref{$T}(0)
             lda = max(1, stride(A, 2))
-            $Xorgqr(m, n, k, A, lda, tau, work, -1, 0)
+            $Xungqr(m, n, k, A, lda, tau, work, -1, 0)
             return work[] |> BlasInt
         end
 
-        function korgqr!(A::Matrix{$T}, tau::Vector{$T}, work::Vector{$T})
-            symb = @blasfunc($Xorgqr)
+        function kungqr!(A::Matrix{$T}, tau::Vector{$T}, work::Vector{$T})
+            symb = @blasfunc($Xungqr)
             m, n = size(A)
             k = length(tau)
             lwork = length(work)
             lda = max(1, stride(A, 2))
-            $Xorgqr(m, n, k, A, lda, tau, work, lwork, 0)
+            $Xungqr(m, n, k, A, lda, tau, work, lwork, 0)
             return nothing
         end
 
-        function $Xormqr(side, trans, m, n, k, a, lda, tau, c, ldc, work, lwork, info)
-            return ccall((@blasfunc($Xormqr), libblastrampoline), Cvoid,
+        function $Xunmqr(side, trans, m, n, k, a, lda, tau, c, ldc, work, lwork, info)
+            return ccall((@blasfunc($Xunmqr), libblastrampoline), Cvoid,
                          (Ref{UInt8}, Ref{UInt8}, Ref{BlasInt}, Ref{BlasInt}, Ref{BlasInt}, Ptr{$T},
                           Ref{BlasInt}, Ptr{$T}, Ptr{$T}, Ref{BlasInt}, Ptr{$T}, Ref{BlasInt},
                           Ref{BlasInt}, Clong, Clong),
                           side, trans, m, n, k, a, lda, tau, c, ldc, work, lwork, info, 1, 1)
         end
 
-        function kormqr_buffer!(side::Char, trans::Char, A::Matrix{$T}, tau::Vector{$T}, C::Matrix{$T})
+        function kunmqr_buffer!(side::Char, trans::Char, A::Matrix{$T}, tau::Vector{$T}, C::Matrix{$T})
             m, n = size(A)
             k = length(tau)
             work = Ref{$T}(0)
             lda = max(1, stride(A, 2))
             ldc = max(1, stride(C, 2))
-            $Xormqr(side, trans, m, n, k, A, lda, tau, C, ldc, work, -1, 0)
+            $Xunmqr(side, trans, m, n, k, A, lda, tau, C, ldc, work, -1, 0)
             return work[] |> BlasInt
         end
 
-        function kormqr!(side::Char, trans::Char, A::Matrix{$T}, tau::Vector{$T}, C::Matrix{$T}, work::Vector{$T})
+        function kunmqr!(side::Char, trans::Char, A::Matrix{$T}, tau::Vector{$T}, C::Matrix{$T}, work::Vector{$T})
             m, n = size(A)
             k = length(tau)
             lwork = length(work)
             lda = max(1, stride(A, 2))
             ldc = max(1, stride(C, 2))
-            $Xormqr(side, trans, m, n, k, A, lda, tau, C, ldc, work, lwork, 0)
+            $Xunmqr(side, trans, m, n, k, A, lda, tau, C, ldc, work, lwork, 0)
             return nothing
         end
     end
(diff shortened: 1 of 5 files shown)


amontoison:
@MaxenceGollier Can you update the workflow for breakages such that the same message is updated by the GitHub bot?

I did the modification for that in Krylov a few years ago but maybe it was not used by all the other JSO packages and you used an old one as template.

I don't want to be spammed at each commit :)

MaxenceGollier:
> @MaxenceGollier Can you update the workflow for breakages such that the same message is updated by the GitHub bot?
> 
> I did the modification for that in Krylov a few years ago but maybe it was not used by all the other JSO packages and you used an old one as template.
> 
> I don't want to be spammed at each commit :)

Yes, good idea, i will look into it this week.

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+
+# """
+#     β, τ = larfg!(α, x)
+#
+# Generate an elementary Householder reflector `H = I - τ vvᴴ` such that
+# `Hᴴ * [α; x] = [β; 0]` (LAPACK convention; `H` is not hermitian in the complex
+# case), where `v = [1; y]` and `β` is the (signed) Euclidean norm of `[α; x]`.
+# On output `x` is overwritten by the tail `y` of `v`.
+# """
+function larfg!(α::FC, x::AbstractVector{FC}) where FC <: FloatOrComplex
```
langou, line 311: I think it'd be better to have α as inout rather than α is input and β is output. Same as in LAPACK. This changes GEQRF below. I think it'd be better. Most of the time (in Numerical Linear Algebra) we'll want the β value to be (in output) where the α value was (in input).

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+    return real(α), zero(FC)
+  end
+  β = -copysign(hypot(abs(α), xnorm), real(α))
+  τ = (β - α) / β
+  kdiv!(n, x, α - β)
+  return β, τ
+end
+
+# """
+#     A = geqrf!(A, tau)
```
langou: This implementation is geqr2 and not geqrf. It is fine to call geqrf though and then declines variants under the geqrf name. I wonder if one will not want a workspace for some variants.

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+# """
+function geqrf!(A::AbstractMatrix{FC}, tau::AbstractVector{FC}) where FC <: FloatOrComplex
+  m, n = size(A)
+  k = min(m, n)
+  for i = 1:k
+    x = view(A, i+1:m, i)
+    βi, τi = larfg!(A[i,i], x)
+    tau[i] = τi
+    if i < n && τi != zero(FC)
+      A[i,i] = one(FC)
```
langou, line 339: I do not recommend to do this. I recommend to store βi in A[i,i] in output of larfg (see comment on larfg) and then break down the kdot / kaxpy below to work on a vector of size m-i (instead of m-i+1) and assume the 1 in the computation. (So handles the computation with A[i,i] separately assuming implicitly that A[i,i] is 1.)

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+# Generate an elementary Householder reflector `H = I - τ vvᴴ` such that
+# `Hᴴ * [α; x] = [β; 0]` (LAPACK convention; `H` is not hermitian in the complex
+# case), where `v = [1; y]` and `β` is the (signed) Euclidean norm of `[α; x]`.
+# On output `x` is overwritten by the tail `y` of `v`.
+# """
+function larfg!(α::FC, x::AbstractVector{FC}) where FC <: FloatOrComplex
+  T = real(FC)
+  n = length(x)
+  xnorm = knorm(n, x)
+  if xnorm == zero(T) && imag(α) == zero(T)
```
langou, line 315: This is good and consistent with LAPACK. I am starting to wonder whether this LAPACK convention is a good thing or not. Maybe a good idea is to remove this special case. I am looking at this now. So I do not have an definitive answer. For now, I think this is good to leave

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+        kaxpy!(p, -conj(τi) * s, v, c)
+      end
+    end
+    A[i,i] = βi
+  end
+  return A
+end
+
+# """
+#     A = orgqr!(A, tau)
```
langou: Do you want to call it ungqr or orgqr? I think ungqr would be better. 

And similalry (1) this is more of a ung2r than a ungqrf, (2) a workspace in interface might be needed for some variants.

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+    A[i,i] = one(FC) - τi
+    for l = 1:i-1
+      A[l,i] = zero(FC)
+    end
+  end
+  return A
+end
+
+# """
+#     C = ormqr!(side, trans, A, tau, C)
```
langou: Same comment: unmqr, and more like a unm2r.

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+  for j = k+1:n
+    for l = 1:m
+      A[l,j] = zero(FC)
+    end
+    A[j,j] = one(FC)
+  end
+  for i = k:-1:1
+    τi = tau[i]
+    if i < n && τi != zero(FC)
+      A[i,i] = one(FC)
```
langou, line 371: I do not think this is good practice.

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+  m, n = size(C)
+  k = length(tau)
+  notran = (trans == 'N')
+  if side == 'L'
+    rng = notran ? (k:-1:1) : (1:k)
+    for i in rng
+      τi = notran ? tau[i] : conj(tau[i])
+      τi == zero(FC) && continue
+      Aii = A[i,i]
+      A[i,i] = one(FC)
```
langou, line 407:
```julia
      Aii = A[i,i]
      A[i,i] = one(FC)
```

I do not think this is good practice.

This is in particular important for UNMQR. For UNMQR you do want A to be input only. You do want to have an A that potentially UNMQR many blocks C in parallel. If values of A are changed, sometimes A[i,i] contains the correct value A[i,i] as it should, sometimes A[i,i] contains 1, then you cannot call UNMQR in a multithreaded shared-memory context, which feels natural to do when you understand the algorithm.

So this is (1) error prone (users assume A is input), and (2) if not error prone, then strongly limits potential for parallelization in shared memory.

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+      end
+      A[i,i] = Aii
+    end
+  else  # side == 'R'
+    rng = notran ? (1:k) : (k:-1:1)
+    for i in rng
+      τi = notran ? tau[i] : conj(tau[i])
+      τi == zero(FC) && continue
+      Aii = A[i,i]
+      A[i,i] = one(FC)
```
langou, line 423: I do not think this is good practice

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+      v = view(A, i:m, i)
+      p = m - i + 1
+      for j = 1:n
+        c = view(C, i:m, j)
+        s = kdot(p, v, c)
+        kaxpy!(p, -τi * s, v, c)
+      end
+      A[i,i] = Aii
```
langou, line 415: I do not think this is good practice

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+        for t = 1:q
+          s += r[t] * v[t]
+        end
+        d = τi * s
+        for t = 1:q
+          r[t] -= d * conj(v[t])
+        end
+      end
+      A[i,i] = Aii
+    end
```
langou, line 438: I do not think this is good practice

src/block_krylov_utils.jl
```diff
@@ -299,3 +299,155 @@ korgqr!(A :: AbstractMatrix{T}, tau :: AbstractVector{T}, buffer:: AbstractVecto
...
+end
+
+# """
+#     A = orgqr!(A, tau)
+#
+# Form the orthonormal factor `Q` from the reflectors produced by [`geqrf!`](@ref), overwriting `A`.
+# """
+function orgqr!(A::AbstractMatrix{FC}, tau::AbstractVector{FC}) where FC <: FloatOrComplex
+  m, n = size(A)
+  k = length(tau)
```
langou, line 360:
```julia
  m, n = size(A)
  k = length(tau)
```
Very good.

Maybe check that k ≤ min(m,n) and return an error if not.

test/test_aux.jl
```diff
@@ -191,4 +191,95 @@
...
+        # geqrf! + orgqr!: reduced QR factorization
+        QR = copy(A)
+        tau = zeros(FC, k)
+        Krylov.geqrf!(QR, tau)
+        R = triu(QR[1:k, 1:k])
+        Krylov.orgqr!(QR, tau)
+        Q = QR[:, 1:k]
+        @test istriu(R)
+        @test norm(Q' * Q - I) ≤ tol
+        @test norm(Q * R - A) ≤ tol
```
langou: relative error please.
```julia
        @test norm(Q * R - A) ≤ tol * norm(A)
```

test/test_aux.jl
```diff
@@ -191,4 +191,95 @@
       Krylov.kref!(n, x, y, c, s)
     end
   end
+
+  @testset "reduced QR (larfg! / geqrf! / orgqr! / ormqr!)" begin
+    @testset "accuracy $FC" for FC in (Float32, Float64, ComplexF32, ComplexF64, BigFloat, Complex{BigFloat})
+      T = real(FC)
+      tol = sqrt(eps(T))
```
langou: I think we can be a tat more ambitious with something like
```julia
      tol = eps(T)
```
And then below you can multiply `tol` by a quantity related to `m`, `n` and `k`.

test/test_aux.jl
```diff
@@ -191,4 +191,95 @@
...
+    @testset "matches LAPACK $FC" for FC in (Float32, Float64, ComplexF32, ComplexF64)
+      m, k = 9, 4
+      A = rand(FC, m, k)
+      Qg = copy(A); tg = zeros(FC, k); Krylov.geqrf!(Qg, tg);  Krylov.orgqr!(Qg, tg)   # pure Julia
+      Ql = copy(A); tl = zeros(FC, k); Krylov.kgeqrf!(Ql, tl); Krylov.korgqr!(Ql, tl)  # LAPACK
+      # Q is defined up to a phase, so compare the (phase-invariant) projectors
+      @test Qg[:, 1:k] * Qg[:, 1:k]' ≈ Ql[:, 1:k] * Ql[:, 1:k]'
+    end
+
+    @testset "no allocations $FC" for FC in (Float16, ComplexF16)
```
langou, line 251: I assume that this test makes sure that Krylov.geqrf! and Krylov.orgqr! and Krylov.ormqr! and Krylov.larfg! do not do any allocation of vectors. If my assumption is correct, this is very nice! Good check!

langou (commented):
This looks great! I think the best way to move forward with this is to merge this now and whomever is not happy can submit a PR.

test/test_aux.jl
amontoison: Yes, I have the same check for all (block) Krylov solvers.
I want to be sure that we don't allocation any additional vector / storage when we provide a `KrylovWorkspace`.

src/block_krylov_utils.jl
amontoison: @langou In Julia, scalar arguments are passed by value rather than by reference, unlike in Fortran.
Because of that, using an input/output scalar argument is not idiomatic.

While it is technically possible to wrap a scalar in a `Base.RefValue` and mutate it inside the function, this pattern is generally reserved for interfacing with C or Fortran libraries.
For pure Julia code, returning the updated scalar is the conventional approach.

So, although matching the LAPACK interface would make sense from a Fortran perspective,
I think following Julia's conventions leads to a more natural API.

src/block_krylov_utils.jl
amontoison: I implemented methods without buffer because we already dispatch to optimized (block) variant when the type is compatible with what is implemented in LAPACK.
I am not sure that we have a speed-up to use blocked geqrf, ungqr, unmqr, etc... with `Float16` or `BigFloat`.

src/block_krylov_utils.jl
amontoison: (1) The function is not exported, it is just for an internal use in `Krylov.jl` and the block Krylov solvers.
(2) I agree but do you think that we have a gain to do that with non-BLAS type ?

src/block_krylov_utils.jl
amontoison: @langou I propose that you do the modification in a follow-up PR, I am not sure to understand the modification that you want to do :)

amontoison:
@langou I addressed a good subset of your comments in https://github.com/JuliaSmoothOptimizers/Krylov.jl/pull/1100/commits/b85531028cbd8136f547c20d39888badca823606.
Thank you for the review!

PR #1104, opened 2026-08-03
Eliminate redundant vector zeroing and redundant norm evaluations in Krylov.jl

tapios:
Accelerates GPU performance of Krylov solver

**Changes**:
- Removed redundant `kfill!(V[i], zero(FC))` pre-allocation loops in `src/gmres.jl`, `src/fgmres.jl`, `src/dqgmres.jl`, `src/diom.jl`,  `src/fom.jl`, and  `src/gpmr.jl`. The basis vectors $V_1$ and $V_{k+1}$ are overwritten directly in-place by `kdivcopy!` during Arnoldi iterations, so pre-zeroing all $mem$ vectors at the start of each solve launched  20-40 unnecessary GPU kernels (incurring substantial latency in our atmosphere modeling test cases in ClimaAtmos.jl).
- Removed redundant residual norm recomputation (`β = knorm(n, r₀)`) at the start of passes in `gmres.jl`, `fgmres.jl`, and `fom.jl` when `npass == 1`.
- Fixed a pre-existing runaway-loop bug in FOM's timemax/callback handling, which the V zeroing was accidentally masking

### DIFF
--- a/src/diom.jl
+++ b/src/diom.jl
@@ -181,13 +181,7 @@ kwargs_workspace_diom = (:memory,)
     (verbose > 0) && @printf(iostream, "%5s  %7s  %5s\n", "k", "‖rₖ‖", "timer")
     kdisplay(iter, verbose) && @printf(iostream, "%5d  %7.1e  %.2fs\n", iter, rNorm, start_time |> ktimer)
 
-    mem = length(V)  # Memory
-    for i = 1 : mem
-      kfill!(V[i], zero(FC))  # Orthogonal basis of Kₖ(MAN, Mr₀).
-    end
-    for i = 1 : mem-1
-      kfill!(P[i], zero(FC))  # Directions Pₖ = NVₖ(Uₖ)⁻¹.
-    end
+    # Set up workspace.
     kfill!(H, zero(FC))  # Last column of the band hessenberg matrix Hₖ = LₖUₖ.
     # Each column has at most mem + 1 nonzero elements.
     # hᵢ.ₖ is stored as H[k-i+1], i ≤ k. hₖ₊₁.ₖ is not stored in H.
--- a/src/diom.jl
+++ b/src/diom.jl
@@ -266,6 +266,11 @@ kwargs_workspace_diom = (:memory,)
       # Compute the direction pₖ, the last column of Pₖ = NVₖ(Uₖ)⁻¹.
       # u₁.ₖp₁ + ... + uₖ.ₖpₖ = Nvₖ             if k ≤ mem
       # uₖ₋ₘₑₘ₊₁.ₖpₖ₋ₘₑₘ₊₁ + ... + uₖ.ₖpₖ = Nvₖ if k ≥ mem + 1
+      if iter < mem
+        # pₐᵤₓ ← Nvₖ
+        # The copy overwrites P[ppos], so the workspace does not need to be zero-initialized.
+        kcopy!(n, P[ppos], z)
+      end
       for i = max(1,iter-mem+1) : iter-1
         ipos = mod(i-1, mem-1) + 1  # Position corresponding to pᵢ in the circular stack P.
         diag = iter - i + 1
@@ -277,8 +282,10 @@ kwargs_workspace_diom = (:memory,)
           kaxpy!(n, -H[diag], P[ipos], P[ppos])
         end
       end
-      # pₐᵤₓ ← pₐᵤₓ + Nvₖ
-      kaxpy!(n, one(FC), z, P[ppos])
+      if iter ≥ mem
+        # pₐᵤₓ ← pₐᵤₓ + Nvₖ
+        kaxpy!(n, one(FC), z, P[ppos])
+      end
       # pₖ = pₐᵤₓ / uₖ.ₖ
       kdiv!(n, P[ppos], H[1])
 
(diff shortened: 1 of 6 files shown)


src/diom.jl
```diff
@@ -181,13 +181,7 @@ kwargs_workspace_diom = (:memory,)
...
     kdisplay(iter, verbose) && @printf(iostream, "%5d  %7.1e  %.2fs\n", iter, rNorm, start_time |> ktimer)
 
-    mem = length(V)  # Memory
-    for i = 1 : mem
-      kfill!(V[i], zero(FC))  # Orthogonal basis of Kₖ(MAN, Mr₀).
-    end
-    for i = 1 : mem-1
-      kfill!(P[i], zero(FC))  # Directions Pₖ = NVₖ(Uₖ)⁻¹.
-    end
+    # Set up workspace.
```
amontoison, line 184:
```suggestion
    # Set up workspace.
    mem = length(V)  # Memory
```

src/dqgmres.jl
```diff
@@ -182,11 +182,6 @@ kwargs_workspace_dqgmres = (:memory,)
     kdisplay(iter, verbose) && @printf(iostream, "%5d  %7.1e  %.2fs\n", iter, rNorm, start_time |> ktimer)
 
     # Set up workspace.
```
amontoison, line 184:
```suggestion
    # Set up workspace.
    mem = length(V)  # Memory.
```

amontoison:
You probably need another modification in `DIOM` and `DQGMRES`.
Do you have a nice speed-up with this modification?

tapios:
> You probably need another modification in `DIOM` and `DQGMRES`. Do you have a nice speed-up with this modification?

Yes, I'm on it. I like the pro-active review (but wasn't quite done :)). This does speed up our use cases (~10% or so, not massive but worthwhile).

The direction update accumulates into `P[pos]` via `kaxpy!` for the first memory iterations, so it must start at zero, unlike the `V` basis vectors, which are fully overwritten before use. So that zeroing must remain and I restored it now.

amontoison:
> > You probably need another modification in `DIOM` and `DQGMRES`. Do you have a nice speed-up with this modification?
> 
> Yes, I'm on it. I like the pro-active review (but wasn't quite done :)). This does speed up our use cases (~10% or so, not massive but worthwhile).

Nice! I was not expecting this step to take 10%.
Did you profile it with GPU tools? Is it just kernel launches or filling the vectors with 0 takes a lot of time?
 
> The direction update accumulates into `P[pos]` via `kaxpy!` for the first memory iterations, so it must start at zero, unlike the `V` basis vectors, which are fully overwritten before use. So that zeroing must remain and I restored it now.

We can use a `kaxpby!` instead of `kaxpy!` for the first index `i` in the following loop (if `iter <= mem`).
It will fuse the `kfill!` and `kaxpy!` if we use `zero(FC)` for the scaling factor of `y`.

```julia
      # Compute the direction pₖ, the last column of Pₖ = NVₖ(Rₖ)⁻¹.
      for i = max(1,iter-mem) : iter-1
        ipos = mod(i-1, mem) + 1  # Position corresponding to pᵢ in the circular stack P.
        diag = iter - i + 1
        if ipos == pos
          # pₐᵤₓ ← -hₖ₋ₘₑₘ.ₖ * pₖ₋ₘₑₘ
          kscal!(n, -H[diag], P[pos])
        else
          # pₐᵤₓ ← pₐᵤₓ - hᵢ.ₖ * pᵢ
          kaxpy!(n, -H[diag], P[ipos], P[pos])
        end
      end
```

tapios:
> Nice! I was not expecting this step to take 10%. Did you profile it with GPU tools? Is it just kernel launches or filling the vectors with 0 takes a lot of time?

The main issue seems to have been kernel launch latency. 

> We can use a `kaxpby!` instead of `kaxpy!` for the first index `i` in the following loop (if `iter <= mem`). It will fuse the `kfill!` and `kaxpy!` if we use `zero(FC)` for the scaling factor of `y`.
> 
> ```julia
>       # Compute the direction pₖ, the last column of Pₖ = NVₖ(Rₖ)⁻¹.
>       for i = max(1,iter-mem) : iter-1
>         ipos = mod(i-1, mem) + 1  # Position corresponding to pᵢ in the circular stack P.
>         diag = iter - i + 1
>         if ipos == pos
>           # pₐᵤₓ ← -hₖ₋ₘₑₘ.ₖ * pₖ₋ₘₑₘ
>           kscal!(n, -H[diag], P[pos])
>         else
>           # pₐᵤₓ ← pₐᵤₓ - hᵢ.ₖ * pᵢ
>           kaxpy!(n, -H[diag], P[ipos], P[pos])
>         end
>       end
> ```

Good idea. I think the above does not cover `iter = 1` and `β = 0` doesn't reliably erase garbage on GPUs (since `axpby!` is `scal!(n, beta, dy); axpy!(n, alpha, dx, dy)` and `scal` multiplies elementwise, so `0·NaN = NaN`.

I implemented an similar alternative in the last commit. PTAL.

src/fom.jl
```diff
@@ -220,19 +217,19 @@ kwargs_workspace_fom = (:memory,)
...
-      β = knorm(n, r₀)
       z[1] = β
       kdivcopy!(n, V[1], r₀, rNorm)  # v₁ = r₀ / ‖r₀‖
 
       npass = npass + 1
       inner_iter = 0
       inner_tired = false
 
-      while !(solved || inner_tired || breakdown)
+      while !(solved || inner_tired || breakdown || user_requested_exit || overtimed)
```
amontoison, line 232: Nice catch!!

amontoison (approved):

amontoison:
Thanks @tapios !
I will do a new release after #1100.

